PR #17, opened 2016-11-15
Fix typo when overriding pagination components

yamadapc:
Hey there, thanks a lot for this great work!

It seems there was a little typo preventing users from overriding the next/previous components. This PR fixes that. These properties are passed in lower-case to the component in `pagination.js`, but accessed in upper-case.


tannerlinsley:
Awesome, thanks!

PR #41, opened 2017-01-25
Moving hardcoded className's into defaultProps

corpix:
Hello.

I have moved hardcoded classes to `defaultProps`. This modification will allow to use this component with css modules(I want make a wrapper component which should look great in material-ui so I need this kind of flexibility).

Is it ok to do it this way? If so I will update the documentation.

Thank you.

### DIFF
--- a/src/index.js
+++ b/src/index.js
@@ -216,7 +216,7 @@ export default React.createClass({
       theadStyle,
       trClassName,
       trStyle,
-      thClassname,
+      thClassName,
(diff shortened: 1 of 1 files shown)


tannerlinsley:
This is cool. But are the core styles broken if you set one of these props? You would be required to set it to the original class name and your new class name

tannerlinsley:
See the changes in version 5. If this is still needed, reopen.

PR #61, opened 2017-02-05
'collapseOnChange' and 'freezeWhenExpanded'

janzheng:
rough draft of "closeSubComponentOnDataChange" and "preventAutoSortWhenComponentIsOpen"

Ok here's a rough idea of implementing what we talked about. My code is less than clean but hopefully, it implements what we want it to correctly. Sorry about all the `console.log` - I added them to understand how your abstractions worked. I can clean the code up once you've given me the green light. I also didn't spend a lot of time thinking of better names for these flags... they're verbose but explain what they do

1. closeSubComponentOnDataChange
- when set to true, all components close when data changes

2. preventAutoSortWhenSubComponentIsOpen
- prevents sorting from happening when any subcomponents are open. Basically, this helps my use case of inserting user data and forms into a subcomponent. When this flag is set to `true` then all the new data gets appended to the end of the `sortedData` array until the user either closes the subcomponents or clicks on a filter, which closes the subcomponents

3. Clicking on a filter now closes subcomponents
- this is to prevent unexpected behavior, since open components now refer to the same row but different data. This is to prevent user confusion

I also decided to split the `if` clause in your `setStateWithData` where one section explicitly compares new `oldState.resolvedData !== newResolvedState.resolvedData` and another compares sorting `oldState.sorting !== newResolvedState.sorting ` since it made adding some of these ideas a little easier. You're welcome to modify and simplify the code.

Also, bear with me as this is my first pull request ever! (I'm also a UX Designer by trade, so again, code might not be as efficient)


janzheng:
Yeah I agree with the new prop names. Some docs can make it clear what they do.

tannerlinsley:
Let's clean up the logging so we can get a better picture for what's going to be committed.

tannerlinsley:
So with these updated changes I pushed, the freezing strategy is much simpler:

- If `freezeWhenExpanded` is set, and at least one row is expanded, the table gains a `frozen` state.
- If the table is `frozen`, the sorted data won't update, but any new data and sorting will be set under the hood.
- When the table is unfrozen, it will update to the state it is meant to be at had the table not been frozen

The downside to this, which wasn't really going to be very possible either way is that when the data is changed, you cannot merge or update the data model at all, else you run the risk of your open component changing. 

This is very edge case right now and I'm still wondering if there is a better way to achieve what you want without altering the core update logic as much as we have.

tannerlinsley:
Okay, I'm good to merge this, just need you to remove the changes on the `SubComponents.js` story :)

janzheng:
I just reversed SubComponents.js to your code.

Thanks for all the help!

tannerlinsley:
You bet!  Good job mate! 👏 🎉

PR #82, opened 2017-02-15
update documentation about accessors

avnunn:
I'm having an issue where the accessor has a dot in it.

For example I have this dataset:
```json
{
  "foo.bar": "foobar"
}
```

Using the accessor "foo.bar" will be translated to ["foo", "bar"] array path.
In this scenario I just want ["foo.bar"] to be the array path.

This pull is not ideal -- what if I have a dataset such as:
```json
{
  "foo.bar": {
    "foo.baz": 1
   }
}
```

It would be ideal if I could supply an array as the accessor which isn't flattened.

For reference, libraries like lodash support this via their 'get' function.
https://lodash.com/docs/4.17.4#get

For now I will just supply a custom render function for reach column that properly accesses the data I want or just my fork.

### DIFF
--- a/README.md
+++ b/README.md
@@ -99,7 +99,7 @@ const data = [{
 
 const columns = [{
   header: 'Name',
-  accessor: 'name' // String-based value accessors !
+  accessor: 'name' // String-based value accessors! 
 }, {
   header: 'Age',
   accessor: 'age',
@@ -245,7 +245,7 @@ Or just define them as props
 ```javascript
 [{
   // General
-  accessor: 'propertyName' or Accessor eg. (row) => row.propertyName,
+  accessor: 'propertyName' // or Accessor eg. (row) => row.propertyName (see "Accessors" section for more details)
   id: 'myProperty', // Conditional - A unique ID is required if the accessor is not a string or if you would like to override the column name used in server-side calls
   sortable: true,
   show: true, // can be used to hide a column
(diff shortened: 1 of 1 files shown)


tannerlinsley:
The only strange edge case that could exist if we accept this is something like:
```
const columnDef = { accessor: 'my.path' }
const rowSample = { 'my.path': 10, my: { path: undefined }}
// The value would access to `10`, not `undefined`
```
Is this simply a liability the user takes on if their data is modeled that way?

tannerlinsley (approved):

tannerlinsley:
Actually, looking at this again, that could prove to be troublesome for your specific example.

tannerlinsley (changes requested):
Not sure yet on the changes required, but I think we could improve.

avnunn:
That is correct, there isn't a clear way forward since it is ambiguous.

This pull is mostly meant to highlight the issue and gather suggestions for improvement.

Ideally I think we could supply an array for the accessor field.
If an array is passed in then all it needs is to be flattened.

avnunn:
```
const columnDef = { accessor: ['my.path'] }
const rowSample = { 'my.path': 10, my: { path: undefined }}
// The value would access to `10`, not `undefined`
```
```
const columnDef = { accessor: ['my', 'path'] }
const rowSample = { 'my.path': 10, my: { path: undefined }}
// The value would access to `undefined`, not `10`
```

tannerlinsley:
This would effectively kill accessors for anyone utilizing recursive flattening:
```javascript
const data = { pre: { fix: { foo: 'bar' } } }
const prefix = 'pre.fix'
{
  accessor: [prefix, 'foo']
}
```

It feels like any schema with keys that contain `.`'s is very uncommon.  I just want to make sure we're not optimizing for an edge case that should ultimately be handled by the user to make sure they are complying with a deterministic api.

tannerlinsley:
Could this also be solved by using lodash itself?

```javascript
const getWithArray = path => row => _.get(row, path)

{
  accessor: getWithArray(['foo.bar', 'foo.baz'])
}
```

avnunn:
That should work and coincidentally how I am currently solving the issue.
There are a few ways to solve this issue on the client's end.

Keys that contain dots are probably uncommon, if this is something that isn't supported perhaps some documentation is in order to let users know the work-around.
For example I didn't even know that my accessor string was being interpreted to an array. Wasn't till I loaded up the code to see why my table was blank.

Unfortunately there isn't a universal standard of how to access data out of a JSON object.
Every framework I use seems to slice the problem in a different way.

Have you used json pointers? I've used them in a recent project and they seem nifty. They also handle the issue of "what happens if the meta characters are part of my key" by being able to encode the special characters used.

tannerlinsley:
Interesting.  Yeah at this point, I think some documentation and a workaround might be helpful in the docs.  In fact, if you want to switch this PR to that, I would accept it in a heartbeat :)

avnunn:
I've updated my commit to add some documentation about accessors.
It makes sense to me, but I also read the code to see how it works, but its still more information currently available.

Feel free to copy and make any changes you'd like.

tannerlinsley:
Very nice!  I'll merge and add some edits changes with the next release. Thanks man!

PR #91, opened 2017-02-19
Better support for subcomponents

manuxio:
I think this is a good start,
probably there should be a better way to pass more useful properties (check TdComponent ...)
Thank you!

### DIFF
--- a/.storybook/config.js
+++ b/.storybook/config.js
@@ -24,6 +24,7 @@ import FunctionalRendering from '../stories/FunctionalRendering.js'
 import CustomExpanderPosition from '../stories/CustomExpanderPosition.js'
 import NoDataText from '../stories/NoDataText.js'
 import Footers from '../stories/Footers.js'
+import ExtendingComponents from '../stories/ExtentingComponents.js'
 //
 configure(() => {
   storiesOf('1. Docs')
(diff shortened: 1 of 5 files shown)


manuxio:
Hi, I'll give it a deeper look...
meanwhile, it would help me a lot to have a linter for your project... isn't there any?
I mean, some eslint configuration?

tannerlinsley:
Standard.js is a dev dependency. Just npm run standard. You could also
install the standard linter plugin for your ide

tannerlinsley:
Did that linter work for you @manuxio ?

manuxio:
Actually, not... 
npm ERR! missing script: standard


But that's no big deal... the point is that I have had very little time to check the PR ... will do it asap (but also found that I don't want to replicate all data... just the very important ones... so the first PR will not be very different from the next release... 
Just a few more lines and I'll be done...

tannerlinsley:
Be sure to yarn or npm install to get standard. If its not there, it should
be a dev dep

manuxio:
Did both... no help... as a matter of fact, the only lint I saw was by github when I made the PR...
Sorry, simply can't understand that :(

tannerlinsley:
To Lint:
- Make sure the linter is installed by running `npm install`.  This will install all dev dependencies.
- Then run `npm run test`.  This will run the standard.js linter and print any warnings to fix
Optional:
- Install the standard.js linter for atom, sublime, or VScode, and you'll get linting right in your editor.

tannerlinsley:
Still very interested in this.  If you want to keep running with this PR, feel free to reopen and we can keep working through it.

PR #100, opened 2017-02-27
Rename CollapseOnPageChange to CollapseOnDataChange

TacticalCoding:
Rename CollapseOnPageChange to CollapseOnDataChange to more accurately reflect its purpose.
Implement CollapseOnPageChange option with a default value of true.

The expanded rows are always collapsed on a page change regardless of what this setting is set to: [methods.js:onPageChange](https://github.com/tannerlinsley/react-table/blob/master/src/methods.js#L288)


tannerlinsley:
This is great. Would you also like to add a real option for collapsing on page changes Defaulting to true?

TacticalCoding:
Sure

TacticalCoding:
Can I make collapseOnSortingChange default to true so that all three collapse options match?
  collapseOnSortingChange: true,
  collapseOnPageChange: true,
  collapseOnDataChange: true,

tannerlinsley:
I initially thought `collapseOnSortingChange: false` was the majority use case.  Would you say otherwise?

TacticalCoding:
It seems to me that if your data is moved around because it's sorted that you wouldn't want a different row than you originally expanded to now be expanded in most cases. Or if you want the particular row to always be expanded, then I would expect all three options to be false. But that's just my opinion. My particular use case requires collapseOnSortingChange to be false, but collapseOnDataChange to be true. Not a big deal either way.

tannerlinsley:
I would be okay with the default change to `true`.  Where false becomes valuable is on a pivotTable with expanded rows.  But honestly there is a whole slew of other questions and issues that needs to be discussed there.  

Make the change 👍

src/lifecycle.js
```diff
@@ -42,7 +42,8 @@ export default {
       oldState.data !== newState.data ||
       oldState.columns !== newState.columns ||
       oldState.pivotBy !== newState.pivotBy ||
-      oldState.sorting !== newState.sorting
+      oldState.sorting !== newState.sorting ||
+      oldState.page !== newState.page
```
tannerlinsley: I think this might be bad. Changing pages would trigger a dataModel update (which is expensive) when it doesn't really need to.

TacticalCoding:
Added support for collapseOnPageChange.

I also updated the SubComponent story to be able to dynamically set table options so that users can see an example of their behavior.

I feel like it would ultimately be useful to have that on every story but I didn't want to go through them right now.

Let me know if you're OK with the new story.

src/lifecycle.js
TacticalCoding: I see what you're saying. I'll change it.

tannerlinsley:
Love the new story :)

tannerlinsley:
Very nice!  Great work!

Issue #114, opened 2017-03-07
Not possible to use table inside of the form


rmv125:
If we try to use table with pagination inside of the form, we'll have nested forms error.

That happens because of 
https://github.com/tannerlinsley/react-table/blob/master/src/pagination.js
pageJump is the form.
So, i think it can be refactored to avoid using form.
Also, is there a way to pass custom pageJump or pagination?

tannerlinsley:
Yes, you can override any core components you want.  Refer to https://github.com/tannerlinsley/react-table#component-overrides.

As for the form element, I would accept a PR to replace the form element with more specific event driven logic.  It would just need to listen for the return key when the input is focus, and perform the action that the form submission is performing right now :)

PR #122, opened 2017-03-09
use button type button in pagination to prevent form submitting if using inside of the form

rmv125:
connected to #114
Sorry notice that after last PR was already merged.


tannerlinsley:
Nice catch :)

Issue #47, opened 2017-01-30
Unable to run repo


deepakkj:
Hi,
I'm trying to run this repo after forking but unable to do so.
I have tried the steps mentioned in the Readme file..

When I do `yarn` , I get the following error:
![yarn](https://cloud.githubusercontent.com/assets/9112236/22436708/3780a1fe-e74b-11e6-80ae-53b76531626d.PNG)

I'm not sure if this is due to yarn itself or something else.
I have also attached the debug log
[npm-debug.txt](https://github.com/tannerlinsley/react-table/files/739962/npm-debug.txt)
.

tannerlinsley:
Looks like the rm come isn't working on Windows. I may just need to update rm to a Windows compatible command.

deepakkj:
This `npm run-script storybook` works in Windows.
But once webpack starts building I get 

> module not not found error
 for react-table.css at line no. 9  in .storybook/config.js.

![1](https://cloud.githubusercontent.com/assets/9112236/22463450/a64fe09a-e7d7-11e6-9ff8-7890f62574b3.PNG)
![2](https://cloud.githubusercontent.com/assets/9112236/22463451/a82615ba-e7d7-11e6-8510-f58796df4ef3.PNG)


On removing this line `import '../react-table.css'`, webpack builds successfully and I can run open localhost but without styling for the react-table.css
![3](https://cloud.githubusercontent.com/assets/9112236/22463456/ac88730a-e7d7-11e6-82eb-79a21e2a0e16.PNG)

tannerlinsley:
Before you run the storybook, you'll want to watch (`yarn run watch:css`) or build (`yarn run build:css`) the styles.  Then the storybook should work just fine.

tannerlinsley:
Closing for now. If you still have issues, reopen.

deepakkj:
@tannerlinsley Would like to reopen this issue and want to submit a PR for this.

deepakkj:
@tannerlinsley Thanks. Submitted a PR (#124) 
`rm -rf` doesnn't work on Windows. Hence updated it with **rimraf** package.

PR #124, opened 2017-03-12
Updated the package.json - Issue #47

deepakkj:
Removed `rm -rf` since it works on Linux environments.
Instead added **rimraf** npm package as it is supported in all environments even Windows.


tannerlinsley:
Nice work :)

PR #147, opened 2017-03-28
Add column filtering

TacticalCoding:
* Add new option to table `showFilters` which adds a header row with input fields that act as filters
* Add a table option to set the default filter method `defaultFilterMethod`
* Add column option `filterMethod` to override the table's default filter method
* Add column option `hideFilter` to selectively hide a column's filter
* Known Issue: Filtering a pivot column works, but filtering an aggregate column may not have the effect that someone wants. It's difficult to know the exact behavior a user would want in this case. It will filter on the individual sub rows, and not the aggregate value. Because of this it will only show rows with exactly 1 subrow that matches the filter.


src/index.js
```diff
@@ -526,7 +667,6 @@ export default React.createClass({
     const makePadRow = (row, i) => {
       const trGroupProps = getTrGroupProps(finalState, undefined, undefined, this)
       const trProps = _.splitProps(getTrProps(finalState, undefined, undefined, this))
-      const tdProps = _.splitProps(getTdProps(finalState, undefined, undefined, this))
```
tannerlinsley: Was this unused?

tannerlinsley: Oh, yep, looks like it was

tannerlinsley:
This is fantastic Aaron!  What do you think we should do about pivoted sorting and filtering.  It would be cool if we could support both asap right?

src/defaultProps.js
```diff
@@ -23,7 +23,7 @@ export default {
   defaultSorting: [],
   showFilters: false,
   defaultFiltering: [],
-  defaultFilterMethod: (filter, row, column) => (row[filter.id] == filter.value),
+  defaultFilterMethod: (filter, row, column) => (row[filter.id] === filter.value),
```
TacticalCoding: This actually doesn't behave how I think it should for columns with numbers. I suppose I could cast the row's value to a string, but that didn't seem as clean as fuzzy equality. I understand the need to pass the linter though for the CI.

TacticalCoding:
I think the pivoted sorting works ok the way it is now.
I can take another look at the pivoted filtering. It looks like I'm going to have to recursively look through each column and it's subkeys to find the right column id to filter on. I don't personally use the pivoting so I haven't dug into how it works too much yet.

tannerlinsley:
Well, do you think we're good to merge?

TacticalCoding:
Let me take another look today. We'll be able to merge by the end of the day.

edward93:
Ohhh this would be nice. I'm so waiting for this feature to be merged. ))

tannerlinsley:
Sounds good.  Let me know!

TacticalCoding:
Ok. I think it's ready to be merged if the code looks OK. The only strange thing I see is the way it deals with aggregate columns, but I'm not sure what to do about it.

Filtering a pivot column works, but filtering an aggregate column may not have the effect that someone wants. It's difficult to know the exact behavior a user would want in this case. It will filter on the individual sub rows, and not the aggregate value. Because of this it will only show rows with exactly 1 subrow that matches the filter.

tannerlinsley:
Fantastic. I'll make a release first thing tomorrow morning

PR #158, opened 2017-03-30
create ISSUE_TEMPLATE.md

mikedevita:
hopefully this will alleviate some clutter on issues.


mikedevita:
what do you mean by the unminified release? do you mean for the library?

tannerlinsley:
Yeah, that way, we could request that they try and reproduce their issues in the bin, and likewise, if the bin was using the latest unminified/unuglified UMD build, we could debug and track down the issue right there in devtools.

PR #160, opened 2017-03-30
Set page to max page available if data changes and has less pages than before

TacticalCoding:
May fix https://github.com/tannerlinsley/react-table/issues/159


src/methods.js
```diff
@@ -10,10 +10,7 @@ export default {
       pivotValKey,
       subRowsKey,
       expanderColumnWidth,
-      SubComponent,
```
tannerlinsley: Unused variables?

PR #162, opened 2017-03-31
Add filterRender column option to customize the filter that is shown

TacticalCoding:
The filters are now completely customizable if needed.
This is useful for columns that are booleans or columns where you want a dropdown filter.


src/index.js
```diff
@@ -419,14 +419,20 @@ export default React.createClass({
...
-                    }}
-                    value={filter ? filter.value : ''}
-                    onChange={(event) => this.filterColumn(column, event, col)}
-                  />
+                  _.normalizeComponent(col.filterRender,
+                    {
+                      onFilterChange: (value) => (this.filterColumn(column, value, col))
+                    },
+                    (
+                      <input type='text'
```
tannerlinsley: We should extract this out into a default prop (which will also involve passing more props to it in addition to `onFilterChange`

src/index.js
```diff
@@ -489,13 +495,20 @@ export default React.createClass({
...
-                width: `100%`
-              }}
-              value={filter ? filter.value : ''}
-              onChange={(event) => this.filterColumn(column, event)}
-            />
+            _.normalizeComponent(column.filterRender,
+              {
+                onFilterChange: (value) => (this.filterColumn(column, value))
+              },
+              (
```
tannerlinsley: It could then be reused here too.

src/index.js
TacticalCoding: I'm not sure I completely follow. I tried to imitate what the normal column renderer was doing. The pivot version also calls a slightly different version of filterColumn() than the non pivot version.

src/index.js
```diff
@@ -419,14 +419,20 @@ export default React.createClass({
...
-                    onChange={(event) => this.filterColumn(column, event, col)}
-                  />
+                  _.normalizeComponent(col.filterRender,
+                    {
+                      onFilterChange: (value) => (this.filterColumn(column, value, col))
+                    },
+                    (
+                      <input type='text'
+                        style={{
+                          flex: 1
```
tannerlinsley: Have you tried the filters with custom column widths? Do they line up?

src/index.js
tannerlinsley: So, instead of defining the default component inline in index.js, you could set the default filter component for every column here: https://github.com/tannerlinsley/react-table/blob/master/src/defaultProps.js#L107, like this:

```javascript
column: {
  ...
  filterRender: ({
    filter,
    onFilterChange
  }) => (
    <input type='text'
      style={{
        flex: 1
      }}
      value={filter ? filter.value : ''}
      onChange={(e) => onFilterChange(event.target.value)}
    />
  )
}
```

Then in index.js, you can use it like this:
```javascript
_.normalizeComponent(col.filterRender, {
  onFilterChange: value => this.filterColumn(column, value, col)
}
//...
_.normalizeComponent(col.filterRender, {
  onFilterChange: value => this.filterColumn(column, value)
```

tannerlinsley: Also, see that I'm sending in `filter`, and `onFilterChange` to the component, which gives more control.  In fact, we should probably send through all of the information as we do here: https://github.com/tannerlinsley/react-table#custom-cell-header-and-footer-rendering

src/index.js
TacticalCoding: Normal columns looked good, but you're right about the pivot columns with custom widths being messed up. I'll take care of that.

src/index.js
TacticalCoding: Sounds good. I like it.

src/index.js
tannerlinsley: 👍

src/index.js
tannerlinsley: 👍

tannerlinsley:
Superb!!

Issue #163, opened 2017-04-02
Odd behavior with sorted data state and `sorting` prop


mannieschumpert:
## Problem Description
Modifying the `sorting` prop via a container state does not have the expected results. As others have noted (https://github.com/tannerlinsley/react-table/issues/145), the header class names update one click before the _data_ does.

In the following example, I have included a simple element to modify the container state, which gets passed to ReactTable. (Note that I am preventing multisort by only ever passing one sort object.) When this element is first clicked, the 'name' column header class is changed and highlights as expected, but the data does not sort. After an _additional_ click of the "Sort" element, the data then sorts. In a different experiment, if logic was built into an `onSortingChange` handler to toggle the sort direction, the header's sort direction class would become the _inverse_ of the data's actual sort direction.

```javascript
return (
    <div>
      <p onClick={()=>this.setState({sort:{id:'name'}})}>Sort</p>
      <ReactTable
        data={data}
        columns={columns}
        defaultPageSize={10}
        getTrProps={this.setRowProps}
        getTdProps={this.setCellProps}
        sorting={[this.state.sort]}
      />
    </div>
)
```

## Steps to Reproduce
1. Add `sorting` prop that is set via container state
2. Modify container state via some action
3. Header of target column changes class immediately, but data does not re-sort until an additional action

## System Information
**Chrome 56.0.2924.87**
**Node Version: v6.10.1**
**OS Version: Mac OS Sierra 10.12.3**
**NPM Version: 3.10.10**

mannieschumpert:
I looked into the `sortColumn` method in the source code. If an `onSortingChange` prop is included, the function returns that passed function immediately. However, when that prop is not included, there are clearly some additional functions triggered by default, e.g. `setStateWithData`. So, there seem to be some critical methods not being triggered if a user passes the `onSortingChange` handler.

Of note: everything works as expected if no `sorting` or `onSortingChange` props are used.

tannerlinsley:
Good observation. I'll investigate our options here. It may be as simple as
making shot that changes to sorting props are properly triggering a sorting
model update.

abbotteric:
I believe this is an issue with the order the arguments are being passed in in "getResolvedState." The way it is currently, the existing props take precedence over the passed in new state. I submitted a PR #166 that fixes the issue for me.

tjphopkins:
Unfortunately the issue still seems to exist when the onSortingChange prop is used.

karanattriCaw:
Hi,
[...]

Issue #145, opened 2017-03-27
Sorting on Fully Controlled Component


janvp:
Hi,

I'm trying to use ReactTable with full control of the state (page, pageSize, sorting), as described in the documentation (chapter Fully Controlled Component). But when using the sorting, the actual sorting is always one sort-click behind. 

I can send an example to show the issue (I wanted to make a fiddle or codepen, but I don't know how to include the react-table library).

tannerlinsley:
Can you paste your code here?

tannerlinsley:
Try upgrading to version 5.3.5

janvp:
ok, here is my code. I didn't support multiple sorting here. 
I'll try also to upgrade to the new version. 
```
class FullyControlledTable extends React.Component {
    constructor(props) {
        super(props);
        this.state = {
            sort: [],
            page: 0,
            pages: Math.ceil(this.props.data.length / 10),
            pageSize: 10
        };
        this.handleSort = this.handleSort.bind(this);
        this.handlePageChange = this.handlePageChange.bind(this);
        this.handlePageSizeChange = this.handlePageSizeChange.bind(this);
    }

    handleSort(column) {
        let sort = [];
        if (this.state.sort[0] && this.state.sort[0].id == column.id) {
            sort.push({
                id: column.id,
                desc: !this.state.sort[0].desc
            });
        } else {
            sort.push({
                id: column.id,
                desc: false
            });
        }
        this.setState({
            sort: sort
        });
    }

    handlePageChange(pageIndex) {
        this.setState({
            page: pageIndex
        });
    }

    handlePageSizeChange(pageSize, pageIndex) {
        this.setState({
            pageSize: pageSize,
            page: pageIndex,
            pages: Math.ceil(this.props.data.length / pageSize)
        });
    }

    render() {
        return (
            <ReactTable
                data={this.props.data}
                columns={this.props.columns}
                page={this.state.page}
                pages={this.state.pages}
                pageSize={this.state.pageSize}
                sorting={this.state.sort}
                onSortingChange={this.handleSort}
                onPageChange={this.handlePageChange}
                onPageSizeChange={this.handlePageSizeChange}
            />
        );
    }
}
```

tannerlinsley:
You should be able to directly setState with the sorting that given to you in `onSortingChange`.

```javascript
 handleSort(newSortingModel) {
        this.setState({
            sort: newSortingModel
        });
    }
```

Does that work for you?

janvp:
[...]

PR #166, opened 2017-04-03
Change precedence in 'getResolvedState'

abbotteric:
* Previously existing props would overwrite passed in state
  * Now passed in state gets precedence

### DIFF
--- a/.storybook/config.js
+++ b/.storybook/config.js
@@ -25,6 +25,7 @@ import CustomExpanderPosition from '../stories/CustomExpanderPosition.js'
 import NoDataText from '../stories/NoDataText.js'
 import Footers from '../stories/Footers.js'
 import Filtering from '../stories/Filtering.js'
+import ControlledTable from '../stories/ControlledTable.js'
 //
 configure(() => {
   storiesOf('1. Docs')
@@ -55,4 +56,5 @@ configure(() => {
     .add('Custom "No Data" Text', NoDataText)
     .add('Footers', Footers)
     .add('Custom Filtering', Filtering)
+    .add('Controlled Component', ControlledTable)
 }, module)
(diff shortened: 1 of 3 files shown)


tannerlinsley:
I'm unsure of the implications of doing this.  It seems to be right, but this could affect a lot of internals.  Have you tested each story in the storybook thoroughly to ensure this will be okay?

abbotteric:
I'll take a look through the storybook tonight and make sure it's covered. If I remember correctly the function only gets called with args in one other instance. Thanks for the feedback, I'll update with my findings soon.

abbotteric:
From what I can tell, the only other call to getResolvedState() that passes arguments is in componentWillReceiveProps(). Line 39 in lifecycle.js accepts nextState as the 2nd argument. In this case nextState represents the React context which I don't believe ReactTable is currently setup to consume so nextState is always the Object prototype.

Currently the precedence goes: new props -> previous props -> context -> component state
This PR would change that to: new props -> context -> previous props -> component state

Since ReactTable doesn't use context right now, swapping it with previous props won't have any effect on current operation but I can see why some might want to keep it the way it is. If that's the case then the call to getResolvedState on line 65 of lifecycle.js would need to swap the order of arguments. That may come with other side effects that would still need to be investigated. Either approach would resolve #163 however.

FWIW I ran through the storybook and didn't run into any issues with the change in this PR. There aren't any stories for fully controlled components however. If I have some time I'll try to add one or two.

Thoughts?

tannerlinsley:
If you could add a fully controlled story entry then I think we're good to
merge :)

abbotteric:
Added a story entry with controlled sorting, page, and pageSize. Everything seems to be working as expected. If you were to revert commit 383554d2 you would see the behavior described in #163 and #145

tannerlinsley:
Stupendous! You rock man. Merging this now.

Issue #58, opened 2017-02-04
Support for resizable columns


bruth:
Are there any plans for supporting resizable columns? [example](http://facebook.github.io/fixed-data-table/example-resize.html)

tannerlinsley:
Its definitely been brought to my attention but I haven't moved on the idea yet. Other than deciding where and how to manage the column state to support it, it would be or straight forward.

NicolaFenix:
+1

spacedarcy:
+1

tannerlinsley:
@NicolaFenix @spacedarcy, please use the thumbs up on the original comment, then remove your +1 comment :)  Just keeping things clean.

TacticalCoding:
I'm working on an implementation that will hopefully be available soon.

https://github.com/TacticalCoding/react-table/tree/Resize-Columns

RajuTicvic:
Hey Hi
I'm using react table with material UI Enhanced table
I don't want a resizable column for checkbox column  

Can I have help Please

PR #170, opened 2017-04-05
Resize columns

TacticalCoding:


tannerlinsley:
Closes #58

tannerlinsley:
This is sooooo cool.  The only thing I found that seems funky is the server-side data example.  Any time I resize, it wants to load new data.  Not sure if it's the way the example is set up or the new resize logic. Thoughts?

TacticalCoding:
I did notice that and then completely forgot about it. Let me take a look.

Issue #169, opened 2017-04-04
Data is being fetch on current page box blur


alebrozzo:
## Problem Description
I have a table that handles paging and sorting on the server side. Everything seems to work fine, except that when I manually enter the page I want to go and hit enter, the page is refreshed but then on blur the page is refreshed again and the same data is being retrieved.

## Steps to Reproduce
1. On a multi-page table, navigate by manually entering the current page and hit enter
2. Leave the current page box

## System Information
**Chrome 57**
**Win10 Pro**

tannerlinsley:
This is definitely a bug and needs to get fixed.  If you're interested in fixing it, I would be happy to help you through it.

tannerlinsley:
Near [this line](https://github.com/tannerlinsley/react-table/blob/master/src/pagination.js#L29), there needs to be logic that checks if the desired page is already the current page.  If it is, then just return without changing the page.  If it not the same page, continue as normal and change the page.

alebrozzoSP:
Hey @tannerlinsley , I just created a PR with a fix to this issue (I am the creator of the issue, but with a different account). Rather than applying the fix where you said (function applyPage), I applied it on changePage. I also added an extra fix to getSafePage for when the parameter date is NaN.

tannerlinsley:
Merge :)

PR #178, opened 2017-04-07
Avoid page refresh if same page, set NaN to current page

alebrozzoSP:
This fixes issue [169](https://github.com/tannerlinsley/react-table/issues/169)

* The top change asks if the new page is really a number before trying to use it (this avoids warning when typing two non-numeric characters) and return the current page otherwise

* The bottom change verifies that the current page is not the same as the new page

### DIFF
--- a/src/pagination.js
+++ b/src/pagination.js
@@ -17,12 +17,17 @@ export default React.createClass({
     this.setState({page: nextProps.page})
   },
   getSafePage (page) {
+    if(isNaN(page)) {
+        page = this.props.page
+    }
     return Math.min(Math.max(page, 0), this.props.pages - 1)
   },
   changePage (page) {
     page = this.getSafePage(page)
     this.setState({page})
-    this.props.onPageChange(page)
+    if(this.props.page!==page) {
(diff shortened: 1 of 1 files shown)


src/pagination.js
```diff
@@ -17,12 +17,17 @@ export default React.createClass({
     this.setState({page: nextProps.page})
   },
   getSafePage (page) {
+    if(isNaN(page)) {
```
alebrozzoSP: This fixes the warning you'd get if typing two non-numeric characters, which caused page to be NaN

src/pagination.js
```diff
@@ -17,12 +17,17 @@ export default React.createClass({
...
+        page = this.props.page
+    }
     return Math.min(Math.max(page, 0), this.props.pages - 1)
   },
   changePage (page) {
     page = this.getSafePage(page)
     this.setState({page})
-    this.props.onPageChange(page)
+    if(this.props.page!==page) {
```
alebrozzoSP: Actual fix to issue #169

tannerlinsley:
Excellent!

PR #179, opened 2017-04-08
Fix storybook code highlight

alebrozzo:
There were some differences between the actual code and the code highlight, particularly important on `Footers.js`


tannerlinsley:
Right. The footers you fixed were actually being generated outside of the table callbacks. We need to properly document the way it was. For performance reasons, it would be smart to precalc things like that. Wanna add those differences somewhere in the docs?

Issue #14, opened 2016-11-06
The default css overwrites the checkbox appearance if included into a custom cell.


hegelstad:
http://stackoverflow.com/questions/20057581/css-webkit-appearance-none-is-causing-checkbox-to-not-be-checked

tannerlinsley:
I would accept a PR that fixes this @hegelstad

tannerlinsley:
Closing for now, but will still accept a PR to fix this edge case. :)

PR #197, opened 2017-04-13
Issue #14 (default CSS for checkboxes)

theCmaker:
Just a little fix. Now all the input tags except checkboxes and radios have their appearance set to none.
Was there any reason for resetting their appearance?


tannerlinsley:
Good call. :)

PR #204, opened 2017-04-17
Expander header text

TacticalCoding:
Addresses https://github.com/tannerlinsley/react-table/issues/198.

Adds an expanderHeaderText option.


tannerlinsley:
I actually had an idea for this.  In the Custom Expander example here: https://github.com/tannerlinsley/react-table/blob/master/stories/CustomExpanderPosition.js, you can use a column that is specifically designed as the placeholder for the expander column.  Wouldn't it be cool if you could configure it as you would any other column, including Header and Render?

TacticalCoding:
Yep you're right. I didn't realize there was already a special column. Went right over my head even though I modified that story...  
We could get rid of the expanderColumnWidth option as well and just set it in the column.

tannerlinsley:
Awesome!

PR #248, opened 2017-05-04
when you have no data pages shows as 0 instead of 1

terencechow:
see the No Data in the example storybook. Pagination at the bottom shows 1 of 0 instead of 1 of 1.

https://react-table.js.org/?selectedKind=2.%20Demos&selectedStory=Custom%20%22No%20Data%22%20Text&full=0&down=0&left=1&panelRight=0&downPanel=kadirahq%2Fstorybook-addon-actions%2Factions-panel

### DIFF
--- a/src/pagination.js
+++ b/src/pagination.js
@@ -104,7 +104,7 @@ export default class ReactTablePagination extends Component {
               </div>
             ) : (
               <span className='-currentPage'>{page + 1}</span>
-            )} {this.props.ofText} <span className='-totalPages'>{pages}</span>
+            )} {this.props.ofText} <span className='-totalPages'>{pages || 1}</span>
           </span>
           {showPageSizeOptions && (
             <span className='select-wrap -pageSizeOptions'>


tannerlinsley:
If there is no data I think we should display something special. Possibly something like "0 Pages"

mikedevita:
agreed @tannerlinsley maybe something like a span across all the cols "No Data" or something, i already do this except i swap out the table for a `<p className="well text-center text-bold">No Data</p>`

tannerlinsley:
So lets change it to show nothing for the page information if there is no data.

terencechow:
I prefer page 1 of 1. It already shows a msg that says no row data and it
makes sense that if there is no row data, you're still technically on page
1 of 1.

If you're gonna hide page information when there is no data, you might as
well hide page information when there is only 1 page of data since the user
can't navigate anyways.

One more thing is if you are displaying the data and allowing real time
changes it will look wierd for pagination to suddenly show up and push the
rest of your layout down. Likewise a real time edit to delete everything
will suddenly push everything up as the pagination disappears. From a
design perspective that is a very amateur experience

My vote goes for showing page 1 of 1.

On 6 May 2017 6:35 pm, "Tanner Linsley" <notifications@github.com> wrote:

So lets change it to show nothing for the page information if there is no
data.

—
You are receiving this because you authored the thread.
Reply to this email directly, view it on GitHub
<https://github.com/tannerlinsley/react-table/pull/248#issuecomment-299651194>,
or mute the thread

PR #251, opened 2017-05-04
Documentation example of using pivotRender

bolte-17:
Thanks @aaronschwartz for pointing out correct usage of pivotRender to remove the `(count)` in the group header. Didn't seem to be documented yet, but looks like it should be possible to work into the Pivoting Options example.


tannerlinsley:
This is now deprecated in the master branch in favor of the PivotValueComponent

Issue #237, opened 2017-04-28
Column info in rowInfo object


skyflyer:
## Problem Description
Currently, custom renderer for cell includes a `rowInfo` object, which provides index, but not the column (field) information when rendering. Would it make sense to include it for dynamic rendering? I'm trying to have certain columns editable... and perhaps there is a better way to achieve this?

## Codepen
http://codepen.io/anon/pen/KmWxBj?editors=0010

Here you can see the custom cell renderer for `age` column. I used a workaround with `getTdProps` to add column info to rowInfo object, so I can use it there (but still need to figure out how to change the underlying data).

## Steps to Reproduce
1. add a custom cell render in a `columns` column definition
2. try to determine which column it is rendering inside the custom render function

## System Information
* 57.0.2987.133 (64-bit)
* **Node Version: Unknown (codepen) **
* **OS Version: Unknown (codepen) **
* **NPM Version: Unknown (codepen) **
* **Yarn Version: Unknown (codepen) **

aaronschwartz:
You could enhance your renderInput function to take in the id like this:

http://codepen.io/anon/pen/XRMxpR?editors=0010

Or I'm sure if you make a PR which passes in the Column ID to the render function that it would be accepted.

skyflyer:
Aaron, thanks for that! Unfortunately, it still requires an "id" to be passed in, but it is a very nice trick with a closure. Also, thanks for taking the time to bind renderInput correctly so that it updates state.

While thinking about PR, I wonder what to do when the accessor is not a string. Do you think it would make sense to pass `columnAccessor` through the `rowInfo` object whatever the value of it might be (`string`, `function`, `undefined`, ...)?

aaronschwartz:
I would be hesitant to pass just the accessor. I think passing the id which is always a string through would be ok. If you really want the accessor, you might as well pass the whole column back to the render function so that you can use anything in the column.

skyflyer:
Oh, I've been playing with it a bit more and found out that the `id` property is mandatory if using accessor as a function; you are completely right, `id` is always a string.

If you think it makes sense for others as well, I'll prepare a PR where I'll add `columnName` to the `rowInfo` object. Otherwise, I'll just fill it with the `getTdProps` for my use case.

skyflyer:
I hope purely referencing this issue in a fork is sufficient as a pull request? If not, let me know I'll specifically create a pull request for this.

[...]

PR #253, opened 2017-05-06
Add column id to the rowInfo object

skyflyer:
Would you consider adding `columnId` property to the `rowInfo` object? It would be very useful when referencing data in a table in a generic way. I've explained the rationale in the #237 as well.

### DIFF
--- a/.storybook/config.js
+++ b/.storybook/config.js
@@ -27,6 +27,7 @@ import Footers from '../stories/Footers.js'
 import Filtering from '../stories/Filtering.js'
 import ControlledTable from '../stories/ControlledTable.js'
 import PivotingOptions from '../stories/PivotingOptions.js'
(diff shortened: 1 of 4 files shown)


tannerlinsley:
Yes. I think we should implement it differently though. Instead of mutating
the row info object, we should create a new object at the beginning of each
column loop. Use the spread property to take on all of the properties of
rowInfo, and then add in the entire column definition itself instead of
just the column id.

skyflyer:
OK. Just be sure that I understand it:

1. rename `rowInfo` to `cellInfo` (seems appropriate)
1. copy all of the props from `rowInfo` and add `column` to it (`const cellInfo = { ...rowInfo, column: ...column } `)
1. Fix storybook and doc references where `rowInfo -> cellInfo`?

tannerlinsley:
Perfect 👌

Issue #210, opened 2017-04-19
Dynamic Columns not compatible with filtering


theCmaker:
## Problem Description
I am currently working on a project that needs columns to be shown/hidden dynamically. To toggle them, I use the `show` property on `columns`. Some of these columns can be filtered (using the default filter). But when I want to hide a column with an active filter, I get the following error:

```log
react-table.js:6 Uncaught TypeError: Cannot read property 'filterMethod' of undefined
    at https://unpkg.com/react-table@5.5.3/react-table.js:6:10800
    at Array.filter (native)
    at https://unpkg.com/react-table@5.5.3/react-table.js:6:10601
    at Array.reduce (native)
    at Object.filterData (https://unpkg.com/react-table@5.5.3/react-table.js:6:10571)
    at Object.getSortedData (https://unpkg.com/react-table@5.5.3/react-table.js:6:10218)
    at Object.setStateWithData (https://unpkg.com/react-table@5.5.3/react-table.js:6:6019)
    at Object.componentWillReceiveProps (https://unpkg.com/react-table@5.5.3/react-table.js:6:5356)
    at https://cdnjs.cloudflare.com/ajax/libs/react/15.4.2/react-dom.js:5065:23
    at measureLifeCyclePerf (https://cdnjs.cloudflare.com/ajax/libs/react/15.4.2/react-dom.js:4529:12)
```

In file `methods.js` near line 266 (in `filterData()`), data filtering is performed on visible columns, but the `filtering` list is kept aside. When a filter is active on a column that gets hidden, `filtering` still has a filterId referencing a column which is not accessible anymore in `visibleColumns` (the origin of the error I got).

Is there a trick to reset the filter before "removing" the column? (Except asking the user to do it :/)

## Example
```js
const ReactTable = window.ReactTable.default

let columns = [{
	header: 'First Name',
	accessor: 'firstName'
}, {
	header: 'Last Name',
	id: 'lastName',
	accessor: d => d.lastName,
	show : true,
}, {
	header: 'Age',
	accessor: 'age'
}]

class MyTable extends React.Component {
	constructor(props) {
		super(props)
		
		this.state = {
			cols: props.cols,
		}
	}
	
	onToggleCol() {
		let {cols} = this.state
		cols[1].show = ! cols[1].show
		this.setState({cols: cols})
	}
	
	render() {
		return (
			<div>
				<h1>React-Table - Example with dynamically shown column</h1>
[...]

PR #265, opened 2017-05-12
Fixed issue #210 - Ignore filters on hidden columns.

theCmaker:
I just added a little change. As a column may be hidden, we must check it is not `undefined` before trying to access the `id` prop. This check was made in commit 6401786, but some changes made it appear a few lines after accessing the `id` prop in v6.0.0.

### DIFF
--- a/src/methods.js
+++ b/src/methods.js
@@ -323,17 +323,7 @@ export default Base => class extends Base {
             (row) => {
               let column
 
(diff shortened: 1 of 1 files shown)


tannerlinsley:
@aaronschwartz What say ye?

aaronschwartz:
Looks good to me. Merge it!

PR #268, opened 2017-05-12
Fix default sort algorithm

aaronschwartz:
Fix comparing rows with strings and null values.
Fix checking if type is string.

### DIFF
--- a/src/defaultProps.js
+++ b/src/defaultProps.js
@@ -32,11 +32,11 @@ export default {
   },
   defaultSortMethod: (a, b) => {
     // force null and undefined to the bottom
-    a = (a === null || a === undefined) ? -Infinity : a
-    b = (b === null || b === undefined) ? -Infinity : b
+    a = (a === null || a === undefined) ? '' : a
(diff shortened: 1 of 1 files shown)


src/defaultProps.js
```diff
@@ -32,11 +32,11 @@ export default {
...
     // force null and undefined to the bottom
-    a = (a === null || a === undefined) ? -Infinity : a
-    b = (b === null || b === undefined) ? -Infinity : b
+    a = (a === null || a === undefined) ? '' : a
+    b = (b === null || b === undefined) ? '' : b
     // force any string values to lowercase
-    a = a === 'string' ? a.toLowerCase() : a
-    b = b === 'string' ? b.toLowerCase() : b
+    a = typeof a === 'string' ? a.toLowerCase() : a
+    b = typeof b === 'string' ? b.toLowerCase() : b
```
tannerlinsley: Wow, I wonder if this was a regression that I made on accident.  Very obvious failure on my part.

PR #283, opened 2017-05-18
Top and Bottom Pagination

mayank23:
- Top and Bottom Pagination, defaults to only the bottom pager being shown.
- Included a React Story for an example
- bumped react-story:  "react-story": "^0.0.10"

### DIFF
--- a/docs/src/stories/EditableTable.js
+++ b/docs/src/stories/EditableTable.js
@@ -51,7 +51,10 @@ class Story extends React.PureComponent {
           columns={this.columns}
           defaultPageSize={2}
           showPageSizeOptions={false}
-          showPagination={false}
+          showPagination={{
+            top: false,
+            bottom: false
+          }}
         />
       </div>
     )
--- a/docs/src/stories/EditableTable.js
+++ b/docs/src/stories/EditableTable.js
@@ -51,10 +51,7 @@ class Story extends React.PureComponent {
           columns={this.columns}
           defaultPageSize={2}
           showPageSizeOptions={false}
-          showPagination={{
-            top: false,
-            bottom: false
-          }}
+          showPagination={false}
         />
       </div>
     )
--- a/docs/src/stories/OneHundredKRows.js
+++ b/docs/src/stories/OneHundredKRows.js
@@ -9,8 +9,8 @@ class Story extends React.PureComponent {
   render () {
     const data = _.map(_.range(100000), d => {
       return {
-        firstName: namor.generate({ words: 1, numLen: 0 }),
-        lastName: namor.generate({ words: 1, numLen: 0 }),
+        firstName: namor.generate({ words: 1, numbers: 0 }),
+        lastName: namor.generate({ words: 1, numbers: 0 }),
         age: Math.floor(Math.random() * 30),
         visits: Math.floor(Math.random() * 100)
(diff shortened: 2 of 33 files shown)


mayank23:
Ready for review now

mayank23:
@tannerlinsley any update for this PR? Thank you

docs/src/stories/EditableTable.js
```diff
@@ -51,7 +51,10 @@ class Story extends React.PureComponent {
           columns={this.columns}
           defaultPageSize={2}
           showPageSizeOptions={false}
-          showPagination={false}
+          showPagination={{
```
tannerlinsley: I love the idea of allowing both top and bottom pagination, but I'm not a big fan of this syntax. Nested properties in "props" usually introduce weirdness with change detection.  I would like to see an API more like this:

```
showPagination={true}
paginationTop={true} // New prop
paginationBottom={true} // New prop

Comment (tannerlinsley) [inline docs/src/stories/EditableTable.js]: I'm open to other ideas. @aaronschwartz What do you think?

File: docs/src/stories/OneHundredKRows.js [hunk 2]
```diff
@@ -61,7 +61,10 @@ class Story extends React.PureComponent {
                     data={data}
                     columns={columns}
                     defaultPageSize={3}
-                    showPagination={false}
+                    showPagination={{
```
tannerlinsley: Please remove the custom pagination settings from all stories except for the pagination story meant to demo it :)

src/index.js
```diff
@@ -779,6 +779,21 @@ export default class ReactTable extends Methods(Lifecycle(Component)) {
         }}
         {...rootProps.rest}
       >
+        {showPagination.top ? (
```
tannerlinsley: This JSX block should only be produced once, and used twice, instead of created twice

docs/src/stories/EditableTable.js
aaronschwartz: I agree I like it flatter.  Your example seems reasonable.

mayank23: Yeah, that sounds better. Will flatten it

src/index.js
mayank23: Oops, will fix

docs/src/stories/OneHundredKRows.js
mayank23: 👍

src/index.js
mayank23: Looking now, I pass the prop, "placement",  to the Pagination Component in order to append either "topPagination" or "bottomPagination" to the className of the root div of the PaginationComponent. That way it's easier to style/select the pagination component.

mayank23: I can wrap the top and bottom pagination components with a div in index.js, so then we can only create it once.

mayank23:
there is an issue with react-story that I created, react-router-dom@next doesn't exist anymore

mayank23:
@tannerlinsley updated the PR based on the feedback

tannerlinsley:
@aaronschwartz I'm good on this.  Are you?

mayank23:
@tannerlinsley would you be able to merge it soon? Thanks a lot!

Issue #130, opened 2017-03-16
Improve accessibility


denisw:
Currently, the default react-table components output just nested `<div>` tags, which means that screen readers and other assistive technologies don't recognize that the markup actually represents a table. Thus, the different parts should have [WAI-ARIA](https://www.w3.org/TR/wai-aria) `role` attributes to expose this information. Relevant here is the [grid](https://www.w3.org/TR/wai-aria/roles#grid) role and its accompanying roles (`row`, `gridcell` etc.).

tannerlinsley:
I would love to have more aria compliance, so anyone who would like to work on this, let me know.

PR #293, opened 2017-05-25
Add an option to hide the pagination if there is no data

cdtinney:
This will take priority over the default showPagination prop. It is set to a default of 'true'.

It is demonstrated in a 'ShowPaginationNoData' story.

Relates to #130, #248. The latter had some valid reasons on why this behavior should not be the default, so I added it as an option.

### DIFF
--- a/README.md
+++ b/README.md
@@ -155,6 +155,7 @@ These are all of the available props (and their default values) for the main `<R
   data: [],
   loading: false,
   showPagination: true,
+  showPaginationNoData: true,
   showPageSizeOptions: true,
   pageSizeOptions: [5, 10, 20, 25, 50, 100],
   defaultPageSize: 20,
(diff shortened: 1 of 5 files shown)


tannerlinsley:
@aaronschwartz Thoughts?

aaronschwartz:
Hmm I'm not sure if the property is a little confusing. You could technically just override the entire PaginationComponent to achieve the same thing. I know that is not as easy but I'm not sure how many people want this use case. It also feels like you could use the handlers to just take control of `showPagination` but I'm not sure if it is feasible or not.

I'm alright either way. Do you prefer `showPaginationNoData` or `hidePaginationWhenNoData`?

tannerlinsley:
I still think it could be inferred without the property somehow

tannerlinsley:
This is an edge case that still needs to be addressed, but I think we should be inferring the display of the NoData component instead of forcing the user to set it manually.  Let's explore some ways to do that :)

PR #295, opened 2017-05-26
Add minRows to list of available props

ldstein:
Hi, this is a feature suggestion to toggle the rendering of padded rows.

This PR adds a new prop `showPaddedRows`. It's useful if you want the height of the table to be determined by the number of visible, populated rows.

We found it a bit more performant than setting `style:{display:none}` on empty rows via `getTrProps` (particularly when iterating through small data sets with large defaultPageSizes).

Included is a story which demonstrates the behaviour of the prop.


tannerlinsley:
I feel like this is unnecessary with the `minRows` prop, which you can set to `0`.  Does this achieve something different than that?

ldstein:
Ha, too true! Looks like I missed that prop before venturing down this rabbit hole.

Suggest adding `minRows` to the props section of `readme.md`. It's currently not listed as one of the available props.

tannerlinsley:
Submit a pr? ;)

ldstein:
Updated this PR.

tannerlinsley:
Putting this as a default prop makes things kinda wonky for me. I think the docs should be enough in their own

PR #301, opened 2017-05-29
fix IE select visual bug

stefaniapavel:
This fix is for a IE bug regarding the select element. On IE, the default browser arrow was still being shown due to a appearance property bug.

image


tannerlinsley:
Nice work Stephania!

PR #352, opened 2017-06-20
Update defaultProps.js

kieronsutton00:
A more reliable defaultSortMethod. Works with uppercase/lowercase, negative to positive numbers, etc.


aaronschwartz:
How does it work for strings? If I do "abc"-"def" in my console I get NaN.

Also where does it put null and undefined's?

tannerlinsley:
This doesn't properly handle the ordering of `null`'s and `undefined`'s, which would be a clear regression.

Issue #373, opened 2017-06-27
How to sort table without first row in React-table ?


olejka91:
Are there anyways to pin a row in a table and to sort only data under that row. 

I want to step back into previous folder and I made a node to quit, but when I sort data, this node sorts with data. How to avoid this and to pin it to the top of the table?

aaronschwartz:
This could get tricky. My best guess would be to modify the `accessor` on the column your sorting by to include some identifier if it is a pinned row. Maybe you could change it to something like 

`accessor: (x)=>`x.filename:${shouldbepinned}`

Then you'll need to provide a custom `sortMethod` that parses the `shouldbepinned` value and makes sure that those are always sorted to the beginning of the list.

olejka91:
Thank you so much, but I found another way) I decided to add
QuitArrowButton as a separate component

С уважением,
Олег Волков.

flood4life:
I'm facing the same problem.
@aaronschwartz not sure if it's possible. `sortMethod` defines the _relation_ between elements, e.g. it says if `a` is lesser, equal or greater than `b`. So it is possible to make the pinned item either lesser or greater than any other item. Therefore, it will be on either the top or the bottom of the table. It's impossible to tell the sort order from within the `sortMethod`.

aaronschwartz:
@flood4life You're right. If you need that level of sorting then I think you need to do the following:
* Disable build in sorting on the column with the `sortable` prop
* Add an `onClick` callback to the column header
* In the callback, sort the `data` being passed to the table however you like

flood4life:
@aaronschwartz this seems like doing too much work that has already been done. Maybe a better way is to mock server-side data (https://react-table.js.org/#/story/server-side-data)? That way it is possible to obtain full control over sorting logic.

aaronschwartz:
You could do that too. But then you need to take control of the paging and filtering as well. This isn't a use case many people asked for. If you have an elegant way to implement it in the table I'm sure it would be considered for a PR.

tannerlinsley:
Right. `manual` mode would be better suited for this. Or, in another thread we could easily add the sort direction to the params that get passed to custom sort methods here: https://github.com/tannerlinsley/react-table/blob/master/src/methods.js#L396-L400. That would make it possible for you to infer the visual "top" of the list and keep some records at the top regardless. If anyone would like to PR that, open it up!

flood4life:
[...]

PR #383, opened 2017-07-10
Make sort order available in `sortMethod` (refs #373)

flood4life:
See #373 for context. I also took the liberty to add an example of using the `dir` param to the CustomSorting Story.


tannerlinsley:
The sort direction is technically not handled this way. It is handled by the `_.orderBy` utility function which accepts an array of directions. We're currently reassessing our options for taking care of a circumstance like #373. Stay tuned.

Issue #387, opened 2017-07-13
Resized state - Handling columns with auto widths?


ldstein:
## Problem Description

Hi all,

Assume a table with three fixed width columns and one column of variable width which uses up any remaining available width:

```
[
	{Header: 'First Name',   accessor: 'firstName' , width:100},
	{Header: 'Last Name' ,   accessor: 'lastName'             },
	{Header: 'Age'       ,   accessor: 'age'       , width:50 },
	{Header: 'Registered',   accessor: 'registered', width:50 }	
]
```
(I'm going to use the phrase 'auto-width' throughout this message. What I mean by that is a column with no `width` property defined.)

I'd like to persist the column width state so that it is restored across sessions. This is easily achieved via `onResizedChanged`  and `resized` properties:

```
function handleResized(resizedState){
	localStorage.setItem('resizedState', JSON.stringify(resizedState));
}

function getPersistedResizedState(){
	return JSON.parse(localStorage.getItem('resizedState'));
}
```

Now we come to where I'm stuck - I'd like to add an option to reset columns modified by the user back to their default state. One approach I considered was to iterate through the columns and generate a default `resized` state:

```
function getDefaultResizedState(){
	var state = [];
	columns.forEach(function(column){
		if (column.width) state.push({id:column.accessor, value:column.width});
	});	
	return state;	
}
```

Which actually works quite well... except for the column with no defined width. If the user has resized a column with no width defined, I'm unable to see a way to set the column's width back to 'auto-width'. 

I've tried setting the width to values such as 0, -1, false, null, undefined, as this would make the effect achievable, but it appears only numbers are accepted.

I should confirm my current approach actually works, but requires a reload of the browser window. I'm looking at ways of achieving this result without the reload.

Any thoughts or other methods I could consider to programmatically set the width back to 'auto-width' mode?

## Code Snippet(s) or Codepen
See https://codepen.io/anon/pen/XgOjLw?editors=0011

## Steps to Reproduce
1. Load https://codepen.io/anon/pen/XgOjLw?editors=0011
2. Resize column 1 and make it bigger.
3. Resize column 2 and make it smaller.
4. Click 'Clear Persisted State'
5. Column 2 is unable to be reset back to a auto-width mode.

## System Information
* **Browser & Browser Version:** Chrome, 64bit
* **Node Version:**na
* **OS Version:**Windows 10 x64
* **NPM Version:**na
* **Yarn Version:**na

tannerlinsley:
[...]

PR #414, opened 2017-08-02
Fix invalid maxWidth style

ldstein:
This PR is related to issues described in #387

When a cell's maxWidth is NaN, the calculated style.maxWidth is `NaNpx`. React won't update the DOM's maxWidth style when given this value, resulting in incorrectly sized columns.

This PR ensures the maxWidth style is properly removed in cases where maxWidth is NaN.


ldstein:
Closed due to conflicts.

PR #429, opened 2017-08-09
Use `isResizable` value in ThComponent classname

jeffreypriebe:
Pretty tiny tweak, but I presume this is the "expected behavior".

Currently, passing `resizable` will add or not add the resizing components (index L354). But the `'rt-resizable-header'` classname is always added.

This change will only add the `'rt-resizable-header'` if `resizable` is `true`.

Similarly on in child div inside the `<ThComponent>`.


tannerlinsley:
Nice!

PR #432, opened 2017-08-10
readme-updated: Fixed incorrect property names

colwynmyself:
Parts of the README.md file had incorrect properties when talking about fully managed state. This just fixes those.


tannerlinsley:
Thanks!

PR #440, opened 2017-08-17
Fixed documentation: was referencing expandedRows props for ReactTabl…

bellp:
…e, but it was renamed to expanded


Tsvetkovv:
@bellp it seems duplicate of #432

tannerlinsley:
Thanks!

PR #446, opened 2017-08-22
Make the css for input and select more specific

s7dhansh:
The current css selectors, eg. `.ReactTable input` cause conflicts with `input` elements inside Cells. So changed it to `.ReactTable .-pagination input`.

### DIFF
--- a/src/index.styl
+++ b/src/index.styl
@@ -195,6 +195,16 @@ $expandSize = 7px
     box-shadow: 0 0px 15px 0px alpha(black, .1)
     border-top: 2px solid alpha(black, .1)
 
+    input
+    select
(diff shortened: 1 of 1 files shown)


tannerlinsley:
Thanks!

s7dhansh:
@tannerlinsley can you kindly release a new version with this?

viljark:
What about the input styles in the filter? With this change they are unstyled.
See: https://codesandbox.io/s/r40ovqp5jq

s7dhansh:
@viljark thank you for informing. I will put another PR on Monday. Will have to re-read `less` to make the proper change.

s7dhansh:
https://github.com/react-tools/react-table/pull/537 should fix it.

PR #456, opened 2017-08-25
Fixed README's Fully Controlled Component filtered

rimselis:
In README's Fully Controlled Component section current filters model is assigned to `filters` property which does not work. It should be assigned to `filtered`.


tannerlinsley:
Thanks!

PR #459, opened 2017-08-28
make rt-tfoot flex behaviour the same as rt-thead

deoxxa:
This brings `rt-tfoot` into line with `rt-thead` in terms of flex behaviour.
Without this, it's possible for an `rt-tr` to be taller than its containing
`rt-tfoot` without causing it to expand, which causes the outer `rt-table` to
scroll vertically.


tannerlinsley:
Thanks!

PR #471, opened 2017-08-31
Fix: no extra checking for toggleSort

Tsvetkovv:
I think the extra check doesn't make sense


tannerlinsley:
This is a pretty standard check to make sure that the function is defined.

Tsvetkovv:
@tannerlinsley if you pass undefined, nothing happened. React always check it, elseways

Issue #516, opened 2017-09-28
Number.isNaN not supported in IE


Fanneyyy:
Not sure about how compatible this package is with IE but I got this error when running my app on IE 11:
`Object doesn't support property or method 'isNaN'`

Noticed that the function `src/utils.js/asPx` is using the `Number.isNaN` method (Not supported in IE).
Wondering if it could be switched out with something like
```javascript
function isNumeric(n) {
  return !isNaN(parseFloat(n)) && isFinite(n);
}
```
Many thanks for an awesome table :)

Fanneyyy:
https://github.com/react-tools/react-table/pull/517
Made the changes I believe should solve the problem without affecting other funtionality, if you’re interested :)

tannerlinsley:
Yeah we suggest using an es6 polyfill when using ie. We would like to keep any browser specific code out of the library as much as possible.

PR #517, opened 2017-09-28
IE compatible (#516)

Fanneyyy:
* Switched Number.isNaN function for an IE compatible helper function


Fanneyyy:
Removed in favour of polyfilling :)
https://babeljs.io/docs/usage/polyfill

PR #518, opened 2017-09-29
Update README.md

alexwoll:
I'm guessing this shouldn't have been left here, or it needs to be updated with the whole description :)


tannerlinsley:
Already fixed in a different PR. Thanks!

Issue #476, opened 2017-09-03
Custom sorting that depends on direction


nicodjimenez:
## What version of React-Table are you using?
6.5.3

## What bug are you experiencing, or what feature are you proposing?
I want to put all the rows with undefined values at the bottom, regardless of the sorting direction.  Is this feature supported? 

Thanks for the lovely library!

tannerlinsley:
Yes it is. You can override the sorting function for both individual
columns and the entire table

nicodjimenez:
Ok I get that, what I'm asking is how to override the sorting function for both ascending order as well as descending order, regardless of whether that's for a single column or the entire table.  The purpose being putting null elements on the bottom regardless of if you're sorting in ascending order or in descending order.

flood4life:
It looks like it is possible for filtering via `filterAll` but not possible for sorting. @tannerlinsley is there any particular reason a similar option is not available for sorting? Perhaps a PR adding it would be welcome?

bloomt:
I am looking for this exact same feature. Any updates?

nicodjimenez:
@tannerlinsley I'm happy to make a PR for this if you're OK with that, also a few pointers before I dig into the code would be welcome.

gary-menzel:
There's a problem I can see with changing how react-table works right now. At present, the "order" is not provided as react-table itself determines how to treat the return value from the sort function based on the internal order state. Providing the order in the call and then changing the algorithm to suit would then confuse the code inside react-table. For those of us who are relying on that default behaviour, it would break things.

tannerlinsley:
Gary is correct. You can override the default sorting method for both the entire table and individual columns. This should suffice.

flood4life:
@tannerlinsley I'm afraid you're missing the point here. It is impossible to define a relation between two objects that would make some group of objects (e.g., `null` objects) either greater or lesser than any other object depending on the direction (because sorting method doesn't know anything about direction). Hence, it is impossible to always place `null` objects at the top (or the bottom) of the sorted result. 
So, overriding default sorting method doesn't suffice.

simon-bond:
This is an issue for me too. It's a very small change to the code to pass the sort direction, and it doesn't change the default behaviour.

simon-bond:
Example usage that always sorts 0 to the bottom:

```javascript
sortMethod: (a, b, desc) => {
  a = (a === 0) ? (desc ? -Infinity : Infinity) : a;
[...]

PR #533, opened 2017-10-04
Supply sort direction to custom sort methods (#476)

simon-bond:
It's sometimes desirable to modify your sort method based on the direction of sort, for example in order to always sort null/undefined to the bottom of the table. This can be achieved simply by passing the direction of sort to the sort method, and doesn't change the default behaviour. The docs had a curious reference to "dir", which might have been intended for this, but the code doesn't reflect that.


edi9999:
Any update on this @tannerlinsley ?

gary-menzel:
This should be addressed in V 6.7.4

Issue #485, opened 2017-09-12
[Feature] Pass sorting state in props to Header and ThComponent


simonedavico:
## What version of React-Table are you using?
Version 6.5.3

## What bug are you experiencing, or what feature are you proposing?
Feature proposal: pass sorting state in props to `Header` and `ThComponent`.
This would allow to implement conditional rendering for the header depending on the column's sorting state.

## What are the steps to reproduce the issue?
No issue to reproduce. Would you ben open to a pull request implementing this change?

gary-menzel:
If we were to pass the sorted data to the head components then it would also make sense to provide it to address #457 ?

simonedavico:
In the meantime, if someone wants to get sorting state in the column header of the table, it can be achieved by providing a custom `getTheadThProps`:

```javascript
export const getTheadThProps = ({ sorted }, rowInfo, column) => ({
  sorted: sorted.find(col => col.id === column.id)
});
```
I use it in a custom `ColumnHeader`:

```javascript
const ColumnHeader = ({ toggleSort, className, children, sorted, ...rest }) => (
  // render logic leverages sorted props
)

ColumnHeader.propTypes = {
  sorted: PropTypes.shape({
    sorted: PropTypes.bool,
    id: PropTypes.string
  }).isRequired,
  // more propTypes
};
```

and finally

```javascript
<ReactTable
    ThComponent={ColumnHeader}
    getTheadThProps={getTheadThProps}
    // more props
/>
```

gary-menzel:
V 6.7.4 addresses a number of PRs - was one submitted for this and is it now in that version?

If so, this issue can be closed. Otherwise, maybe someone would like to submit a PR for a suggested change.

simonedavico:
@gary-menzel is there a changelog somewhere? I wanted to check the additions to v6.7.4, but the changelog file is out of date. Nevertheless, if this is not implemented yet and you think it could be useful I can work on a PR, just let me know :)

gary-menzel:
I don't think what you want was submitted in any PR.  The best place to look at the changes is from the Pull Requests item (look at the closed ones and sort by recently updated). There is no automated change log - but if there were, it would just be pulled from the PRs.

Basically, if it wasn't in a PR then it hasn't been implemented as tanner is up to his neck in changes for react-static at present.

PRs seem to take between 2-3 weeks to get merged at present and usually when there are enough of them to warrant all the deployment work (updating npm, etc.).

simonedavico:
[...]

PR #589, opened 2017-10-31
Passes sorting state as props to Th components

simonedavico:
This pull request implements #485, by passing to every column header component its sorting state as a `sorted` prop in the default implementation of `getTheadThProps`. It is easier to render column headers depending on sorting. For example, augmenting the default `ThComponent`:

```javascript
const getIconClass = sorted => {
  if (sorted.desc === undefined) {
    return '';
  } else if (sorted.desc) {
    return '-desc';
  }
  return '-asc';
};

const ColumnHeader = ({ toggleSort, className, children, sorted, ...rest }) => (
  <div
    role='button'
    className={classNames(className, 'rt-th', 'wd-th')}
    onClick={e => {
      if (toggleSort) { toggleSort(e) }
    }}
    {...rest}
  >
    <div
      className={classNames('th-cell', { 'th-sorted': sorted.desc !== undefined })}
    >
      {children}
    </div>
    <div
      className={classNames('th-sort', { 'th-sorted': sorted.desc !== undefined })}
    >
      {
        <Icon
          name={
            `sort${getIconClass(sorted)}`
          }
          size='tiny'
        />
      }
    </div>
  </div>
);
```

The README is updated accordingly.

### DIFF
--- a/README.md
+++ b/README.md
@@ -245,7 +245,9 @@ These are all of the available props (and their default values) for the main `<R
   getTheadGroupThProps: () => ({}),
   getTheadProps: () => ({}),
   getTheadTrProps: () => ({}),
-  getTheadThProps: () => ({}),
+  getTheadThProps: ({ sorted }, rowInfo, column) => ({
+    sorted: sorted.find(col => col.id === column.id)
+  }),
   getTheadFilterProps: () => ({}),
   getTheadFilterTrProps: () => ({}),
   getTheadFilterThProps: () => ({}),
--- a/src/defaultProps.js
+++ b/src/defaultProps.js
@@ -97,7 +97,9 @@ export default {
   getTheadGroupThProps: emptyObj,
   getTheadProps: emptyObj,
   getTheadTrProps: emptyObj,
-  getTheadThProps: emptyObj,
+  getTheadThProps: ({ sorted }, rowInfo, column) => ({
+    sorted: sorted.find(col => col.id === column.id)
+  }),
   getTheadFilterProps: emptyObj,
   getTheadFilterTrProps: emptyObj,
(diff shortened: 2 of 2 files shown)


gary-menzel:
@simonedavico I've done a review of the approach here and am concerned the changes are modifying the `defaultProps`. If you look at all the other defaults for the `get...Props` they all start as returning an `emptyObj` arrow function (which is basically a `null` function).

The main purpose for these is in writing HOC type extensions so they should remain "clean".

In looking at the code, there is a line in the `makeHeader` function (index.js) that grabs the sort information and then uses that to set the CSS. This same `sort` variable could be added to the `rest` object and would then be passed in as a `prop` to the `ThComponent` (without having to touch the defaultProps). That same `sort` variable could also be passed as a third parameter to the `column.Header` component (so that people could get that information in the `Header` on a specific column without having to override `ThComponent`.

NOTE: To keep it in line with the existing code, the new parameter would be called `sort` (rather than `sorted` - which has been used internally to mean the full set of what is sorted rather than just one parameter of the sort).

The end result would be that `ThComponent` would still get the sort information (as would the `Header`) and the defaultProps would stay "clean".

simonedavico:
Thanks Gary, I'll look into your suggestions and rethink my approach as soon as I can.

tannerlinsley:
Going to close this for now.  Reopen if you want to revisit this :)

jadengore:
@tannerlinsley Willing to rework this if @simonedavico is busy

gary-menzel:
@jadengore Happy to accept solid PRs if you are able to put one together.  The idea is good but just needs to fit the suggestions above to be more aligned with how the code is generally written.

PR #609, opened 2017-11-12
Example HOCs for TreeTable and CheckboxTable

gary-menzel:
I've added some example HOCs to the docs sub-project.

As discussed with Tanner, it is not easy at present to integrate with the codesandbox.io examples as they are not under source control.

So, I have added an examples directory and then appended the two examples to the end of the react-story menu but they run in place and can't be modified like the codesandbox.io examples.

I have put two examples up on codesandbox.io and could have referenced these but that is the splitting the samples away from a central point.

**_NOTE: this also includes a PR from @frankolson with regard to the linter work he was doing but there is still a lot of linter cleanup to do that would be best done as a separate undertaking (i.e. just lint cleanup and no functionality changes)._**


tannerlinsley:
Looked like your missing shortid as a dependency?

gary-menzel:
I've added the dependency and pushed it to my forked master. I am still largely a novice on "team" github and was surprised to see it knew about the change without having to put in another PR (yet I see it says that is possible above).

PR #619, opened 2017-11-15
Examples Refactor + multiSort flag

gary-menzel:
I've refactored the Examples so they point at my codesandbox samples for using the HOCs (although, I really need to get the build done so I can remove the temporary HOC files in the examples).

I've also added a README in the HOC folder that I have added.

I have also added a `multiSort` flag that defaults to `true`. If it is set to `false` then it basically disables the `additive` sorting. This was a feature request for someone doing server-side sorting where they don't want to do multi-sort.

[I'm not quite sure (yet) why this can't auto merge.]


gary-menzel:
Maybe I needed to do a PULL before I did the PUSH.  I've fixed the conflicts though,

PR #644, opened 2017-11-26
Fix for 'undefined' in compactObject

gary-menzel:


gary-menzel:
@frankolson While there were a lot things that could have been done better (from my side) regarding this PR, there seems to have been an introduced problem with the `ThComponent` having a `role` and `tabIndex` added (which are now causing TH elements to get focus (which isn't really supported in ReactTable - and does cause it to look a little ugly).... Here is the file that is causing the problem:

https://github.com/react-tools/react-table/pull/644/commits/a6036004afba997114c20daa0c6ae4b319b2e679

Was that purely related to the linting - or was there some other purpose ?

I think the `tabIndex={0}` should be `tabIndex={-1}` (if it is to retain the previous behaviour - or they could just both be removed with a eslint switch added to ignore them).

gary-menzel:
NOTE: I have made a change in my local that removed the `tabIndex` and changes the `role` to `role="header"` and that seems to resolve the problem.

PR #663, opened 2017-12-04
Ensure updating columns triggers a re-render

### LINKED ISSUE #662: Updating the columns prop doesn't cause a re-render
## What version of React-Table are you using?
6.6.0

## What bug are you experiencing, or what feature are you proposing?
Updating the columns prop doesn't cause a re-render

## example
https://codesandbox.io/s/2j0kz92y6j

2 buttons have been added to (naively) demonstrate the issue:

Button 1 doesn't work but should, Button 2 forces a re-render by adding an int key prop and incrementing when changing the columns. 

##
suggested fix is to make the componentWillReceiveProps comparison of old columns vs new columns a deep comparison. (I'll create a pull request with a fix shortly).

jontansey:
Add a deep comparison of the columns prop (using `JSON.stringify()`) to the componentWillReceiveProps function.

fixes #662

### DIFF
--- a/src/lifecycle.js
+++ b/src/lifecycle.js
@@ -43,7 +43,7 @@ export default Base =>
       // Props that trigger a data update
       if (
         oldState.data !== newState.data ||
-        oldState.columns !== newState.columns ||
+        JSON.stringify(oldState.columns) !== JSON.stringify(newState.columns) ||
         oldState.pivotBy !== newState.pivotBy ||
         oldState.sorted !== newState.sorted ||
         oldState.filtered !== newState.filtered


gary-menzel:
@tannerlinsley Can you have a look at this one (happy to discuss on Slack).

tannerlinsley:
Please follow this example to see how to update the columns correctly: https://codesandbox.io/s/o4q0lqqwxy

jontansey:
Thanks for your help @tannerlinsley that makes sense.

debovis:
The moral of the above link is use the follow to update columns correctly

`this.setState({
      columns: [...this.state.columns]
 })`

PR #667, opened 2017-12-06
Add column to getResizerProps

Thimey:
Pass the column to `getResizerProps` callback

Solves https://github.com/react-tools/react-table/issues/666


gary-menzel:
From a code style perspective, I'd probably just move the variable declaration down into the scope where the `column` info is available (rather than destructure the return of the function call - although there is nothing specifically wrong with that, the general style in the code seems to be to use a variable).

Yeah - sometimes it is tricky to know where to hook in.... there is generally a separation between the component (and the props) versus what is rendered in the component (e.g. Header vs. getTheadThProps vs. ThComponent) but I think the basic suggestion of passing the `column` makes sense in this case.

PR #673, opened 2017-12-11
Fix ThComponent classnames ordering in defaultProps

jadengore:
This PR fixes a bug where the `ThComponent's` base class comes after the configurable class name. This prevents a user from customizing the styling of the header without using `!important` in their CSS. The change allows for a custom header classname to extend the base styles.


jadengore:
@gary-menzel Are you the primary reviewer on this project?

gary-menzel:
Probably the only reviewer at the moment.  There are other PRs queued up that are ahead of this one (and my time is limited).  In addition, the repo owner has been away and I usually run these by him before finalising them.

I've taken a quick look and I can't see any major issue with the PR but it just has to come in behind some of the others.

PR #687, opened 2017-12-17
CodeSandbox is breaking on iterations over 5000

kevincolten:
screen shot 2017-12-17 at 3 33 29 pm


gary-menzel:
This was not an issue with ReactTable. It was an issue with codesandbox and has been fixed on their end.

PR #739, opened 2018-01-12
Solves issue 553

piyuesh-dev:
Corrected custom filtering example on CodeSandbox, and use that new example link in App.js. This solves issue 553.


gary-menzel:
Unfortunately, there is no agreed way to update the code sandbox examples. They are notionally controlled by the owner. @tannerlinsley for review.

gary-menzel:
This PR will not be merged as it splits the samples away from the maintainers of the codebase. Alternatives for sharing the examples in codesandbox are being considered.

Issue #634, opened 2017-11-21
Accessible Table using Keyboard


mghuy:
## What version of React-Table are you using?
Latest Version (v6.7.4)

## What bug are you experiencing, or what feature are you proposing?
Keyboard navigation is currently non-existent to the rows of data that are expandable. While users who can use a mouse can easily access this collapsible feature, those using the tab button for navigation cannot. I plan to resolve this issue by adding tab features to expandable rows in compliance with WAI-ARIA accessibility standards.

gary-menzel:
When you have the update ready, please feel free to submit the PR for review.

gary-menzel:
@mghuy Happy to accept any PR you can prepare in regard to this but will close the issue for now. Please re-open once the PR is ready.

PR #758, opened 2018-01-18
Implemented basic screen reader support.

dumptruckman:
This adds basic screen reader support to tables that are otherwise invisible to a screen reader. Ultimately I would like to add full accessibility support as described in WAI-ARIA 1.1 specifications.

### DIFF
--- a/src/defaultProps.js
+++ b/src/defaultProps.js
@@ -175,24 +175,58 @@ export default {
   rowsText: 'rows',
 
   // Components
-  TableComponent: _.makeTemplateComponent('rt-table', 'Table'),
+  TableComponent: ({ children, className, ...rest }) => (
+    <div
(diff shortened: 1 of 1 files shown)


dumptruckman:
I fixed the test issues here. I used the `columnheader` role exclusively for the ThComponent because I couldn't find any instances of ThComponent being used as a row header. If that were to change in the future we'll have to have some other kind of solution I guess.

gary-menzel:
Please note that (previously) using a tabIndex caused a focus box to appear when it shouldn't.

Just wanted to make sure you had checked that.

dumptruckman:
When would the focus occur? When I tab through the Tester page, focuses the React Table header image followed by the page number at the bottom.

mikaello:
This PR relates to issue #130 and #634

Issue #808, opened 2018-02-15
[Feature] Manual firing fetchData


MAKARD:
Manual fetching data may be useful in interactive tables. For example, data on server is updating every 1 minute and its will be good to not  unmount(or change sort/page/filter etc) table for fetch this data.

gary-menzel:
You should fetch that data outside of ReactTable.  `onFetchData` is not really meant for real-time updates.  It is designed to let you know when the USER has taken an action that requires you to get new data - not for when the server has added data and you want to refresh ReactTable.

State management tools like Redux or MobX are meant for that type of work.

MAKARD:
What about `refresh table` button? User make action in this case.

gary-menzel:
You loaded the data externally and passed it into ReactTable.  You should also reload the data externally and pass it in.

React works by observing when properties change to refresh data.

There is no action that a use can take INSIDE ReactTable to request a reload of data.  It is an external action.  Your data came either from props you passed in or state you initialised in something like `componentDidMount` in the Component that is wrapping ReactTable.  That is where the data should be reloaded.

0777101:
@MAKARD, try use react ref and statefull react component:
```
class TableWithRefresh extends React.Component {
  updateTableData = () => this.table.fireFetchData();
  render() {
    const { tableProps } = this.props;
    return (
      <div>
        <button onClick={this.updateTableData}>Refresh data</button>
        <ReactTable
          {...tableProps}
          ref={(instance) => { this.table = instance; }}
        />
      </div>
    );
  }
}
```
If needed you can also use ref with stateless components, see https://reactjs.org/docs/refs-and-the-dom.html#refs-and-functional-components

MichaelBuen:
The instance can now be accessed from render props https://github.com/react-tools/react-table#functional-rendering

cdaringe:
@gary-menzel 

> You loaded the data externally and passed it into ReactTable. You should also reload the data externally and pass it in.

this isn't fully true in all cases. react-table owns/controls state to.  that is, i passed it `onFetchData={this.fetchData}`, `react-table` owns the stateful data that gets passed in as args to `onFetchData`.  consequently, i can't just load my own data without a bump from react-table sharing its state with me

PR #809, opened 2018-02-15
Add prop to manual fetching data

MAKARD:
#808


gary-menzel:
This PR will not be merged as it does not fit in with the current architecture of ReactTable.

PR #846, opened 2018-03-08
Improve document event binding for colum resizer

phucpnt:
feat(column-resize): improve document event binding for resizer

* Use case: the table is mounted into a `non-src` iframe


gary-menzel:
This PR will not be merged as it has conflicts and is not clear about what functionality is it providing.

phucpnt:
thanks @gary-menzel I think it is fair to do that.

Issue #860, opened 2018-03-15
"No rows found" shown underneath "Loading..." when loading is true and data is undefined


mrubin:
## What version of React-Table are you using?
6.8.0

## What bug are you experiencing, or what feature are you proposing?
My component fetches data from the server and has a `loading` prop (which is set to `true` while we're fetching data and changes to `false` once the data is available). I want to use this existing `loading` prop to set the correct loading state on ReactTable by doing the below:

```
let tableData = loading ? undefined : myData

<ReactTable
  loading={loading}
  data={tableData}
  ...
```

What I see is that when the "Loading..." indicator is shown, underneath it I also see "No rows found" even though I am setting `data={undefined}`. This seems like a bug. When the table is in the loading state, "No rows found" should not be shown.

I also tried not setting the `data` prop at all, but there was no change in behavior:

```
if (loading) {
  <ReactTable
    loading={true}
    ...
} else {
  <ReactTable
    data={myData}
    ...
}
```

maxrosecollins from the Slack channel said they see the same behavior as well.

saichandesh:
Hi @mrubin , I have created a PR ( [https://github.com/react-tools/react-table/pull/861](https://github.com/react-tools/react-table/pull/861) ) to fix this issue. Go through it and let me know if there are any comments on it.

tannerlinsley:
Closing due to issue age. If you think this issue needs to be reopened or if it should be turned into a pull-request, please check the latest version of React-Table for the issue or feature once more and reopen or create a PR if appropriate. Thanks!

Billy-:
@tannerlinsley this is definitely still an issue, a PR was made (#861) but has been closed by @gary-menzel as he doesn't agree that it is an issue. We (the community) are unanimous in our stipulation that this is, in fact, an issue.

The discussion should have been had here, rather than the PR (PR's are for discussing the code, we should have discussed the issue here), but if you could take a look at the conversation over there (#861) and weigh in, I'm sure we would all very much appreciate that.

Kindest regards,

PR #861, opened 2018-03-15
fix(issue - 860) :  fix the 'No rows found' shown underneath 'Loading...' when loading is true and data is undefined

saichandesh:
fix the 'No rows found' shown underneath 'Loading...' when loading is true and data is undefined

### DIFF
--- a/src/index.js
+++ b/src/index.js
@@ -991,7 +991,7 @@ export default class ReactTable extends Methods(Lifecycle(Component)) {
               {pagination}
             </div>
             : null}
-          {!pageRows.length &&
+          {!pageRows.length && !this.props.loading &&
             <NoDataComponent {...noDataProps}>
               {_.normalizeComponent(noDataText)}
             </NoDataComponent>}


mrubin:
@saichandesh Thanks for the quick PR! This looks good.

gary-menzel:
This PR will not be merged as it changes existing behaviour.

Billy-:
@gary-menzel I don't understand your comment on closing this issue - surely changing existing behaviour is not an argument for rejecting a change - the whole point in the change is to change the existing behaviour because it's incorrect?

batjko:
> surely changing existing behaviour is not an argument for rejecting a change

I would agree with this assertion. If the change is rejected because it breaks an API the maintainer wants to maintain as is, that would be an acceptable answer. At least @gary-menzel should clarify his response.

gary-menzel:
Here's my clarification...

First, I don't agree it is a bug. I also don't agree that the only reason to reject something is if it breaks the API. Finally, I am responsible for triaging the Issues and PRs to reduce all the cross-check work that any code change requires.  I had planned to comment on the original Issue but even those are piling up right now and I have to concentrate on the highest value ones.

I realise that doesn't sit well when some people want a change - and it's a tough call - but there you have it.

mrubin:
@gary-menzel Please clarify why you think this is not a bug? Quoting someone above, "The existing behaviour is erroneous - it stipulates "No Data Found", when in fact, the data is still loading." This is very clearly undesirable/confusing behavior.

batjko:
> If it is still "loading" then "No Data Found" is true. I do not see any dichotomy in this state.

That makes no sense. The UX of this is objectively bad, since it does confuse the user. Loading and No Data Found are conceptually two different states, both in theory and in the practice of a user's expectation.

batjko:
> I also don't agree that the only reason to reject something is if it breaks the API.

To be fair, that's not what I said. I just gave an example of what could be a possible reason for the rejection, even though it would still have been wrong from a change management perspective.

However, if we want to preserve the current behavior for some reason and still give people a way of aligning the table with user expectations, @Billy- 's suggestion to introduce a new optional prop would achieve this and probably make everyone happy.
(Although I'd choose a more descriptive name, perhaps `noDataWhileLoading=false` or to be more customizable: `noDataMsg={{ showWhileLoading: false }}` ?)

ronkot:
I kind of worked around this by giving empty `noDataText` prop when being in loading state:

    <ReactTable
      loading={loading}
      data={data}
      noDataText={loading ? '' : 'No results')} />

The visuals are not perfect though. I'd really love to see this PR functionality in `react-table` 🙏

cdock1029:
> First, I don't agree it is a bug. 
> If it is still "loading" then "No Data Found" is true. I do not see any dichotomy in this state.

I completely disagree with the logic here, simply because I dispute what *No Data found* is actually telling you... what it exists for (we could even say, the author's original intent since we're speculating about that..)

It's not saying currently we have no rows to display. You know why? Because the fact there are no rows displaying data is **already telling you that**.

What `No data found` is saying is more: that we've **looked for the data** and didn't find it. If you disagree, then why do we need to show it if we have empty rows?

But let's put that aside and concede that some may like a constant display in the UI that there is currently no data (apart from the already existing instance of that being the empty rows..).

Why on Earth would anyone want **both UI components nearly super imposed on top of each other** as the current implementation results in? 

screen shot 2018-07-27 at 12 45 44 pm


In that case, what information are those UI components providing you? Nothing, because you can't clearly decipher them. It's just noise.

ronkot:
I don't want to insult authors in any way, but I think a little story could enlight why many of us think this PR behavior is preferable. This happened me while ago when I called my bank...

Me: Good day
Bank: Good day, sir. How could I help You?
Me: I'm thinkng to get a new table and I'd like to know my account balance.
Bank: Okay, sir. You have no money on your account.
Me: What? Surely there's something - and how could you know that so fast?
Bank: I'm still checking the balance, but it takes some time. So as long as I don't know exactly, You have no on your account.
Me: But there surely is something!
Bank: Not currently from our point of view, sir.
Me: But...
Bank: Yes, sir?
Me: Nevermind. Could you tell me when you've checked the balance and then tell me how it looks like from your point of view then?
Bank: Sure, sir. Just checked that. You've got 2 cents.
Me: 2 cents? Umm..Okay. I think I've got to stop playing that Boom beach.. Anyway, no new table ahead. Please, keep my 2 cents and invest them well as you always do. After all I think I don't even need a new table, the old is very good actually! (But maybe I could get some pink gems..) Thank you for your help.
Bank: Glad to help, sir!

Billy-:
#860

PR #875, opened 2018-03-26
feat: added optional totalPages prop to React Table to manually set pages value

barrymichaeldoyle:
Implemented a `totalPages` prop on `ReactTable` which allows you to manually set the pages value in the Pagination section of the table (another controlled feature).

This new prop is optional, if it's value is undefined (falsey), then the table will continue to work as normal.

### DIFF
--- a/README.md
+++ b/README.md
@@ -205,6 +205,7 @@ These are all of the available props (and their default values) for the main `<R
   // Controlled State Overrides (see Fully Controlled Component section)
   page: undefined,
   pageSize: undefined,
+  totalPages: undefined,
   sorted: [],
   filtered: [],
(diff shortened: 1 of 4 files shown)


barrymichaeldoyle:
Ok I've confused myself, my bad, this feature is already implemented as the pages prop, ignore this PR.

boss225:
how can i added optional totalPages. 
example server response :
{
  data: [
     {id: 0},{id: 1},{id: 2},{id: 3},{id: 4},{id: 5}
  ],
  pages: { totalPages: 10 }
}
i want show "Page 1 of 10" with 5 rows 5 records but i can't custom totalPages

PR #880, opened 2018-03-28
[POC/WIP] codesandbox auto import

JasonRitchie:
To enable maintenance of the codesandbox examples by simple PRs to this react-table repository. This replaces the unique id per example in the iframe URL with usage of the /embed/github/* endpoint. 

This requires creating a minimal package.json for each example to indicate its dependencies

### DIFF
--- a/docs/src/App.js
+++ b/docs/src/App.js
@@ -4,6 +4,7 @@ import React from 'react'
 //
 import ReactStory, { defaultProps } from 'react-story'
 import CodeSandbox from './CodeSandbox.js'
+import CodeSandboxAutoImport from './CodeSandboxAutoImport.js'
 import './stories/utils/prism.css'
 import '../../react-table.css'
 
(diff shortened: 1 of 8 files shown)


gary-menzel:
So @JasonRitchie, if I understand this right... Ives' new feature on codesandbox allow a sandbox to be updated from from specific code in a github repo.  Your solution would allow the examples to remain part of the source and codesandbox would be able to detect the change in repo and automatically pull the changed examples across - thereby keeping the examples in sync with the source?

Is there any downside or would it all just be seamless to the end-user of the doco site?

If that is correct, it is the solution that @tannerlinsley and I discussed some months back (before Ives had the feature on codesandbox).

gary-menzel:
@JasonRitchie so is that because it pulls from github each time it is hit ?

JasonRitchie:
During testing it was behaving as if it has some form of caching on codesandbox's side. The first load of a particular example would be slow and then it would be much faster.

You can compare for yourself by checking out my branch and running the command to locally dev the documentation. (Was it npm run docs?)

PR #881, opened 2018-03-28
Remove header focus outline.  Fixes react-tools/react-table#685

maccuaa:


oim5nu:
My project relied heavily on react-table. Sticked with 6.7.4 so far.

gary-menzel:
It was fixed and someone "unfixed" it - we are battling between a "clean" look and "accessibility".

maccuaa:
@gary-menzel what is the goal then?  To keep the blue hover outline but not have it overlap the next column?

gary-menzel:
@maccuaa Personally, I prefer the "clean" look. I don't want a blue line at all.  It was introduced accidentally as part of eslint checking and took quite a while to resolve.  ReactTable is FAR from meeting accessibility guidelines (no keyboard support - for example) so I don't think the focus box should be there at all - but I am not focussed on accessibility.

I think the solution is fine - but I'd say this is more of a "fix recommendation" rather than a PR.  The react-table CSS is meant to be copied and hacked to suit the end-user needs.

But, that is really only my opinion.  I've not upgraded since 6.7.6.

victorbueno:
Since I was going to commit the same fix again. 👍

PR #889, opened 2018-03-30
Added react-dom as a peer dependency

shinonomeiro:
Issue: https://github.com/facebook/react/issues/10320

screen shot 2018-03-30 at 17 02 59

I've simply added `react-dom` as a peer dependency, which for some reason wasn't (but `react` is). This did solve the problem for my company's project codebase.

I don't know if some other internal compatibility issues might arise due to specific conditions that I cannot possibly test. I do hope  for full support of React v16 from the authors though.

### DIFF
--- a/package.json
+++ b/package.json
@@ -46,7 +46,8 @@
     "classnames": "^2.2.5"
   },
   "peerDependencies": {
-    "react": "^15.x.x || ^16.x.x"
+    "react": "^15.x.x || ^16.x.x",
+    "react-dom": "^15.x.x || ^16.x.x"
   },
   "devDependencies": {
     "autoprefixer": "^6.7.0",


gary-menzel:
It is not a dependency because it is not required by React-Table.  React-Table is deliberately "minimal dependency".

From the issue you linked, it appears to be more of an issue with recent version changes between React and React-Dom.

shinonomeiro:
It is a dependency, albeit a 'peer' one as obviously React-table would be useless without React. I don't understand why 'react' is listed in `peerDependencies` yet React-dom isn't, although both libraries form a _pair_ and are always updated _simultaneously_. Most other libraries which encountered this very issue solved it the same way I did in this pull request.

Per the `devDependencies` listed in `package.json` (https://github.com/react-tools/react-table/blob/master/package.json#L72), React-dom is locked to v15.4.2 while React itself is allowed to be updated to v16.x.x as listed under `peerDependencies`, which results in a version mismatch.

PR #896, opened 2018-04-05
Added classes to table headers to represent sort order/precedence.

brendonsoto:
Hi! My added code adds classes to sorted table headers to distinguish the primary sort from the secondary, tertiary, etc.
Thought it might be useful to have in the main code base.

### DIFF
--- a/src/index.js
+++ b/src/index.js
@@ -299,6 +299,7 @@ export default class ReactTable extends Methods(Lifecycle(Component)) {
     const makeHeader = (column, i) => {
       const resizedCol = resized.find(x => x.id === column.id) || {}
       const sort = sorted.find(d => d.id === column.id)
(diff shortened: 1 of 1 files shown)


gary-menzel:
The code looks OK.  I am just wondering how it would be used in a practical situation (i.e. when is it necessary to change the style based on the sort hierarchy?). Also, is there another way (through `get*Props` that this could be accomplished for those who require it?

brendonsoto:
Hi. The use case is to tell the hierarchy of multi-sorted columns from a glance (i.e. which is the primary, the secondary, etc.), saving the user from having to examine the data to figure it out.
Ah, I misunderstood `getProps` use then. Managed to get the same result through that. Thanks for pointing that out!

PR #941, opened 2018-05-01
Add defaultPage prop

### LINKED ISSUE #676: Add defaultPage prop
## What version of React-Table are you using?
react-table@6.6.0

## What bug are you experiencing, or what feature are you proposing?
Add property defaultPage

## What are the steps to reproduce the issue?
There are several default properties already defined like defaultPageSize or defaultFiltered, but what I need is a defaultPage prop, it is an small change to your library
I'm working with redux and I need to save the table state in redux and load it when the component is mounted again. All the default props works fine, but the table is always initialised in page 1, and not in the page stored in redux
The change is in `lib/index.js` line 76 
instead of
`page: 0,`
it should be 
`page:  props.defaultPage,`

### LINKED ISSUE #701: Missing 'defaultPage' option
Well, I can't find this apparently obvious option in the documentation of react-table, and I can't figure out why is not implemented. Is there any reason? Just proposing its addition here :)

In one of my tables, I need to start showing the last page independently of its number of rows, but then let the user paginate all over (so the fully controlled option through 'page' is not valid for me).

I think it could be achieved with a 'defaultPage' option (like the existing 'defaultPageSize' for pageSize).
In the meantime... could you suggest me any workaround to face this issue?

Thank you very much.

efimweb:
This closes #676 and closes #701

### DIFF
--- a/src/defaultProps.js
+++ b/src/defaultProps.js
@@ -15,6 +15,7 @@ export default {
   showPaginationBottom: true,
   showPageSizeOptions: true,
   pageSizeOptions: [5, 10, 20, 25, 50, 100],
+  defaultPage: 0,
   defaultPageSize: 20,
   showPageJump: true,
   collapseOnSortingChange: true,
--- a/src/index.js
+++ b/src/index.js
@@ -34,7 +34,7 @@ export default class ReactTable extends Methods(Lifecycle(Component)) {
     this.resizeColumnMoving = this.resizeColumnMoving.bind(this)
 
     this.state = {
-      page: 0,
+      page: props.defaultPage,
       pageSize: props.defaultPageSize,
       sorted: props.defaultSorted,
       expanded: props.defaultExpanded,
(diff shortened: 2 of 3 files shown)


TimNZ:
I manually implemented the two lines of change and find that state.page is reset to 0 due to 'pages' being recalculated in 'setStateWithData()' as per below.

```
        // Calculate pageSize all the time
        if (newResolvedState.sortedData) {
          newResolvedState.pages = newResolvedState.manual ? newResolvedState.pages : Math.ceil(newResolvedState.sortedData.length / newResolvedState.pageSize);
          newResolvedState.page = Math.max(newResolvedState.page >= newResolvedState.pages ? newResolvedState.pages - 1 : newResolvedState.page, 0);
        }
```

I haven't investigated the code much, but this fixes it by doing manual check, consistent with 'pages' above.

```
        newResolvedState.page = newResolvedState.manual ? newResolvedState.page : Math.max(
          newResolvedState.page >= newResolvedState.pages
            ? newResolvedState.pages - 1
            : newResolvedState.page,
          0,
        )
```

TimNZ:
Just realised this module doesn't have any tests...

gary-menzel:
Please note, this is unlikely to get merged as the author is looking into architectural changes for the next major release that is likely to have sweeping changes to the codebase.

TimNZ:
Thanks @tannerlinsley

PR #950, opened 2018-05-06
Don't recalc page if in manual mode

TimNZ:
Add 'manual' check for page recalc to be consistent with pages.
Allows page prop setting to stick when # of pages is unknown.

Enables this:
https://gist.github.com/TimNZ/83631516f3e7a1ca1be33228bc3d56d6


tannerlinsley:
Fixed with a different PR

PR #957, opened 2018-05-08
Fixed text mismatch when doing SSR

hornta:
I tried to render a simple table in next.js but that caused a mismatch on a text element.

I also took the liberty to add a few Windows specific things to make it easier for Windows users to contribute to this project.


roginfarrer:
Just wanted to say I get the same error in SSR contexts. Looking forward to this getting pulled in!

vimutti77:
I don't know if it was fixed yet. I am using react-table v6.8.6 and I got this error in next.js.

> Warning: Text content did not match. Server: "5 rows" Client: "5"

PR #961, opened 2018-05-10
Support for more than 2 levels of headers

mpaupulaire4:
Added support for multiple levels of headers

### DIFF
--- a/src/index.js
+++ b/src/index.js
@@ -125,7 +125,7 @@ export default class ReactTable extends Methods(Lifecycle(Component)) {
       // Data model
       resolvedData,
       allVisibleColumns,
-      headerGroups,
+      headerGroupLayers,
       hasHeaderGroups,
       // Sorted Data
       sortedData,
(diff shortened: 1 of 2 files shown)


jharmer-rms:
This would be really great!

tannerlinsley:
Can we see an example of this?

aflansburg:
Would this allow for usage of a "higher tier" filter? For instance you might have the default text input filter, but above that also have a drop-down filter for specific values.

arttay:
+1ing

aflansburg:
+1ing as well

Fornori:
A feeling that the authors are on vacation; (

mpaupulaire-rms:
@tannerlinsley Would a screen shot suffice?

tannerlinsley:

tannerlinsley:
For realsy's though, this is a huge feature in the next major version. You should see the implementation on a new version branch in github very soon.

PR #1007, opened 2018-06-09
Fix propTypes validation for getFooterProps

rhjones:
The prop type for `getFooterProps` is currently `PropTypes.object`, but `getFooterProps` (like all of the other `getXProps` props) should be a function. Providing a function for this prop works as expected but throws a PropType validation error.

This change updates the type to `PropTypes.func` to match the other `getXProps` props.


bminer:
Can this be backported to v6.8.6, please?

PR #1029, opened 2018-06-20
Add additional column's renderers in TreeTable HOC

kkaczmarczyk:
Pass Expander, PivotValue and Pivot renderers to column builder in TreeTable HOC.


matejstrasek:
Could this merge get tagged?

PR #1051, opened 2018-07-10
Disable the ESLint warnings for the foldable table HOC

MrLeebo:
My other PR #1050 as well as many others have failing CI steps because of eslint errors from this component.

I tried fixing the warnings themselves, but I struggled to test the foldable table HOC because the docs example for it doesn't even use the component in question and the dependent SVGs aren't even included in the npm package. That leads me to conclude that the component is probably not being used and is likely dead code.

### DIFF
--- a/src/hoc/foldableTable/index.js
+++ b/src/hoc/foldableTable/index.js
@@ -1,3 +1,4 @@
+/* eslint-disable */
 import React from 'react';
 import left from './left.svg';
 import right from './right.svg';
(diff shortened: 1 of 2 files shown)


baoduy:
I have the other pull request to fix this issue. 
I have an sample for Fordable table in [codesanbox here](https://codesandbox.io/s/8pkrj5yorl)

tannerlinsley:
Fixed in a different PR

PR #1067, opened 2018-07-23
fixes link for server-side data example

k8k:
the link in the readme to the detailed server-side example currently 404s - this PR points it to the correct location.


sasuomi:
Fixes https://github.com/react-tools/react-table/issues/929

PR #1108, opened 2018-08-31
Feature - Option to display filters above headers

alpha1d3d:
Use Case
---------

The current filtering system, while being pretty robust, is not flexible enough to enable certain UX desires. 

For instance it is not possible to display the filters above the table head in such way:
image

I currently achieved it with some CSS hacks:
```
.ReactTable {
    padding-top: 4em;
}
.ReactTable .rt-thead.-filters {
    position: absolute;
    top: 0;
    border-bottom: none;
    width: 100%;
}
```
But this has proven to be unreliable (if a filter has a dynamic height, it overflows to the table, and not to mention z-index issues with position absolutes.)

Solution
---------
This PR adds a new ReactTable prop `showFiltersAbove`, if set to true, will display the filter rows above the table headers.

And the following becomes possible:
image

Which then can easily be styled to look like:
image

### DIFF
--- a/README.md
+++ b/README.md
@@ -203,6 +203,7 @@ These are all of the available props (and their default values) for the main `<R
   multiSort: true,
   resizable: true,
   filterable: false,
+  showFiltersAbove: false,
   defaultSortDesc: false,
   defaultSorted: [],
   defaultFiltered: [],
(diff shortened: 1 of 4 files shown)


jrock17:
I would love this option as well

tannerlinsley:

alpha1d3d:
👍 Thanks for the work you guys put in this repo. Looking forward to headless components.

PR #1121, opened 2018-09-07
Add callback to subComponent for child components to close itself

ljmerza:
I needed a way to close the expanded component from within the component so added a callback to the subComponent arguments.


tannerlinsley:

PR #1123, opened 2018-09-07
Use pivot value for nesting path

cipacda:
This would be really helpful so we can store expanded row preferences based on row data and not on row index which can change with sorting.


tannerlinsley:
This would break a lot of current examples and users. So we'll slate this for the next major release. You'll see a new branch for it in the near future. Thanks!

PR #1127, opened 2018-09-12
Pagination accessibility

tragessere:
This is a small change to add an 'aria-label' attribute to the page jump input and the page row count selector. It adds a prop for each so that custom text can be passed in for the label


tannerlinsley:

PR #1129, opened 2018-09-15
Fixed defaultProps.selectType.

colshacol:
Changed defaultProps.selectType from "check" to "checkbox".

While working with react-table for my first time (aiming to replace an overly-complicated table solution at work), I found that by default the selectTable HOC component gave me an input for the checkbox, which I found to be quite odd.

Here is a screenshot and a CodeSandbox demonstrating how I ran into this issue:

https://codesandbox.io/s/z3ylko96pp

react-table

I found that defaultProps.selectType (on the selectTable HOC) was set to "check", which is an invalid value to be passed to an `<input>` element in the `type` attribute. I changed it to "checkbox" in this PR.

Here is a screenshot using the correct "checkbox" value:

react-table-0

Let me know if "check" and the text-input behavior was intended. :+1:

### DIFF
--- a/src/hoc/selectTable/index.js
+++ b/src/hoc/selectTable/index.js
@@ -106,7 +106,7 @@ export default Component => {
     toggleAll: () => {
       console.log('No toggleAll handler provided.')
     },
-    selectType: 'check',
+    selectType: 'checkbox',
     SelectInputComponent: defaultSelectInputComponent,
     SelectAllInputComponent: defaultSelectInputComponent,
   }


tannerlinsley:

davidhenley:
How did you fix this?

colshacol:
Honestly, I don't remember anything about this. 😅 3 months is a reallll long time.

colshacol:
@davidhenley If this is merged and published, then this bug will be fixed in the latest version.

PR #1135, opened 2018-09-27
feature: allow minResizeWidth to be set

jannikkeye:
This PR implements the `minResize`-Prop which specifies to what minimum width a column can be resized.

I found [this](https://github.com/react-tools/react-table/issues/536#issuecomment-334904841) issue where @gary-menzel mentions something like this. I found this after I added the functionality.

Please review this PR and suggest changes. As of now the functionality is in place and satisfying my current needs. I would like to see this in the main repo otherwise I will have to use my fork from now on.

### DIFF
--- a/src/defaultProps.js
+++ b/src/defaultProps.js
@@ -134,6 +134,7 @@ export default {
     filterable: undefined, // use table default
     show: true,
     minWidth: 100,
+    minResizeWidth: 11,
     // Cells only
     className: '',
     style: {},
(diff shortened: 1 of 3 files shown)


visshaljagtap:
Issue with minResizeWidth props. I am unable to set minResizeWidth. It's not working.
It will be great if I get solution. Thanks!

jannikkeye:
@visshaljagtap As far as I am aware there has not been a new release after this PR got merged. Are you installing react-table from npm? If so and you want to use this new prop then you should install directly from GitHub. But to be honest I would hold of on it until it either gets released or next major version is around the corner. While I did implement the prop in this PR we decided not to use It until a normal release arrives. Maybe you can do the same 👍

PR #1139, opened 2018-10-01
virtualizedTableHOC 

sgrishchenko:
I tried using your library to create a virtualized table and ran into performance problems. To improve the situation, I added a configuration setting to API of Table Component. This [Sandbox example](https://codesandbox.io/s/zn2q157w2m) shows how the responsiveness of the interface can improve using this setting. 
Before: Slow Virtualized ReactTable
After: Virtualized ReactTable

### DIFF
--- a/docs/package.json
+++ b/docs/package.json
@@ -42,6 +42,7 @@
     "react-script": "^2.0.5",
     "react-scripts": "^0.9.5",
     "react-story": "^0.0.10",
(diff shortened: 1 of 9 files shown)


tannerlinsley:
Wow that is impressive. This is something, however that we will be handling with the next major version of react-table much better by giving you full control of the rendering.

PR #1155, opened 2018-10-13
Updating for foldableTable

Silveress:
Added --copy-files flag to babel, without it the left and right svg files are not copied to the lib directory.

Since you dont plan on pushing to npm for a few months and I have a need for foldableTable I decided to use the git version and compile it myself.
Came into an issue where the svg's for the icons were not in the lib folder.
Adding the above flag fixes this.


tannerlinsley:
This fix has been added via another pull request. Thanks for the help!

PR #1176, opened 2018-10-28
Extract pagination elements renderers to pagination props

Gelio:
Pagination elements are now customizable without having to create a new `Pagination` component, that would mostly mimic the existing one in terms of functionality.

I have updated the README to include the newly added props.


Gelio:
The build failing is not my fault. I have not touched the files with errors.

PR #1192, opened 2018-11-09
Update README.md - add className to getProps methods

gazdagergo:
Returning any className seems to be required from any get...Props function, so I suggest to add it in readme.


tannerlinsley:
I would rather see a fix to the implementation to not require a class name. Good catch on that though. That's definitely important.

PR #1205, opened 2018-11-22
adding position prop into pagination component

AndrejGajdos:
I use styled-components and custom pagination component on top and bottom of the table. Pagination component has a bit different style on top of the table and I need to differentiate in styled component if pagination is on top or bottom. It would be nice to add a prop `isTop` into pagination component to support this use case.

### DIFF
--- a/src/index.js
+++ b/src/index.js
@@ -796,7 +796,7 @@ export default class ReactTable extends Methods(Lifecycle(Component)) {
       )
     }
 
-    const makePagination = () => {
+    const makePagination = (isTop) => {
       const paginationProps = _.splitProps(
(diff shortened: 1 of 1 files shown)


timkrins:
Hey man. This PR isn't really useful (at least, not any more), as the `showPaginationTop` prop is passed to the Pagination component.

timkrins:
What is probably a better solution and more in line with the customisability of the library is to allow passing in not just a `PaginationComponent`, but a `PaginationTopComponent` and a `PaginationBottomComponent`.

evgenyboxer:
@timkrins sounds like a great idea. should I send a PR for this?

Issue #1198, opened 2018-11-15
Ids for select table checkboxes


dclark27:
**Is your feature request related to a problem? Please describe.**
Currently, the Select Table HOC does not have the ability to add IDs to the select all checkbox, nor the individual checkboxes that are generated. Automation team members of mine need these IDs to create automatic test suites for the front end. 

**Describe the solution you'd like**
Add ID fields to checkbox props that I can pass to add IDs to the select all/individual checkboxes on the Select Table HOC.

**Describe alternatives you've considered**
Automatically add IDs based on the keyfield prop.

I am happy to create a PR for this issue.

tannerlinsley:
I would accept a PR to add/update/fix this functionality, or you can wait for v7 (a month or two). Thanks for the input!

igodorogea:
Previous PR #1208 breaks current solutions which are using `keyField="id"`, because `toggleSelection` in 6.8.6 was sending one thing(the desired id) but now in 6.9.0 is sending `select-#id`

ashkank83:
I'm experiencing the same issue as explained by @igodorogea . returned key in toggleSelection is select-#id now which I don't think is the correct behaviour. Any plans to fix this?

ggascoigne:
Why is this closed when it's clearly still broken?

tannerlinsley:
I am no longer actively triaging or working on issues for v6. That doesn't mean someone can't submit a fix for this issue (which I will accept as long as it is non-breaking), but v7 will eliminate this class of issue entirely, so it's not on my immediate radar.

ggascoigne:
I do understand this, I'd feel less frustrated if this hadn't been broken after you stopped actively working on v6.  From my point of view the latest v6 has been broken for 6 months.  Like several others, I'm still pinned to 6.8.6.

And like several others, I don't understand why the change was made in the first place, which makes submitting a PR to fix it rather awkward.

tannerlinsley:
Trust me, I understand your frustration. The decision was not purely logical though. I would have loved to continue to support v6 indefinitely, but the amount of issues that were coming in every day vs the contributors I had working with me (1 or 2) and the time we had for the project were an extremely unhealthy ratio. 

[...]

PR #1208, opened 2018-11-26
Adding IDs for select table, fixing backwards compatibility on floating left checkbox

dclark27:
Resolving issue #1198

### DIFF
--- a/src/hoc/selectTable/index.js
+++ b/src/hoc/selectTable/index.js
@@ -7,6 +7,7 @@ const defaultSelectInputComponent = props => {
     <input
       type={props.selectType || 'checkbox'}
       checked={props.checked}
+      id={props.id}
       onClick={e => {
         const { shiftKey } = e
         e.stopPropagation()
@@ -31,8 +32,8 @@ export default (Component, options) => {
         checked,
         onClick: toggleSelection,
         selectType,
-        id: row[keyField],
         row,
+        id: `select-id-${row[this.props.keyField]}`
       }
       return React.createElement(this.props.SelectInputComponent, inputProps)
     }
@@ -46,6 +47,7 @@ export default (Component, options) => {
         checked,
         onClick: toggleAll,
         selectType,
+        id: 'select-all'
       }
 
       return React.createElement(SelectAllInputComponent, inputProps)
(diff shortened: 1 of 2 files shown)


igodorogea:
@dclark27 this change https://github.com/tannerlinsley/react-table/pull/1208/files#diff-f874ee4d8a94bdabac6ab2be4706cebfR36 breaks current solutions which are using `keyField="id"`, because `toggleSelection` in 6.8.6 was sending one thing(the desired id) but now in 6.9.0 is sending `select-#id`

src/hoc/selectTable/index.js
```diff
@@ -31,8 +32,8 @@ export default (Component, options) => {
         checked,
         onClick: toggleSelection,
         selectType,
-        id: row[keyField],
         row,
+        id: `select-${row[keyField]}`
```
ryancanulla: Would this change make more sense on the defaultSelectInputComponent? If the intent is solely to create a unique ID for automation purposes, it seems wrong to break the pattern for managing the selection state in the SelectTable. 

If this is the approach that'll be used moving forward, I think we need an additional pull request to clarify what is being sent back, and update the example for SelectTable to maybe pull the selection ID from the row argument. 

I'm happy to create either, but would just ask for some direction from the powers at be as to which approach they would prefer; maybe I am missing something obvious and/or just flat out wrong here. :-)

Thanks in advance!

ryancanulla: For example:
`const defaultSelectInputComponent = props => {
  return (
    <input
      type={props.selectType || 'checkbox'}
      aria-label={`${props.checked ? 'Un-select':'Select'} row with id:${props.id}` }
      checked={props.checked}
      id={`select-${props.id}`}
      onClick={e => {
        const { shiftKey } = e
        e.stopPropagation()
        props.onClick(props.id, shiftKey, props.row)
      }}
      onChange={() => {}}
    />
  )
}`

mikemanyoung: Can someone comment on this? This broke for me too. I now need to reference row.id instead of id for the proper keyField id.

igodorogea: @mikemanyoung
The main thread is here #1198
My suggestion is to stay 6.8.6 fixed, and wait for next major release.

PR #1232, opened 2019-01-10
cleanup all but one linting issue

JoshMcCullough:
There were a few lint issues on `master`. I fixed the ones I could; one remains but I'm not sure what the fix is. the `cellInfo` variable is out of scope where it is used.


JoshMcCullough:
CI is failing due to one remaining lint issue.

PR #1249, opened 2019-02-05
revert 3baa7139fbb5cf5f706ec228036b0e2eac806113

jonniedarko:
## What
Revert Support for recursive Header Groups https://github.com/tannerlinsley/react-table/pull/1241

## Why
Causing bugs https://github.com/tannerlinsley/react-table/issues/1243
and https://github.com/tannerlinsley/react-table/issues/1248

### DIFF
--- a/src/methods.js
+++ b/src/methods.js
@@ -31,17 +31,27 @@ export default Base =>
       } = newState
 
       // Determine if there are Header Groups
-      const hasHeaderGroups = columns.some(column => column.columns)
+      let hasHeaderGroups = false
+      columns.forEach(column => {
+        if (column.columns) {
+          hasHeaderGroups = true
+        }
+      })
 
-      // Find the expander column which could be deep in tree of columns
-      const allColumns = _.iterTree(columns, 'columns')
-      const expanderColumn = _.getFirstDefined(allColumns)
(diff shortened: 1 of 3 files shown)


jonniedarko:
@tannerlinsley Since my changes are causing more issues then they are helping it might be best to revert. I will try to spend some time looking at a better solution for v6 or may even wait to v7

src/methods.js
```diff
@@ -31,17 +31,27 @@ export default Base =>
...
+      let columnsWithExpander = [...columns]
 
-      // Find the expander column which could be deep in tree of columns
-      const allColumns = _.iterTree(columns, 'columns')
-      const expanderColumn = _.getFirstDefined(allColumns)
+      let expanderColumn = columns.find(
+        col => col.expander || (col.columns && col.columns.some(col2 => col2.expander))
+      )
 
+      f (expanderColumn && !expanderColumn.expander) {
```
jbauerrfid: this looks like a typo, should be `if` IMO

jonniedarko: @jbauerrfid thanks, not sure how I missed this. I'm away for awhile, maybe you would be willing to submit PR to fix this?

PR #1279, opened 2019-02-23
Fixes for useExtended in useTable plus small bug

gary-menzel:
I wanted to get `useExtended` going but it was failing so fixed up the problems where the row specific hooks were not being applied and the path string for the key was not unique.

There was also a bug (which I saw in the useGroupBy code).

[PS - sorry for the incorrect username - it is fixed in my local .git/config now]


tannerlinsley:
Nice work Gary!

PR #1284, opened 2019-03-08
added useResizer

DDinenno:
I thought since you guys are porting over old features, I would add a resizer.  I tried to keep it as close to how you had it before. 

Here's a working demo
https://codesandbox.io/s/73qlyvj0y0


src/hooks/useResizer.js
```diff
@@ -0,0 +1,118 @@
...
+  const getResizer = (columns, api) => {
+    let bottomLevelColumnCount = 0;
+
+    columns.forEach(column => {
+      // if column is not a group (bottom level), add a resizer
+      if (!column.columns) {
+        const index = bottomLevelColumnCount;
+
+        column.resizer = () => (
+          <span
```
tannerlinsley: With the new headless approach, RT hooks should not expose or provide any UI. It would be better to provide a prop getter to the user like `column.getResizerProps(userPropOverrides)` that the user can then use in their own markup.

src/hooks/useResizer.js
```diff
@@ -0,0 +1,118 @@
...
+
+    if (updatedWidth !== initialWidth) {
+      setState(old => {
+        let { resizedColumns } = old;
+
+        resizedColumns[columnIndex] = updatedWidth;
+        columns[columnIndex].width = updatedWidth;
+
+        if (onResizedChange) {
+          onResizedChange(resizedColumns, e);
```
tannerlinsley: Triggering change functions is now an antipattern, since the user can now hoist the state out of react-table and use an effect to watch it, eg.

```
const state = useTableState()

const [{ resizedColumns }] = state

useEffect(() => {
  // resizedColumnsChanged
}, [resizedColumns])

useTable({
  ... 
  state
})

File: src/hooks/useResizer.js [hunk 3]
```diff
@@ -0,0 +1,118 @@
...
+
+              const onMouseUp = () => {
+                window.removeEventListener("mousemove", resizeColumnHandler);
+
+                // once completed, unsubscribe self
+                window.removeEventListener("mouseup", onMouseUp);
+              };
+
+              window.addEventListener("mousemove", resizeColumnHandler);
+              window.addEventListener("mouseup", onMouseUp);
```
tannerlinsley: All of this event listening works, but I get the feeling there is more performant way to do it. I'm going to take a deeper look at this.

tannerlinsley:
This is really great! I've added a few notes to help us move forward.

src/hooks/useResizer.js
DDinenno: Haha, I mean't to change that.  Since it worked I must have forgotten about it

src/hooks/useResizer.js
DDinenno: If you could find a better way, go for it.  I'll try to see what I could do to improve upon this also

DDinenno:
Thanks for the feedback.  I refactored the whole thing, I think it's alot cleaner this time around.  Lemme know if I can push it further.

JoseVf:
@DDinenno IMHO it'll be great to check for a `resizable` prop on the columns in cases that there are unresizable ones (maybe default to `true`?)

DDinenno:
@JoseVf  I think that would be pretty sweet. I added the prop in my last commit

src/hooks/useResizer.js
```diff
@@ -0,0 +1,124 @@
...
+    }
+
+    setState(old => {
+      let newResizedColumns = { ...old.resizedColumns };
+
+      newResizedColumns[index] = updatedWidth;
+
+      if (onResizedChange) {
+        onResizedChange(resizedColumns, e);
+      }
```
tannerlinsley: We don't use this callback style anymore. If users want to detect changes to internal state, they are supposed to hoist the state out (preferably using `useTableState`) and watch it in their own code using an effect or something.

src/hooks/useResizer.js
```diff
@@ -0,0 +1,124 @@
...
+    state: [{ resizedColumns }, setState],
+    getResizerProps,
+    onResizedChange
+  } = props;
+  const currentlyResizingInfo = useRef(undefined);
+
+  useEffect(() => {
+    if (resizedColumns) {
+      Object.keys(resizedColumns).forEach(index => (columns[index].width = resizedColumns[index]));
+    }
```
tannerlinsley: I'm confused by what this is doing. Is this updating the column with the new width? Seems like this should be done in a hook somewhere, no?

src/hooks/useResizer.js
```diff
@@ -0,0 +1,124 @@
...
+        column.isBottomLevel = true;
+
+        column.resizable = column.resizable !== undefined ? column.resizable : true;
+
+        column.getResizerProps = () => ({
+          ...(column.resizable && {
+            onMouseDown: e => onDragStart(e, column, index),
+
+            // prop overrides
+            ...(getResizerProps && getResizerProps)
```
tannerlinsley: I get what this is meant to do, but I think it would be better just to have the user pass any overrides to the `getResizerProps()` function instead of using a top-level option on the hook. Something like:

```
getResizerProps = ({onMouseDown, ...overrides } = {}) => ({
  onMouseDown: e => {
    onDragStart(e, column, index)
    if (onMouseDown) onMouseDown(e)
  },
  ...overrides
})

File: src/hooks/useResizer.js [hunk 7]
```diff
@@ -0,0 +1,124 @@
...
+  }, [resizedColumns]);
+
+  const addResizer = (columns, api) => {
+    let bottomLevelColumnCount = 0;
+
+    columns.forEach(column => {
+      if (!column.columns) {
+        const index = bottomLevelColumnCount;
+
+        column.isBottomLevel = true;
```
tannerlinsley: Is the bottom level detection necessary? If it is, then I wonder how we could make hooks like this one easier to write in the future. Detecting bottom-level headers shouldn't be this difficult...

src/hooks/useResizer.js
DDinenno: This is how i'm triggering a re-render when the columns widths change.  

If you were to put it somewhere else, where were you thinking and how would it be done?

src/hooks/useResizer.js
DDinenno: Fair enough.  That's actually better since there's less bulk features that not all users will tap into.

src/hooks/useResizer.js
DDinenno: It's  not  entirely necessary, just need to know what direction to go.  Could have it affect all columns including groups, which could be desired, or could affect only the bottom level.   You could argue that leaving it up to the user might be more in-line with your v7 philosphies.

Yea I was actually not super happy with this, but it worked for the time.

src/hooks/useResizer.js
DDinenno: Yea, i'm cool with that

tannerlinsley:
In light of changes that were made for the `useRowSelect` hook, I think we're going to have to make more similar core changes to support this one too. Going to close this for now, but may ask for feedback on the feature here as I build it. Thanks guys!

DDinenno:
Yea that's cool, I was having some difficulty implementing it in a clean way.  

If you're interested I did another refactor last week , which I had not posted.  It is alot more performant than this current implementation as well as V6.  You can increase the row-count to around 500 and it's still somewhat responsive (Mind you this is an un-optimized build and inside codesandbox).  It's fast because I'm doing manual DOM manipulation while the dragging the mouse and invoking setState once `onDragEnd`.    

If you could cannibalize it somehow, go ahead.

https://codesandbox.io/s/sly2n

PR #1285, opened 2019-03-09
Added useSimpleLayout hook to facilitate TR/TD markup

gary-menzel:
Along with the previous fixes for `useExtended`, the `useSimpleLayout` hook provides a "bare-bones" layout mechanism that doesn't make any assumptions about the CSS or styling used in the table markup.

This means a very basic table (without any pagination and traditional CSS classes or style props can be used outside of ReactTable).

The imported hooks look like this:
```
import {
  useTable,
  useColumns,
  useRows,
  useSimpleLayout
} from "react-table"
```
The instance setup looks like this:
```
  const instance = useTable(
    {...props, manualPagination: true},
    useColumns,
    useRows,
    useSimpleLayout,
  )
  
  const {
    rows,
    headerGroups,
    getRowProps,
    prepareRow,
  } = instance
```

Once the PR is merged, I'll put together a codesandbox (I have one running locally) that renders normal `<table>`, `<tr>` and `<td>`.  It is only about 95 lines long.

I should have created a new branch for this  but it is the original branch - so all the other commits that have already been merged are in here as well.  The PR request is saying it can't be merge automatically.  If that is the case, respond to the PR and I will branch from MASTER and reapply the changes (it's only one file added and the `index.js` updated to export the `useSimpleLayout` hook).


gary-menzel:
OK - I can see what the conflicts are now.  They are additional fixes for `useExtended` that were still breaking and causing a failure if `useExtended` was not included in the instance creation (impacting the `useSimpleLayout` when trying to cut down on the number of hooks required).

Once again, if it is too tricky to merge them, then I can just branch again and migrate those over.

src/hooks/useSimpleLayout.js
```diff
@@ -0,0 +1,65 @@
...
+      getHeaderRowProps,
+      getHeaderProps,
+      getCellProps
+    }
+  } = props
+
+  rows.forEach((row, i) => {
+    if (row.cells === undefined) {
+      row.cells = columns.map(column => {
+        const cell = {
```
tannerlinsley: Shouldn't cell generation happen regardless of what renderer or hook we use? I thought it was part of the `prepare` script.

src/hooks/useSimpleLayout.js
```diff
@@ -0,0 +1,65 @@
...
+          state: null,
+          value: row.values[column.id]
+        }
+        return cell
+      })
+    }
+  })
+
+  columnsHooks.push((columns, api) => {
+    columns.filter(column => {
```
tannerlinsley: This could just be `columns.forEach`

src/hooks/useSimpleLayout.js
```diff
@@ -0,0 +1,65 @@
...
+    })
+
+    const rowStyles = {
+      style: {
+      }
+    }
+    api.rowStyles = rowStyles
+
+    getRowProps.push(() => rowStyles)
+    getHeaderRowProps.push(() => rowStyles)
```
tannerlinsley: These are unnecessary I think?

src/hooks/useSimpleLayout.js
tannerlinsley: If cell generation is happening in `useFlexLayout`, then I don't blame you for putting it here. It might be better to put that in the `prepareRow` function and let hooks plug in to that. Thoughts?

src/hooks/useTable.js
```diff
@@ -139,14 +139,16 @@ export const useTable = (props, ...plugins) => {
   // This function is absolutely necessary and MUST be called on
   // any rows the user wishes to be displayed.
   api.prepareRow = row => {
-    const { index } = row
+    const { path, index } = row
     row.getRowProps = props =>
       mergeProps(
-        { key: ['row', index].join('_') },
+        { key: ['row', path || index].join('_') },
```
tannerlinsley: Should we just use `path` all the time? Is `path` set when not using `useExpanded`?

src/hooks/useTable.js
```diff
@@ -139,14 +139,16 @@ export const useTable = (props, ...plugins) => {
...
         applyHooks(api.hooks.getRowProps, row, api),
         props
       )
 
+    // need to apply any row specific hooks (useExpanded requires this)
+    applyHooks(api.hooks.row, row, api)
```
tannerlinsley: I could have sworn we were already executing a `row` specific hook, were we not?

tannerlinsley:
Now that I see this simple table layout coming together, I think it might be cool to also do the `useAbsoluteLayout` at the same time. It would give us even more insight into the repeated code and patterns for layout hook plugins. Thoughts?

gary-menzel:
I've had a read through all your comments and think you are right in all cases.

Specifically, cell generation was happening in `useExpanded` but call filtering is happening in `useTable`.

Now that we've had this conversation, I'll go back and look at those points and see if we can refactor and streamline it according to your suggestions and let you know if there are any issues.

What did you imagine the `useAbsoluteLayout` would do?  Happy to work on that but would need to understand what your thinking is on it.

I suspect that `layout` hooks always come last in the chain and this is where the line starts to blur between what RT does and what the developer does.  But it doesn't mean RT can't supply a bunch of them out-of-the-box.

NOTE: I also have the first real chapter of the eBook written now (based on using the `useSimpleLayout` hook.

tannerlinsley:
The absolutely positioned layout hook would use absolute styles for columns and cells and rows.  That would then allow for animations when the rows and columns are added, removed, or rearranged.  Much like you see on AG grid

gary-menzel:
I'll just focus on doing the refactor and making sure the layering of the hooks has a logical progression as well as them being relevant to their purpose.

I'll probably just branch off MASTER again and reapply my changes (it's only one file and a few lines).  This will give me a clean slate to work with.

Let's leave this PR open for now so we know what I am working on - I'll close it off when I complete this next round of work and submit a new PR.

gary-menzel:
@tannerlinsley I have not forgotten about this - just other things have gotten in the way of working on ReactTable (and the book).  I do need to get this part done though so I can move on to the next chapter.  Just thought I would let you know.

tannerlinsley:
Would love to get this merged soon. Any idea on ETA?

gary-menzel:
Hi Tanner,

I’m in Japan at the moment but will have a look and let you know.

gary-menzel:
OK - I have my head back in this now.

I'll move the cell preparation into `prepareRow` as per your previous comments and also tidy up any "messy" code (like more sensible iteration - where possible).  I'll review the `applyHooks` in regard to single rows but I think it definitely needs to be inside the `prepareRow` function.

If I could just ask quickly for completion sake... Did you originally envision there being no dependency chain in the hooks or did you expect a specific order?  For example, in the sample TD/TR renderer I have based on `useSimpleLayout` the order is:
* props
* useColumns
* useRows
* useSimpleLayout

As far as I can tell, the columns state needs to be set up before any rows can be set up - so that appears to be a clear non-negotiable dependency.  And any `layout` hook would require both column and row state to have been initially established.  My question (for general knowledge - and probably the book) is how hook dependencies can be managed beyond that - with sorting, filters, expanders, toggles, etc.

tannerlinsley:
- Make sure you are up to date with master and take a look at my recent changes. I have moved even more base logic into the prepareRow function and also added some defaults to help the user navigate situations where they attempt to render rows that are not prepared. This may align with your goals, either way.
- You are absolutely right, there is most definitely a dependency order requirement. We **could** remove that requirement by turning everything into a fully-hook based system, which would allow you to register plugin hooks *mostly* in any order you want. Ordering would still matter for things that hook into the same table lifecycles though, since earlier plugins would be added first, and also run first in the chain. However, for things like setting up columns and rows before anything else runs, that would be easy to control since people won't (and shouldn't) be loading up multiple column or row plugins.  One could also argue that the column and row plugins should be built in, but IMO, by removing them and dogfooding our own system, we are ensuring their robustness, testability, and also maintainability.

gary-menzel:
Yes - up-to-date with `master` - I could see the changes in there and have adjusted the `useSimpleLayout` accordingly (there's not much code left in their now).

I understand about the layering and leaving the `useColumns` and `useRows` as they are (it allows for a lot of flexibility).

As far as the `useSimpleLayout` is concerned I think it is "done".  What I plan to do is copy my change out and branch off `master` again to ensure it is completely dependency free in the most recent version.  In which case, I will close this PR off and open another.


I do have another question though....

I was looking to make sure I could still use pagination in my very cut down TR/TD renderer and wondered how to set a default page size.  I have a suspicion I have to supply my own `state` but wondered if there was a more obvious/intuitive way to supply some of those types of defaults.

I plan to expand slowly in my book on the basic example I started with and show how to first add Pagination and then eventually build up to a more complex table.

I know we spoke about the `useAbsoluteLayout` but I'd like to try and ensure that all the dependency pieces are decoupled as much as possible before embarking on a different type of layout.

As far as "renderers" are concerned - how do you plan to include those in the core?  Clearly, there need to been some default ones - maybe a traditional `<table>` style render (as I have been working on) as well as the `<div>` based one that can work with both `flex` or `absolute` layouts.

Although - I realise the `power` of RT7 is that you can just build your own renderer.

gary-menzel:
Closing in favour of #1322

PR #1300, opened 2019-03-26
Pass defaults via props into useTable state

berdyshev:
This will allow to override any defaults for the table state (like `pageSize` for example)


tannerlinsley:
This doesn't quite cover the use case of hoisting the state out yourself for control. Take a look at the server-side rendering example in the code sandbox. 

Using `useTableState()` in your own code and passing to to the `useTable` hook is a necessary point of inversion of control to the user.

tannerlinsley:
I'm not sure why you wouldn't just do this:

```javascript
const state = useTableState(defaults)

....
const table = useTable({ state })
```

berdyshev:
I renamed it to `defaults` to make more self-explanatory, but I missed SSR point. So possible it makes sense to return back to the `state` name. But...
In response to the second question: I cannot use `useTable({ state })` because this requires to pass the complete state, which is not a good option when you need to just change only particular default value (e.g. `pageSize` in my first comment)

tannerlinsley:
This should still be sufficient and extremely easy:

```javascript
const state = useTableState({ pageSize: 10 })
const table = useTable({ state })
```

berdyshev:
ok, agree. should be sufficient.

PR #1314, opened 2019-04-10
Remove usage JSON.stringify where possible

kennetpostigo:
This adds 2 new utils:
- compare
- clone

I saw a previous issue where you did not want to include lodash, so I added these 2 utils from the[ `just` library](https://github.com/angus-c/just)

I also ran a benchmark on them and they run significantly faster. Hoping this can be released as a v6 patch. Are there any unit tests to ensure behavior is the same @tannerlinsley ?

### DIFF
--- a/docs/src/stories/HOCReadme.js
+++ b/docs/src/stories/HOCReadme.js
@@ -2,7 +2,7 @@
 import React from "react";
 import marked from "marked";
 //
-import HOCReadme from "!raw!../../../src/hoc/README.md";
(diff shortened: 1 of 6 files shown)


tannerlinsley:
While I think this would be a good addition to the v6 API, I think it's better if we take these thoughts and move on to v7. A lot of this won't be needed in v7 regardless. If you absolutely need these changes in v6, I suggest that you maintain a fork until v7 is ready and you can migrate over.

kennetpostigo:
@tannerlinsley when will v7 be released?

PR #1322, opened 2019-04-28
Feature/useSimpleLayout

gary-menzel:
This should have all the outstanding changes that we discussed in the other PR.

I'll close the other one off now in favour of this one (which is on top of the current `master`).


src/hooks/useTable.js
```diff
@@ -143,7 +143,7 @@ export const useTable = (props, ...plugins) => {
     const { path } = row
     row.getRowProps = props =>
       mergeProps(
-        { key: ['row', path].join('_') },
+        { key: ['row', path.join('_')].join('_') },
```
tannerlinsley: Just spread the path here. ...path

gary-menzel: Changed now.

JoseVf:
@gary-menzel do you have any code sample/snippet on how to use it?

PR #1367, opened 2019-06-18
Fixed getRowProps hook prop aggregation

DDinenno:
Simple one liner fix.  I think you mistakenly used `applyProps` instead of `applyPropHooks` and it was causing the the next hook to overwrite the previous.

I created an issue for this here: [https://github.com/tannerlinsley/react-table/issues/1366](url)


tannerlinsley:
Great PR. Thanks!

PR #1369, opened 2019-06-18
Fix pagination resetting to page zero with manualPagination

ggascoigne:
To be honest I'm not sure what this useLayoutEffect is there to do.
It has no visible effect if you don't use manualPagination, and if you
do it, simply jumps you back to the first page, defeating the point of
you having control of the pagination.

### DIFF
--- a/src/hooks/usePagination.js
+++ b/src/hooks/usePagination.js
@@ -1,4 +1,4 @@
-import { useMemo } from 'react'
+import { useMemo, useLayoutEffect } from 'react'
 import PropTypes from 'prop-types'
 
 //
@@ -23,6 +23,7 @@ export const usePagination = props => {
   const {
     rows,
     manualPagination,
+    disablePageResetOnDataChange,
     debug,
     state: [
       {
@@ -37,6 +38,17 @@ export const usePagination = props => {
     ]
   } = props
 
+  const rowDep = disablePageResetOnDataChange ? null : rows
+  useLayoutEffect(() => {
+    setState(
+      old => ({
+        ...old,
+        pageIndex: 0
+      }),
+      actions.pageChange
+    )
+  }, [setState, rowDep, filters, groupBy, sortBy])
+  
   const { pages, pageCount } = useMemo(() => {
     if (manualPagination) {
       return {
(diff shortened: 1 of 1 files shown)


src/hooks/usePagination.js
```diff
@@ -37,16 +37,6 @@ export const usePagination = props => {
...
 
-  useLayoutEffect(() => {
-    setState(
-      old => ({
-        ...old,
-        pageIndex: 0
-      }),
-      actions.pageChange
-    )
-  }, [rows, filters, groupBy, sortBy])
```
tannerlinsley: When rows, filters, groupBy or sortBy are changed, it essentially invalidates the data on pages that you've previously viewed (say you were on page 2, page 1's data has likely changed now). This effect is there to take you back to page 1, so you can see the new data that was generated from your data + filters + grouping + sorting.

I fully intend to keep this as the default (since this is what I have found to be the default in almost every other table), but I would also immediately accept a PR to add an option to disable this effect if you choose to do so.

ggascoigne: And I'll admit that that is a feature that I want and makes sense.  My problem is that with remote data it simply resets to page zero on every refresh.  You can see that on the codesandbox if you switch from the local react-table files and use the alpha-5 version, the serverData version can no longer page.

ggascoigne: actually, should rows be in that dependency list?  In a scenario where the remote data is being updated periodically, but everything else is remaining constant, would you want it to jump back to the first page?

tannerlinsley: I think you just proved the need for this option. Can't wait to see the PR ;)

tannerlinsley: And yes, I think by default, rows should be in there. 99% of the time if the core data changes, I think the page should reset. There are clearly times when that shouldn't happen though. A boolean should be enough for now.

src/hooks/usePagination.js
```diff
@@ -37,6 +38,7 @@ export const usePagination = props => {
     ]
   } = props
 
+  const rowDep = disablePageResetOnDataChange ? null : rows
```
ggascoigne: note that this construct is to ensure that lint doesn't give this error:

```
warning  React Hook useLayoutEffect has a second argument which is not an array literal. This means we can't statically verify whether you've passed the correct dependencies
```
when I tried to inline the dep change.

PR #1371, opened 2019-06-19
Fix lint errors from react-hooks/exhaustive-deps

ggascoigne:
They all seemed like reasonable warnings, this squashes them.

### DIFF
--- a/.eslintrc
+++ b/.eslintrc
@@ -1,14 +1,19 @@
 {
-  "parser": "babel-eslint",
-  "extends": ["react-app", "prettier"],
-  "env": {
-    "es6": true
-  },
-  "parserOptions": {
-    "sourceType": "module"
-  },
-  "rules": {
-    "space-before-function-paren": 0,
-    "react/jsx-boolean-value": 0
-  }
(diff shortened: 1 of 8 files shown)


Grsmto:
I confirm this.
Also good to note that using VSCode and Prettier, these get added automatically as well, so no need to manually enter the deps for the hooks or even care about Eslint.

tannerlinsley:
I'll have to work through some of these. I initially built this version of RT when there was no literature or rules for effects, so it was all manually architected for change detection and sometimes purposely leaving out specific dependencies from some of the effects. Applying the exhaustive deps rule to this will most likely alter the way the effects run. Shouldn't take long though. I want that rule and I want it to be applied to 100% of the code (it's fantastic). Just need to upgrade it.

ggascoigne:
I can say that using this version is working very nicely for me, and I didn't notice any problems using the version with these changes, but yes, some of them are somewhat impactful to the code.

tannerlinsley:
I don't think many of them will impact the outcome too much. Just want to make sure my expectations for when the effects were going to run were still met.

PR #1372, opened 2019-06-19
Fix useExpanded wrongly flattening subRows

Grsmto:
Hey guys,

Here is a PR that fixes `useExpanded` that was not usable at the moment. The reason is that all the `subRows` were flattened at the root level. Which was breaking the `usePagination` component.
Now the rows are correctly nested and pagination works fine.

What is did is the following:

**Before**
```
[
  { id: 'row-0' }
  { id: 'row-0-0' }
  { id: 'row-1' }
  { id: 'row-2' }
  { id: 'row-3' }
]
```

**After**
```
[
  { 
    id: 'row-0', 
    subRows: [
      { id: 'row-0-0' } 
    ] 
  }
  { id: 'row-1' }
  { id: 'row-2' }
  { id: 'row-3' }
]
```

Please see the following code sandbox with a working example: https://codesandbox.io/s/react-table-v7-sandbox-8wr0e

Screenshot 2019-06-20 at 12 42 17

### DIFF
--- a/src/hooks/useExpanded.js
+++ b/src/hooks/useExpanded.js
@@ -21,7 +21,6 @@ export const useExpanded = props => {
 
   const {
     debug,
-    columns,
     rows,
     expandedKey = 'expanded',
     hooks,
@@ -48,8 +47,6 @@ export const useExpanded = props => {
   const expandedRows = useMemo(() => {
     if (debug) console.info('getExpandedRows')
 
-    const expandedRows = []
-
     // Here we do some mutation, but it's the last stage in the
     // immutable process so this is safe
     const handleRow = (row, index, depth = 0, parentPath = []) => {
@@ -62,17 +59,15 @@ export const useExpanded = props => {
       row.isExpanded =
         (row.original && row.original[expandedKey]) || getBy(expanded, path)
 
-      expandedRows.push(row)
-
       if (row.isExpanded && row.subRows && row.subRows.length) {
         row.subRows.forEach((row, i) => handleRow(row, i, depth + 1, path))
       }
+
+      return row
     }
 
-    rows.forEach((row, i) => handleRow(row, i))
-
-    return expandedRows
-  }, [rows, expanded, columns])
+    return rows.map((row, i) => handleRow(row, i))
+  }, [debug, rows, expandedKey, expanded])
 
   const expandedDepth = findExpandedDepth(expanded)
 
--- a/src/hooks/useExpanded.js
+++ b/src/hooks/useExpanded.js
@@ -13,7 +13,8 @@ addActions({
 })
 
 const propTypes = {
-  expandedKey: PropTypes.string
+  expandedKey: PropTypes.string,
+  paginateSubRows: PropTypes.bool
 }
 
 export const useExpanded = props => {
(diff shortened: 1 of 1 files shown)


tannerlinsley:
This is the way the old React Table worked, by nesting the rows. We've found though that problems start to happen with that approach when you require the user of the library to recurse the rows and choose whether to nest/unnest them in the actual markup. Do you use a row-group element? Also, this makes pagination inconsistent if you have a page of 10 rows, and one of those rows has like 100 sub-rows. I think we'll want to keep the expanded rows flattened for now and just make sure that the keys are correct and that they are passed the right props for things like nesting level and nesting path (I think they already have those)

With that explanation, what was broken with the previous approach?

Grsmto:
It was broken because the `page` property used to do the pagination work was wrong: if a row had a `subRow`, then the `page` was taking that `subRow` into account.
So for ex, for a page of 10 items, if you had an item with a `subRow` it would actually display 9 items + 1 `subRow`.

tannerlinsley:
I see. Sounds like we should instead make sure that the pagination is not taking subrows into account (since they are being flattened). Thoughts?

Grsmto:
Okay, so I understand you want to keep the flattened organisation. I'll see to rework this fix to keep that untouched then!
Thanks for the clarification.

tannerlinsley:
Yep :) The flattened organization makes it easier to consume when building your own renderer (which is now required in v7) and it also helps keep things simple (no recursion, row groups, etc).

Grsmto:
Could you clarify this:
>Also, this makes pagination inconsistent if you have a page of 10 rows, and one of those rows has like 100 sub-rows. 

I'm not sure to understand what you mean. What is the expected behaviour? For me if you have 10 rows and one has 100 sub-rows, the page should still display the 10 rows + the 100 sub-rows.
Is that correct or I'm mistaken?

tannerlinsley:
Under the nested subrow approach (instead of flattening), If you have your pagination set to 10 rows per page, then expand the first row's subrows, then you would technically be seeing:
```
1 normal row
- n subrows
9 normal rows
```

That could easily result in more than 10 items. That's all I was saying.

tannerlinsley:
To be clear, if the pageSize is set to 10, then there should only be 10 total rows (including sub rows) showing. That's now how it behaved in V6, but that's how we're going to default in V7. If we need or find a use case to have sub-rows not affect pagination, then we can add that option in later

Grsmto:
Oh really ok. So I was considering this a bug.
Seems really weird to me UX wise that expanding a row would actually affect the pagination.
I suppose you can close this PR then if you consider this the expected behaviour.

tannerlinsley:
Okay. On the other hand, I'm not opposed to having a flag that you could set that would disable paginating sub rows. Thoughts?

Grsmto:
I changed this with your suggestion, there is now a `paginateSubRows` option that's defaulting to `true`.
Passing `false` will not consider the `subRows` for pagination.

This allow us to do such thing like having a custom `subRow` that is not an actual table row but any custom content. In this case it doesn't make sense for it to have an impact on the pagination:

image

tannerlinsley:
Fantastic!

PR #1374, opened 2019-06-21
Add support for es module build

Grsmto:
Hi Tanner,

Here is a PR that provides an ESM build to the tools that support it.
For example Webpack now uses the `.es` version of the library instead of the umd one, making things lighter.

Hope this helps,


tannerlinsley:
Awesome work

PR #1377, opened 2019-06-26
[WIP]Feat/use select row

Grsmto:
Hey,

Here is a new `useSelectRow` hook.  
- Adds a `isSelected` prop to the `row` object.
- Adds a `getSelectRowToggleProps` to the global API. This function return an object with a `onChange` event and `checked` boolean. This is meant to be passed to a `checkbox` component (I tested it with both native input like in the CodeSandbox or with a fully custom component both can work).

Here is a CodeSandbox example https://codesandbox.io/s/react-table-v7-sandbox-u87y8


Grsmto:
@lishine regarding your comment [there](https://github.com/tannerlinsley/react-table/issues/1252#issuecomment-506353591) we should better keep talking here instead. 
Please tell me if I understand you well:
You can't have access to the selected rows from outside the `Table` components, is that correct?

Grsmto:
>But the click on the checkbox gets registered as well, need to be excluded, this is what I will add to my code now.

Regarding your click issue, that's up to you to `stopPropagation` in your `click` callback or do any other event handling. The hook doesn't do anything about this.

lishine:
> @lishine regarding your comment [there](https://github.com/tannerlinsley/react-table/issues/1252#issuecomment-506353591) we should better keep talking here instead.
> Please tell me if I understand you well:
> You can't have access to the selected rows from outside the `Table` components, is that correct?

Yes, and I solved it like this, I wrapped the useTable hook and use it my component, so I have access to rows.
```
import React, { useState, useEffect } from 'react'
import {
    useTable,
    useColumns,
    useRows,
    useFilters,
    useSortBy,
    useExpanded,
    usePagination,
    useFlexLayout,
    useSelectRow,
    useTableState,
} from '../react-table/index'

export const useRTable = ({ tableStateInit, tableProps }) => {
    const [tableState, setTableState] = useTableState(tableStateInit)
    const [totalCount, _setTotalCount] = useState(0)

    const tableInstance = useTable(
        { state: [tableState, setTableState], ...tableProps },
        useColumns,
        useRows,
        useFilters,
        useSortBy,
        useExpanded,
        usePagination,
        useFlexLayout,
        useSelectRow
    )
    const { rows } = tableInstance

    const setTotalCount = totalCount => {
        _setTotalCount(totalCount)
        const { pageSize } = tableState
        setTableState(old => ({
            ...old,
            pageCount: Math.ceil(totalCount / pageSize),
        }))
    }

    return {
        tableInstance,
        totalCount,
        setTotalCount,
        tableState,
        rows,
    }
}
```
```
export const Signs = () => {
    const loading = useLocalStore('loading')
    const load = useLocalActions('load')
    const data = useLocalStore('data')

    const tableInstanceProps = useRTable({
        tableStateInit: { pageCount: 0, pageSize: 3 },
        tableProps: {
            data,
            columns,
            manualSorting: true,
            manualFilters: true,
            manualPagination: true,
            disableMultiSort: false,
            disableGrouping: true,
            disableSorting: true,
            disableFilters: true,
        },
    })
    const { setTotalCount, tableState, rows } = tableInstanceProps
    const { sortBy, filters, pageIndex, pageSize } = tableState

    useEffect(() => {
        load({ tableState, setTotalCount, columns })
    }, [sortBy, filters, pageIndex, pageSize])

    return (
        <RTable
            {...{
                loading,
                onRowClick,
                ...tableInstanceProps,
            }}
        />
    )
}
```

lishine:
I added to my useRTable, calculation of { rowsSelected, totalRowsSelected }, see bellow.
And for it to work your useSelectRow hook should return new reference for rows, something like this:
```
    const mutatedRows = useMemo(() => {
        rows.forEach(row => {
            row.isSelected = selected.includes(row.index)
            row.toggleSelected = () => toggleSelected(row.index)
        })

        return [...rows]
    }, [rows, selected, toggleSelected])
```
(There is some option for useSelectRow to use state [selected, setSelected]  received externally, that way selected can be seen outside)
```
import React, { useState, useEffect, useMemo } from 'react'
import {
    useTable,
    useColumns,
    useRows,
    useFilters,
    useSortBy,
    useExpanded,
    usePagination,
    useFlexLayout,
    useSelectRow,
    useTableState,
} from '../react-table/index'

export const useRTable = ({ tableStateInit, tableProps }) => {
    const [tableState, setTableState] = useTableState(tableStateInit)
    const [totalCount, _setTotalCount] = useState(0)

    const tableInstance = useTable(
        { state: [tableState, setTableState], ...tableProps },
        useColumns,
        useRows,
        useFilters,
        useSortBy,
        useExpanded,
        usePagination,
        useFlexLayout,
        useSelectRow
    )
    const { rows } = tableInstance

    const { rowsSelected, totalRowsSelected } = useMemo(() => {
        let totalRowsSelected = 0
        let rowsSelected = []
        rows.forEach(({ isSelected, index }) => {
            rowsSelected[index] = isSelected
            if (isSelected) {
                totalRowsSelected += 1
            }
        })
        return { rowsSelected, totalRowsSelected }
    }, [rows])

    const setTotalCount = totalCount => {
        _setTotalCount(totalCount)
        const { pageSize } = tableState
        setTableState(old => ({
            ...old,
            pageCount: Math.ceil(totalCount / pageSize),
        }))
    }

    return {
        tableInstance,
        totalCount,
        setTotalCount,
        tableState,
        totalRowsSelected,
        rowsSelected,
    }
}
```

lishine:
> > But the click on the checkbox gets registered as well, need to be excluded, this is what I will add to my code now.
> 
> Regarding your click issue, that's up to you to `stopPropagation` in your `click` callback or do any other event handling. The hook doesn't do anything about this.

Thanks for the idea!
So the onRowClick with stopPropagation looks like this.
Clicking on row is also usual use case so code section like this can make itself into the official sandbox example
```
    const renderRow = (row, index, style = {}) => {
        if (!row) {
            return (
                <Row {...{ style, even: index % 2 }}>
                    <Cell>Loading more...</Cell>
                </Row>
            )
        }
        prepareRow(row)
        return (
            <Row
                onClick={() => onRowClick({ index })}
                {...row.getRowProps({ style, even: index % 2, selected: row.isSelected })}
            >
                {row.cells.map(cell => {
                    return (
                        <Cell
                            {...cell.getCellProps()}
                            onClick={e => {
                                if (cell.column.id === 'selectRow') {
                                    e.stopPropagation()
                                }
                            }}
                        >
                            <span>{cell.render('Cell')}</span>
                        </Cell>
                    )
                })}
            </Row>
        )
    }
```

src/hooks/useSelectRow.js
```diff
@@ -0,0 +1,56 @@
...
+  )
+
+  const mutatedRows = useMemo(() => {
+    rows.forEach(row => {
+      row.isSelected = selected.includes(row.index)
+      row.toggleSelected = () => toggleSelected(row.index)
+    })
+
+    return rows
+  }, [rows, selected, toggleSelected])
```
tannerlinsley: I know this is necessary, but something here is smelly to me. Could be:
- That forEach loop could get expensive on large datasets. I wonder if we could hook into an existing loop over the data?
- Using useMemo, but *mutating* the rows, instead of replacing them with new objects. Usually I favor immutability, but maybe this is the right way?
- Every time the selected state changes, it will loop over all of the rows. Again, this could get expensive and slow for large datasets.

src/hooks/useSelectRow.js
```diff
@@ -0,0 +1,56 @@
...
+
+  const getSelectRowToggleProps = () => ({
+    onChange: toggleSelectAll,
+    checked: selected.length === rows.length
+  })
+
+  return {
+    ...props,
+    rows: mutatedRows,
+    getSelectRowToggleProps,
```
tannerlinsley: I think this could have a better name like `getSelectAllToggleProps`. I wasn't sure at first if it was meant for single rows or all of them until I inspected the code closer. Thoughts?

src/hooks/useSelectRow.js
```diff
@@ -0,0 +1,56 @@
+import { useMemo, useState, useEffect, useCallback } from 'react'
+
+export const useSelectRow = props => {
+  const { rows, data } = props
+  const [selected, setSelected] = useState([])
```
tannerlinsley: The state living here technically works, but similar to other hooks, I think it should use the `useTableState` initial state and be hoistable like all of the other state.

tannerlinsley: Take a look at the usePagination hook to see how that's done.

src/hooks/useSelectRow.js
Grsmto: >That forEach loop could get expensive on large datasets. I wonder if we could hook into an existing loop over the data?
>Every time the selected state changes, it will loop over all of the rows. Again, this could get expensive and slow for large datasets.
Is that something the `api.hooks.rows` could solve?
If I understand the `api.hooks` thing well, you designed that exactly for that purpose, or I'm mistaken?

>Using useMemo, but mutating the rows, instead of replacing them with new objects. Usually I favor immutability, but maybe this is the right way?
Yeah I agree absolutely, tbh I think I just followed an example from another hook (I think it was `usePagination` but anyway).

fantactuka:
@tannerlinsley, @Grsmto Question related to this one:

I've noticed that `usePagination` + `useExpanded(paginateSubRows: false)` changes the structure of rows and does not flatten them. With these settings it seems like `useSelected` won't work properly for nested rows. (I suppose that checking parent row should check all nested rows and vice versa.)

tannerlinsley:
Decided to go with a more internal implementation of this feature. Turns out I needed to add a few more core features to make it work, but I'm happy with it. Thanks for your input guys!

fantactuka:
@tannerlinsley v7 already looks great, thank you for making it happen!

PR #1381, opened 2019-06-29
Fixed grouped column Header that were not adjusting based on hidden child columns

DDinenno:
The grouped header columns are not adjusting width if child columns update their `show` property.  This is because the Headers are basing their width on the sum of the child columns, but not factoring in the `show` property.

You can see an example below.  I added a button to toggle the fix on and off.

https://codesandbox.io/s/react-table-v7-sandbox-2qyr4

### DIFF
--- a/src/hooks/useColumns.js
+++ b/src/hooks/useColumns.js
@@ -114,7 +114,7 @@ export const useColumns = props => {
     const recurse = columns => {
       columns.forEach(d => {
         if (!d[childKey]) {
-          flatColumns.push(d)
+          if (d.show === undefined || d.show) flatColumns.push(d)
         } else {
           recurse(d[childKey])
         }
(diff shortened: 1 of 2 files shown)


tannerlinsley:
I don't see a difference when clicking the fix on or off. Maybe make sure your sandbox is up to date?

Also, would it be better to just update the measuring to not use the hidden columns, instead of affecting anything using the `flattenColumns` function?

DDinenno:
Yea my sandbox has been edited since, I thought I forked it since making some changes. 

You're absolutely right about leaving the `flattenColumns` util alone.  I was able to just add a filter before calculating the sum inside `useFlexLayout` 

Heres an updated sandbox: https://codesandbox.io/s/react-table-v7-sandbox-zw2xi

PR #1401, opened 2019-07-23
Memoize table data, fix useRows text

codepunkt:


tannerlinsley:
This is no longer needed, given the recent changes to `useColumns` and `useRows`

PR #1412, opened 2019-07-29
Small readme example update

jaapjanfrans:
in commit ea79cd838891a21db6754c387a73d3ea4fffd127 the`getRowProps` member of the headerGroup object changed into `getHeaderGroupProps`

The example in readme still referenced the old `getRowProps`, I think this should be updated to `getHeaderGroupProps`.


tannerlinsley:
Good catch!

PR #1420, opened 2019-07-30
Avoid react error if the data source doesn't have the property.

CyrilSiman:
I have a data source where for some rows I don't have a property define in columns.

### DIFF
--- a/src/utils.js
+++ b/src/utils.js
@@ -39,7 +39,7 @@ export function decorateColumn(column, defaultColumn, parent, depth, index) {
 
   column = {
     Header: ({ id }) => id,
-    Cell: ({ value }) => value,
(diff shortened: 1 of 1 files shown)


src/utils.js
```diff
@@ -39,7 +39,7 @@ export function decorateColumn(column, defaultColumn, parent, depth, index) {
 
   column = {
     Header: ({ id }) => id,
-    Cell: ({ value }) => value,
+    Cell: ({ value }) => value ? value : '',
```
tannerlinsley: How about `typeof value !== 'undefined' ? value : ''`?

CyrilSiman: Off course, you right ! With few reflection my pseudo fix isn't a really one :-(

PR #1429, opened 2019-08-04
[Minor] use actual param in `ensurePluginOrder`

gargroh:


src/utils.js
```diff
@@ -330,7 +330,7 @@ export function ensurePluginOrder(plugins, befores, plugin, afters) {
     const afterIndex = plugins.findIndex(plugin => plugin.pluginName === after)
     if (afterIndex < pluginIndex) {
```
gargroh: It seems check for `-1` is also needed here, whats ur take on this @tannerlinsley ?

tannerlinsley: Yeah. It is. I'll merge this PR, then make a few more changes, including getting tests to pass now that this is working better :)

PR #1442, opened 2019-08-13
Allow row generator to listen for column changes

DDinenno:
Proposing a change to re-generate rows when columns update.  Reason being, I hit a wall when creating custom hooks which rely on columns.  I was able to get the Headers to adjust to the new config, but the rows ignore all changes.

I'm not sure if you omitted the `columns` variable from the row generator memo listeners for some design reason or other.  The only bug I found was inside the `usePagination` hook, it caused an infinite loop of setStates when adding the variable, but it was fixed rather easily.


src/plugin-hooks/usePagination.js
```diff
@@ -52,13 +52,14 @@ function useMain(instance) {
 
   useLayoutEffect(
     () => {
```
tannerlinsley: If this wasn't working before then I think the bug is deeper in some memoization.

src/hooks/useTable.js
```diff
@@ -206,7 +206,7 @@ export const useTable = (props, ...plugins) => {
         console.timeEnd('getAccessedRows')
       return [accessedData, rowPaths, flatRows]
     },
-    [debug, data, subRowsKey]
+    [debug, data, columns, subRowsKey]
```
tannerlinsley: Is the columns variable used in this block?

DDinenno: Nope, `instanceRef.current.columns` is.  

In this case it was being used as a flag to re-run the memo body when columns changes, but is mapping from the `instance.columns`.  

I could fix that.

tannerlinsley: Let's do that. I would rather have the exhaustive deps lint rule do it and have our code reference the actual `columns` variable.

PR #1444, opened 2019-08-13
Sub rows are selected if the parent row is selected.

Codar97:
Selecting a row that has subRows will now also select the subRows. It makes sense to select subRows if the parent row is selected but this functionality may not be desired for react-table.

Parent rows are also updated according to the subRows selection.

Let me know if any optimisations can be made.

### DIFF
--- a/src/plugin-hooks/tests/__snapshots__/useRowSelect.test.js.snap
+++ b/src/plugin-hooks/tests/__snapshots__/useRowSelect.test.js.snap
@@ -1,6 +1,6 @@
 // Jest Snapshot v1, https://goo.gl/fbAQLP
 
-exports[`renders a table with seletable rows 1`] = `
+exports[`renders a table with selectable rows 1`] = `
 Snapshot Diff:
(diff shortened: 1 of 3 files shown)


tannerlinsley:
This looks good to me. If possible, could you update the snapshot test for the useSelectRows hook to utilize this code?

Codar97:
Yeah, update it to test the added functionality.

Codar97:
Rebased onto master to resolve conflicts.
Updated the snapshot test for useRowSelect.

The snapshots now additionally test the following:
- A row with two subRows is selected (Snapshot 3)
- A row with two subRows is deselected (Snapshot 4)
- One of two subRows is selected (Snapshot 5)
- Both of the subRows is selected (Snapshot 6)
- One of the subrows is deselected (Snapshot 7)

This should cover the added functionality checking that the parent row updates the subRows and the subRows update the parent rows accordingly.

tannerlinsley:
Fantastic

PR #1447, opened 2019-08-14
Fix #1446: Only count visible columns for determining the colSpan

### LINKED ISSUE #1446: Header groups colSpan doesn't adjust with hidden columns
> Thanks for using the alpha version of React Table v7! We're very excited about it.

Yes, and I'm excited too! 🔥 

**Describe the bug**
When hiding columns, the colSpan of their respective headers doesn't change, resulting in columns being shifted to wrong headers.

**To Reproduce**
1. Go to https://codesandbox.io/s/tannerlinsleyreact-table-basic-b33h9
2. Play around with the checkboxes to hide the columns
3. Notice how the colSpan of the header groups doesn't change and columns get shifted to wrong headers

**Expected behavior**
The colSpan property of the header groups should be calculated with hidden columns in mind.

**Screenshots**
All columns shown
All columns show in this one.

Last name hidden
Last name is hidden, age gets shifted to the Name header group.

This bug seems to be easy to fix. I'm happy to follow up with a PR!

mbrandau:
Only count visible columns for determining the colSpan

fix #1446

### DIFF
--- a/src/hooks/useTable.js
+++ b/src/hooks/useTable.js
@@ -100,30 +100,27 @@ export const useTable = (props, ...plugins) => {
   ])
 
   // Allow hooks to decorate columns (and trigger this memoization via deps)
-  columns = React.useMemo(
-    () => {
-      if (process.env.NODE_ENV === 'development' && debug)
-        console.time('hooks.columnsBeforeHeaderGroups')
-      const newColumns = applyHooks(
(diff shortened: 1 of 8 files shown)


tannerlinsley:
Yeah, something is funky with prettier + vscode. Could even be on my machine...

PR #1451, opened 2019-08-15
improve(use-table): use single instance of defaultColumn property

hyzhak:
in case of using construction like:
```javascript
{
  defaultColumn = {}
}
```
`defaultColumn` will get new instance each time, so as result it force to recalculation each of `React.useMemo`.


tannerlinsley:
Excellent catch. Gotta love those new hooks eh? 😜

hyzhak:
@tannerlinsley oh definitely! They make code much cleaner :)

PR #1454, opened 2019-08-16
fix(use-sort-by): sorting now ignores column ids that no longer exist

paolostyle:
I have a table that dynamically generates columns (there is a switch that transposes rows to columns and vice versa). However if I sort the existing table and then click that "switch" button, I'm getting error `Cannot read property 'sortType' of undefined`, because the column by which the content was sorted before no longer exists. This PR fixes this issue - if the column no longer exists, the sorting function for it always returns 0 (so the items stay in its original place). I don't remove them from `sortBy` state, because, as I said, my table has a switch, so if I go back to previous view it will keep the sorting that I set before switching. It works the same way with filters, so I think it's fine.

By the way, I'd recommend installing Prettier as dev dependency instead of relying on globally installed one (I suppose that's how it works right now) - I modified only small part of this file but it reformatted much more than that and I guess it's because of version mismatch - I didn't have Prettier installed globally and I had to install the latest one. But that's a topic for a separate PR most likely.

edit: It seems like in alpha.22 it doesn't work with filters as well... I can prepare a separate PR for that too but now I'm not even sure if you want to support this kind of behavior, though I think in any case the component should not completely fail...

### DIFF
--- a/src/plugin-hooks/useSortBy.js
+++ b/src/plugin-hooks/useSortBy.js
@@ -209,42 +209,48 @@ function useMain(instance) {
     if (process.env.NODE_ENV === 'development' && debug)
       console.time('getSortedRows')
 
+    // Filter out sortBys that correspond to non existing columns
+    const availableSortBy = sortBy.filter(sort =>
+      columns.find(col => col.id === sort.id)
+    )
+
     const sortData = rows => {
       // Use the orderByFn to compose multiple sortBy's together.
       // This will also perform a stable sorting using the row index
       // if needed.
       const sortedData = orderByFn(
         rows,
-        sortBy.map(sort => {
+        availableSortBy.map(sort => {
           // Support custom sorting methods for each column
           const column = columns.find(d => d.id === sort.id)
 
-          if (column) {
-            const { sortType } = column
-            // Look up sortBy functions in this order:
-            // column function
-            // column string lookup on user sortType
-            // column string lookup on built-in sortType
-            // default function
-            // default string lookup on user sortType
-            // default string lookup on built-in sortType
-            const sortMethod =
-              isFunction(sortType) ||
-              (userSortTypes || {})[sortType] ||
-              sortTypes[sortType] ||
-              sortTypes.alphanumeric
-
-            // Return the correct sortFn
-            return (a, b) =>
-              sortMethod(a.values[sort.id], b.values[sort.id], sort.desc)
-          } else {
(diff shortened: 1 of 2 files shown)


tannerlinsley:
I just installed prettier and did a global format. You should be able to clean up your diffs now by doing the same.

tannerlinsley:
`yarn && yarn format` should do the trick

paolostyle:
Thanks! Done.

paolostyle:
Okay, I took a closer look at the issue with filters and it seems like it was just a mistake on your side, you were assigning value to a property of column before checking if column exists, and that column check was just below that. So I fixed that in this PR.

src/plugin-hooks/useSortBy.js
```diff
@@ -217,31 +217,38 @@ function useMain(instance) {
...
+              sortTypes.alphanumeric
+
+            // Return the correct sortFn
+            return (a, b) =>
+              sortMethod(a.values[sort.id], b.values[sort.id], sort.desc)
+          } else {
+            // if a column with id from sortBy state doesn't exist,
+            // return sorting function that keeps items in its original place
+            return () => 0
+          }
```
tannerlinsley: Another way we could do this is filter the sortBy by column existence before the map. Then we don't need to fake anything at all. Thoughts?

paolostyle: Sounds good, it's probably more performant for very large lists if there's a few of those columns that don't exist. Will take care of that.

src/plugin-hooks/useSortBy.js
```diff
@@ -209,42 +209,48 @@ function useMain(instance) {
...
-              sortTypes.alphanumeric
-
-            // Return the correct sortFn
-            return (a, b) =>
-              sortMethod(a.values[sort.id], b.values[sort.id], sort.desc)
-          } else {
-            // if a column with id from sortBy state doesn't exist,
-            // return sorting function that keeps items in its original place
-            return () => 0
+          if (!column) {
```
tannerlinsley: This is technically an impossible logical tree now ;)

paolostyle: That's true, but I guess I have some habits from TypeScript, you'd have to check column for null anyway. I can remove it, up to you.

tannerlinsley: No it's fine. The bundle size difference is negligible anyway.

PR #1461, opened 2019-08-19
Run rename-unsafe-lifecycles on v6 branch

nbar1:


tannerlinsley:
I'm not sure I can do this without doing a major semver release. This would break a lot of people as a minor or patch release

nbar1:
No worries, as v6 won't be receiving any large changes in anticipation of v7, I will import it from my repo until we upgrade.

tannerlinsley:
Sounds good. I think that's the best solution for now. Thanks!

fttriquet:
Are you going to do a 6.11 version with this ? If so when is it planned ?

Thanks

benflodge:
This fix has broken compatibility with Preact, as they are not supporting the UNSAFE* methods preactjs/preact#1047. I have fixed my react-table version to 6.10.0 as this version still works with Preact.

benflodge:
There is a now a bug for Preact tracking the broken compatibility issue https://github.com/preactjs/preact/issues/1941.

vis2992:
Seems like the preact issue is resolved now, could this PR be opened up again and released as a new major version?

PR #1474, opened 2019-08-22
fix: replace 'outline: none' with 'outline-width: 0'

### LINKED ISSUE #1471: Replace "outline: none" with "outline-width: 0"
**Affected version**
v6

**Description**
Recently "outline: none" was added for some controls at table header, which disables outline for table and there is no way to enable it back with default browser settings. This causes an issue for those who needs outlines for accessibility.

**To Reproduce**
1. Open table with input filters
2. Set focus to input filter (click on it)
3. Input is not highlighted with outline, because of "outline: none"

**Expected behavior**
"outline: none" should be replaced with "outline-width: 0", which also hides outline, but it still possible to turn it back with CSS and default browser settings, which could be adjusted by user.

**Additional context**
Outline for form controls is essential to pass accessibility requirements, there should be at least an option to turn it back with default browser styling.

stl-vitalinedzvedz:
Replace outline: none with outline-width: 0 to allow reset of outline styles to default browser settings. Closes #1471

### DIFF
--- a/src/index.styl
+++ b/src/index.styl
@@ -9,7 +9,7 @@ input-select-style()
   font-size: inherit
   border-radius: 3px
   font-weight: normal
-  outline:none
+  outline-width: 0
 
 .ReactTable
   position:relative
@@ -74,7 +74,7 @@ input-select-style()
         border-right: 0
 
     .rt-th:focus
-      outline: none
+      outline-width: 0
 
     .rt-resizable-header
       overflow: visible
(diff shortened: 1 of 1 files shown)


tannerlinsley:
Thank you for submitting this PR. I typically am not accepting any more work on the v6 branch, but I will merge this PR anyway as long as you know I will not be making any new releases on the v6 tag.  Just wanted to make you aware.

PR #1495, opened 2019-09-04
Fixing react warnings

orangecoding:
React 16.9.0(and upwards) deprecated usage of `componentWillMount` and `componentWillReceiveProps`.

This pr renames the methods to use the renamed UNSAFE ones. It suppresses the warnings. As you guys have decided to not continue working on v6, I guess this is the best we can do.

Edit: I've seen you guys still support React 15.x in package.json. Can we drop the support and only support React v16? This would make this fix much easier.


tannerlinsley:
I feel like there are more of these. Is it just those two?

orangecoding:
@tannerlinsley ups, you're right. I've updated the pr. sry :)

PR #1504, opened 2019-09-08
isAllRowsSelected can only be true if there are rows.

Codar97:
When using the useRowSelect hook if you had no rows in the table `isAllRowsSelected` would be `true` as the number of selected rows equals the number of rows and a checkbox will be checked. This shouldn't be the case. You can preview the issue [here](https://codesandbox.io/s/tannerlinsleyreact-table-row-selection-qizrj).
This is just a simple fix for that issue.


tannerlinsley:
Excellent!

PR #1517, opened 2019-09-13
V7 streaming

sivagollapalli:


tannerlinsley:
Hi! Thanks for wanting to contribute to React Table. Here are my thoughts on this PR:

- First off, great work. This looks really cool!
- I would agree that this utility feels out of scope for React Table mostly because it isn't integrating directly with React Table, and also that even for a utility hook, it is very general in terms of what it does. My guess is that this hook would be useful for anything needing streaming data.

As far as bringing this into React Table, I don't think that's something I plan on doing right now, given that the scope is pretty broad. What I would like to see though and would happy to merge would be a link in the Readme somewhere to either a blog post or video about how you built your own hook to stream data into React Table. It could serve as a good starting point for people wanting to do the same without taking root in the React Table repo itself.

PR #1521, opened 2019-09-15
TypeScript updates

ggascoigne:
I've found that trying to get solid TypeScript typings for this library
are a bit of a challenge.  The composable nature of the library means
that the types for the builtin functions are by their nature somewhat
minimal, and that the user needs to be able to extend those interfaces
to reflect the specific plugins that are in use.

With that in mind I propose something like this.

To get the best out of them, you then need to extend those interfaces
using declaration merging, see
https://www.typescriptlang.org/docs/handbook/declaration-merging.html

e.g.

```ts
declare module 'react-table' {
  export interface TableInstance<D = any>
    extends UseFiltersValues<D>,
      UsePaginationValues<D>,
      UseExpandedValues,
      UseGroupByValues {}

  export interface TableState<D = any>
    extends UsePaginationState,
      UseGroupByState,
      UseSortbyState<D>,
      UseFiltersState<D> {}
}
```

This also puts the ability to extend those types with any user defined
hooks completely in the users hands.

This gives the greatest flexibility, but I'll admit that it isn't
particularly obvious.  Perhaps a typescript readme and example is needed.

### DIFF
--- a/index.d.ts
+++ b/index.d.ts
@@ -237,7 +237,7 @@ declare module 'react-table-hooks' {
   ): TableOptions<D> & {
     toggleExpandedByPath: () => any
     expandedDepth: []
-    rows: T[]
+    rows: Row<D>[]
   }
 
   export function useTableState<D = any>(
--- a/index.d.ts
+++ b/index.d.ts
@@ -1,4 +1,4 @@
-declare module 'react-table{
+declare module 'react-table' {
   import { ReactNode, ReactText, Dispatch, SetStateAction } from 'react'
     export type SortingRule<D> = {
     id: keyof D
@@ -144,7 +144,7 @@ declare module 'react-table{
   }
 
   export interface UseFiltersValues<D> {
-    setFilter: (columnID: keyof T, value: string) => void
+    setFilter: (columnID: keyof D, value: string) => void
     setAllFilters: (values: Filters<D>) => void
     disableFilters: boolean
   }
--- a/index.d.ts
+++ b/index.d.ts
@@ -2,9 +2,10 @@ declare module 'react-table' {
   import { ReactNode, useState } from 'react'
 
   type StringKey<D> = Extract<keyof D, string>
+  type IdType<D> = StringKey<D> | string
 
-  type SortingRule<D, K extends StringKey<D> = StringKey<D>> = {
-    id: K
+  type SortingRule<D> = {
+    id: IdType<D>
     desc: boolean
   }
 
@@ -12,9 +13,7 @@ declare module 'react-table' {
 
   export type SortByFn = (a: any, b: any, desc: boolean) => 0 | 1 | -1
 
-  export type Filters<D> = {
-    [key: StringKey<D>]: string
-  }
+  export type Filters<D> = Record<IdType<D>, string>
 
   export interface Cell<D = {}> extends TableInstance<D> {
     cell: { value: any }
@@ -55,18 +54,18 @@ declare module 'react-table' {
     }
   ) => unknown
 
-  export interface HeaderColumn<D, K extends StringKey<D> = StringKey<D>> {
+  export interface HeaderColumn<D> {
     /**
      * This string/function is used to build the data model for your column.
      */
-    accessor: K | string | AccessorFn<D>
+    accessor: IdType<D> | AccessorFn<D>
     Header?: ReactNode | ((props: TableInstance<D>) => ReactNode)
     Filter?: ReactNode | ((props: TableInstance<D>) => ReactNode)
     Cell?: ReactNode | ((cell: Cell<D>) => ReactNode)
     /**
      * This is the unique ID for the column. It is used by reference in things like sorting, grouping, filtering etc.
      */
-    id?: K | string
+    id?: IdType<D>
     minWidth?: string | number
     maxWidth?: string | number
     width?: string | number
@@ -84,10 +83,10 @@ declare module 'react-table' {
 
   export type Page<D = {}> = Row<D>[]
 
-  export interface EnhancedColumn<D, K extends StringKey<D> = StringKey<D>>
+  export interface EnhancedColumn<D>
     extends Omit<Column<D>, 'columns'>,
       TableInstance<D> {
-    id: K | string
+    id: IdType<D>
     column: Column<D>
     render: (type: 'Header' | 'Filter', userProps?: any) => any
     getHeaderProps: (userProps?: any) => any
--- a/index.d.ts
+++ b/index.d.ts
@@ -1,184 +1,261 @@
-import { Dispatch, SetStateAction, ReactNode } from 'react';
+declare module 'react-table-hooks' {
+  import { ReactNode, ReactText, Dispatch, SetStateAction } from 'react'
+    export type SortingRule<D> = {
+    id: keyof D
+    desc: boolean
+  }
 
-declare module 'react-table' {
-    export interface Cell<D> {
-        render: (type: string) => any;
-        getCellProps: () => any;
-        column: Column<D>;
-        row: Row<D>;
-        state: any;
-        value: any;
+  export type SortingRules<D> = SortingRule<D>[]
+
+  export type Filters<D> = {
+    [key: keyof D]: string
+  }
+
+  export interface Cell<D = any> extends TableInstance<D> {
+    cell: { value: any }
+    column: Column<D>
+    row: Row<D>
+    render: (type: 'Cell' | 'Aggregated', userProps?: any) => any
+    isGrouped?: boolean
+    isAggregated?: boolean
+    isRepeatedValue?: boolean
+    getCellProps: () => any
+  }
+
+  export interface Row<D = any> {
+    index: number
+    cells: Cell<D>[]
+    getRowProps: (userProps?: any) => any
+    original: D
+    path: any[]
+    values: any[]
+    depth: number
+  }
+
+  export interface UseExpandedRow<D = any> {
+    subRows?: D[]
+    groupByID?: string | number
+    toggleExpanded?: () => any
+    isExpanded?: boolean
+    isAggregated?: boolean
+  }
+
+  export interface HeaderColumn<D> {
+    /**
+     * This string/function is used to build the data model for your column.
+     */
+    accessor: string | ((originalRow: D) => string)
+    Header?: ReactText | ((props: TableInstance<D>) => ReactNode)
+    Filter?: ReactText | ((props: TableInstance<D>) => ReactNode)
+    Cell?: ReactText | ((cell: Cell<D>) => ReactNode)
+    /**
+     * This is the unique ID for the column. It is used by reference in things like sorting, grouping, filtering etc.
+     */
+    id?: string | number
+    minWidth?: string | number
+    maxWidth?: string | number
+    width?: string | number
+    disableSorting?: boolean
+    canSortBy?: boolean
+    sortByFn?: (a: any, b: any, desc: boolean) => 0 | 1 | -1
+    defaultSortDesc?: boolean
+    isAggregated?: any
+  }
+
+  export interface Column<D> extends HeaderColumn<D> {
+    id?: string | number
+    show?: boolean | ((instance: TableInstance<D>) => boolean)
+    columns?: Column<D>[]
+  }
+
+  export type Page<D = any> = Row<D>[]
+
+  export interface EnhancedColumn<D> extends Pick<Column<D>, Exclude<keyof Column<D>, 'columns'>>, TableInstance<D> {
+    column: Column<D>
+    render: (type: 'Header' | 'Filter', userProps?: any) => any
+    getHeaderProps: (userProps?: any) => any
+    getSortByToggleProps: (userProps?: any) => any
+    isSorted: boolean
+    isSortedDesc: boolean
+    sortedIndex: number
+    isVisible: boolean
+    canSort?: boolean
+  }
+
+  export interface HeaderGroup<D = any> {
+    headers: EnhancedColumn<D>[]
+    getHeaderGroupProps: (userProps?: any) => any
+  }
+
+  export interface Hooks {
+    columnsBeforeHeaderGroups: any[]
+    columnsBeforeHeaderGroupsDeps: any[]
+    useMain: any[]
+    useColumns: any[]
+    useHeaders: any[]
+    useHeaderGroups: any[]
+    useRows: any[]
+    prepareRow: any[]
+    getTableProps: any[]
+    getRowProps: any[]
+    getHeaderGroupProps: any[]
+    getHeaderProps: any[]
+    getCellProps: any[]
+  }
+
+  export interface RowsProps {
+    subRowsKey: string
+  }
+
+  export interface FiltersProps {
+    filterFn: () => void
+    manualFilters: boolean
+    disableFilters: boolean
+    setFilter: () => any
+    setAllFilters: () => any
+  }
+
+  export interface UsePaginationState {
+    pageIndex: number
+    pageSize: number
+    pageCount: number
+    rowCount: number
+  }
+
+  export interface UsePaginationValues<D = any> {
+    page: Page<D>
+    pageIndex: number // yes, this is on instance and state
+    pageSize: number // yes, this is on instance and state
+    canPreviousPage: boolean
+    canNextPage: boolean
+    nextPage: () => any
+    previousPage: () => any
+    setPageSize: (size: number) => any
+    pageOptions: any[]
+    manualPagination: boolean
+    paginateExpandedRows: boolean
+    disablePageResetOnDataChange: boolean
+    pageCount: number
+    gotoPage: (page: number) => any
+  }
+
+  export interface UseFiltersState<D> {
+    filters?: Filters<D>
+  }
+
+  export interface UseFiltersValues<D> {
+    setFilter: (columnID: keyof T, value: string) => void
+    setAllFilters: (values: Filters<D>) => void
+    disableFilters: boolean
+  }
+
+  export interface UseGroupByValues {
+    groupByFn: any
+    manualGroupBy: boolean
+    disableGrouping: boolean
+    aggregations: any
+  }
+
+  export interface UseGroupByState {
+    groupBy: string[]
+    isAggregated?: boolean
+  }
+
+  export interface UseExpandedValues {
+    toggleExpanded?: () => any
+  }
+
+  export interface UseSortbyOptions {
+    sortByFn?: (a: any, b: any, desc: boolean) => 0 | 1 | -1
+    manualSorting?: boolean
+    disableSorting?: boolean
+    defaultSortDesc?: boolean
+    disableMultiSort?: boolean
+  }
+
+  export interface UseSortbyState<D> {
+    sortBy?: SortingRules<D>
+  }
+
+  export interface TableInstance<D = any> extends TableOptions<D> {
+    hooks: Hooks
+    rows: Row<D>[]
+    columns: EnhancedColumn<D>[]
+    headerGroups: HeaderGroup<D>[]
+    headers: HeaderGroup<D>[]
+    getTableProps: (userProps?: any) => any
+    getRowProps: (userProps?: any) => any
+    prepareRow: (row: Row<D>) => any
+  }
+
+  export interface TableOptions<D = any> {
+    data: D[]
+    columns: HeaderColumn<D>[]
+    state: State<D>
+    debug?: boolean
+    loading: boolean
+    defaultColumn?: Partial<Column<D>>
+  }
+
+  // eslint-disable-next-line @typescript-eslint/no-empty-interface
+  export interface TableState<D = any> {}
+
+  export type SetState<D> = (
+    updater: (old: TableState<D>) => TableState<D>,
+    actions: any
+  ) => void
+
+  export type State<D> = [TableState<D>, SetState<D>]
+
+  export function useTable<D = any>(
+    props: TableOptions<D>,
+    ...plugins: any[]
+  ): TableInstance<D>
+
+  export function useFilters<D = any>(
+    props: TableOptions<D>
+  ): TableOptions<D> & {
+    rows: Row<D>[]
+  }
+
+  export function useSortBy<D = any>(
+    props: TableOptions<D>
+  ): TableOptions<D> & {
+    rows: Row<D>[]
+  }
+
+  export function useGroupBy<D = any>(
+    props: TableOptions<D>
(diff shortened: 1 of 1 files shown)


ggascoigne:
@stramel I'd be very interested in your opinion about this.

tannerlinsley:
This definitely looks like a step in the right direction.

stramel:
Will take a look when I get back tomorrow evening

janhartmann:
Thanks for this @ggascoigne, I actually forked the library yesterday to "fix" these typings myself. Happy to see you came just before mine.

I can help out also in the longer run, if needed.

index.d.ts
```diff
@@ -1,184 +1,261 @@
...
+  export function usePagination<D = any>(
+    props: TableOptions<D>
+  ): UsePaginationValues<D>
+
+  export function useExpanded<D = any>(
+    props: TableOptions<D>
+  ): TableOptions<D> & {
+    toggleExpandedByPath: () => any
+    expandedDepth: []
+    rows: T[]
```
Abdallatif: Did you mean `Row<D>[]`, because T is not defined!
```suggestion
    rows: Row<D>[]
```

JimmyMultani:
Wow, this is great @ggascoigne ! I think some plugin hooks are missing, like `useRowSelect` & `useRowState`.

index.d.ts
ggascoigne: good catch, I use T as my generic type and though I'd found all of them :)

ggascoigne:
> Wow, this is great @ggascoigne ! I think some plugin hooks are missing, like `useRowSelect` & `useRowState`.

agreed. I'll admit that what's here is rather focused on what I've been using, there are clearly some gaps.

I'll try and take a pass at looking at the samples -  they might help me flesh out some of the gaps.  Perhaps having a couple of the samples be in typescript might be useful anyway.

index.d.ts
```diff
@@ -1,184 +1,261 @@
...
-        render: (type: string) => any;
-        getCellProps: () => any;
-        column: Column<D>;
-        row: Row<D>;
-        state: any;
-        value: any;
-    }
+  import { ReactNode, ReactText, Dispatch, SetStateAction } from 'react'
+    export type SortingRule<D> = {
+    id: keyof D
```
stramel: Spacing appears to be off in this file. Do you mind fixing it?

index.d.ts
```diff
@@ -1,184 +1,261 @@
...
-    export interface Cell<D> {
-        render: (type: string) => any;
-        getCellProps: () => any;
-        column: Column<D>;
-        row: Row<D>;
-        state: any;
-        value: any;
-    }
+  import { ReactNode, ReactText, Dispatch, SetStateAction } from 'react'
+    export type SortingRule<D> = {
```
stramel: What do y'all think about adding a second generic for the key? I find it useful even if it's a bit of a PITA to setup initially. 

`SortingRule<D, K extends Extract<keyof D, 'string'> = never>` 

The `never` just means that this one should always get the second generic. Alternatively, we can default it to `Extract<keyof D, 'string'>` which makes it optional.

Also, NOTE, the `Extract<keyof D, 'string'>` enforces that the keys that we use are strings. I have seen numbers have issues when used as keys.

stramel: What do you think of using more descriptive generics? Maybe something like `Data` and `Accessor` or `Key` if we add the second?

index.d.ts
```diff
@@ -1,184 +1,261 @@
...
-        defaultSortDesc?: boolean;
-    }
+  export type Filters<D> = {
+    [key: keyof D]: string
+  }
 
-    export interface Column<D, A extends keyof D = never> extends HeaderColumn<D, A> {
-        id: string | number;
-    }
+  export interface Cell<D = any> extends TableInstance<D> {
```
stramel: Should we make D extend an object? since we are looking for keys? This feels like a bit too broad of a type based on what we know.

index.d.ts
```diff
@@ -1,184 +1,261 @@
...
-        getRowProps: [];
-        getHeaderRowProps: [];
-        getHeaderProps: [];
-        getCellProps: [];
-    }
+  export interface HeaderColumn<D> {
+    /**
+     * This string/function is used to build the data model for your column.
+     */
+    accessor: string | ((originalRow: D) => string)
```
stramel: We should add the `K`/`Key`/`A`/`Accessor` here as well for improved intellisense.

`accessor: A | string | ((originalRow: D) => string)`

index.d.ts
```diff
@@ -1,184 +1,261 @@
...
+    Header?: ReactText | ((props: TableInstance<D>) => ReactNode)
```
stramel: `ReactText` is `string | number`. 

https://github.com/DefinitelyTyped/DefinitelyTyped/blob/239f52df4b9d258b3ed6862499b0f369ce71ffb7/types/react/index.d.ts#L186-L191

IMO if the function can return a `ReactNode` then we should be able to put a `ReactNode` in without using the function.

index.d.ts
```diff
@@ -1,184 +1,261 @@
...
+    isAggregated?: boolean
+  }
+
+  export interface UseExpandedValues {
+    toggleExpanded?: () => any
+  }
 
-    export function useTable<D>(props: TableOptions<D>, ...plugins: any[]): TableInstance<D>;
+  export interface UseSortbyOptions {
+    sortByFn?: (a: any, b: any, desc: boolean) => 0 | 1 | -1
```
stramel: Should we pull the sortByFn out as a type since it is used multiple places?

index.d.ts
```diff
@@ -1,184 +1,261 @@
...
+    defaultColumn?: Partial<Column<D>>
+  }
 
-    export function useSortBy<D>(
-        props: TableOptions<D>,
-    ): TableOptions<D> & {
-        rows: Row<D>[];
-    };
+  // eslint-disable-next-line @typescript-eslint/no-empty-interface
+  export interface TableState<D = any> {}
```
stramel: What's the benefit of this?

index.d.ts
```diff
@@ -1,184 +1,261 @@
...
+  export function useTableState<D = any>(
+    initialState?: Partial<TableState<D>>,
+    overriddenState?:  Partial<TableState<D>>,
+    options?: {
+      reducer?: (
+        oldState: TableState<D>,
+        newState: TableState<D>,
+        type: string
+      ) => any
+      useState?: Dispatch<SetStateAction<any>>
```
stramel: What do you think of using `ReturnValue<typeof useState<any>>`? I _think_ that's right? If not, this should be an array. AFAIK, this is the result of the `useState` hook?

index.d.ts
```diff
@@ -1,184 +1,261 @@
...
+  ): State<D>
 
-    export function useExpanded<D>(
-        props: TableOptions<D>,
-    ): TableOptions<D> & {
-        toggleExpandedByPath: () => any;
-        expandedDepth: [];
-        rows: [];
-    };
+  export const actions: any
```
stramel: This should be `Record<string, string>` I _believe_

index.d.ts
```diff
@@ -1,184 +1,261 @@
...
-            reducer?: (oldState: any, newState: any, type: string) => any;
-            useState?: [any, Dispatch<SetStateAction<any>>];
-        },
-    ): any;
+  export function addActions(...actions: string[]): void
 
-    export const actions: any;
-}
\ No newline at end of file
+  export const defaultState: any
```
stramel: `defaultState` is `{}` by `useTableState` but plugins extend it, so maybe `Record<string, any>` would be a better typing

index.d.ts
ggascoigne: I like the idea, but I've had so much trouble actually getting it to work.

e.g. I *think* that this should work: 

```ts
type SortingRule<D, K extends Extract<keyof D, 'string'> = never> = {
    id: K
    desc: boolean
  }

type User = {
  name: string
  email: string
}

const DEFAULT_SORT: SortingRule<User>[] = [
  {
    id: 'email',  // <- error here: Type 'string' is not assignable to type 'never'
    desc: false
  }
]
```

see: http://www.typescriptlang.org/play/#code/C4TwDgpgBAyg9gJ2ASwHYHMBKBXANhAHgBEAaKAaSggA9gJUATAZygFFaEBDAY2AIGsIIOADMopKAHImwBGnSSAfFAC8UVBABuEBMrUBvAFBQTUZAwBcFY6YYQm3KwCM4cfJ1Q2AvocOhIUACqTDqqUEYmqJwAthBWMnIYNhDRnMi48bLyhj6G3HCoMuKsAGIAgoEAMgAqAPowAPKY1VbwSPI4+ATBOooA2gC6YX02EabmVpIpabiSJDYmdg5WIpy4Id6GA4ZAA

stramel: Ah sorry, I was too lazy to look up the syntax last night and I used `Extract` incorrectly. In the example you provided, 

`type SortingRule<D, K extends Extract<keyof D, 'string'> = never> = {` 

should be 

`type SortingRule<D, K extends Extract<keyof D, string> = never> = {`

That fixes the issue with it always being `never`.

The second issue in the example you provided is you are using the typing incorrectly. The `= never` portion means that you MUST provide the second argument.

I personally, prefer the simplicity of it handling the key by default. so the rule would go from:

`type SortingRule<D, K extends Extract<keyof D, string> = never> = {`

to

`type SortingRule<D, K extends Extract<keyof D, string> = Extract<keyof D, string>> = {`

which works in the provided typescript playground you provided.

index.d.ts
```diff
@@ -1,184 +1,261 @@
...
-        getRowProps: () => any;
-        original: any;
-    }
+  export type SortingRules<D> = SortingRule<D>[]
 
-    export interface HeaderColumn<D, A extends keyof D = never> {
-        /**
-         * This string/function is used to build the data model for your column.
-         */
-        accessor: A | ((originalRow: D) => string);
```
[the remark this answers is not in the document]
stramel: Agreed, we should probably use the type of `unknown` for the return value of the function. 

We should be able to properly represent the required `id` if the accessor is a function as well. Maybe worth doing after this PR just so we can get some types in that work.

Also, we should probably update the accessor function definition since its implementation is actually closer to this,

`(originalRow: D, index: number, { subRows: D[], depth: number, data: D[] }) => unknown`

index.d.ts
ggascoigne: On that's very cool, I didn't know you could do that.

That said, on thinking about it a bit more I wonder if this isn't the wrong approach.

Because it suits my own use-case so well, I've tended to focus on the accessor being either a valid key of the type, or a function returning a value, in which case id is the key.

The problem is that that's not really the whole story.

From the docs:

>accessor: String | Function
>...
>If a string is passed, the column's value will be looked up on the original row via that key, eg. If your column's accessor is firstName then its value would be read from row['firstName']. **You can also specify deeply nested values with accessors like info.hobbies or even address[0].street**

I don't know how we can type that last part other than using string.

Thoughts?

ggascoigne: I actually like the single character types in generics, it seems to be the more prevalent typescript pattern, but I'm also not wedded to it.  If you aren't too bothered, then we can leave it as is, if you strongly want to change it, I'll be good with that too.

index.d.ts
ggascoigne: makes sense.

index.d.ts
ggascoigne: I suppose that this comment mostly applies to accessor, and I see now that you were suggesting accessor as K | string | function - which I guess improves the intellisense, though it doesn't do as much for the type checking.

index.d.ts
ggascoigne: it also causes a lot of lazy typing around unspecified generics to show up - pretty easily fixed in my code.  So all in all I suspect that this is a good thing.

index.d.ts
ggascoigne: This was out of date.  in https://github.com/tannerlinsley/react-table/commit/0a03ea3a7521d62e34c6df9dba33e0aacb35bf41 the behavior changed and allows JSX again - so yes, ReactNode is now the right type.

It actually used to be this way, then react-table stopped supporting it, now it does again.  This is better :)

index.d.ts
ggascoigne: This gives us a defined type (ok an interface), for the state, that can then be extended in the users code.

e.g.

```ts
export interface TableState<D = any>
    extends UsePaginationState,
      UseGroupByState,
      UseSortbyState<D>,
      UseFiltersState<D> {}
```

I don't know how to let the user do this in the user code without giving them a root type to extend - since it's the type that gets passed around by the library interfaces.

Perhaps this particular point would be better documented in the code rather than in the PR.

index.d.ts
stramel: I'm partial to more descriptive generics as they seem to be easier to follow. However, I'm alright with leaving them as single characters for now.

stramel: Correct. So this comment definitely applies more so to the `accessor` but can be useful for keeping the key type abstracted for multiple uses and keeping the string extraction logic in one place in the type.

index.d.ts
ggascoigne: actually I think that it should just be `typeof useState` since it's a function and defaults to React.useState.

index.d.ts
```diff
@@ -1,184 +1,261 @@
...
-        disableSorting?: boolean;
-        defaultSortDesc?: boolean;
-        disableMultiSort?: boolean;
-    }
+  export type Page<D = any> = Row<D>[]
 
-    export interface RowsProps {
-        subRowsKey: string;
-    }
+  export interface EnhancedColumn<D> extends Pick<Column<D>, Exclude<keyof Column<D>, 'columns'>>, TableInstance<D> {
```
stramel:
```suggestion
export interface EnhancedColumn<D> extends Omit<Column<D>, 'columns'>, TableInstance<D> {
````

File: index.d.ts (hunk 9 above)
Comment (stramel) [reply index.d.ts]: You're correct. `typeof useState` should work

File: index.d.ts [hunk 14]
```diff
@@ -1,184 +1,289 @@
...
 
-    export interface EnhancedColumn<D, A extends keyof D = never> extends Column<D, A> {
-        render: (type: string) => any;
-        getHeaderProps: (userProps?: any) => any;
-        getSortByToggleProps: (userProps?: any) => any;
-        sorted: boolean;
-        sortedDesc: boolean;
-        sortedIndex: number;
-    }
+  export type Filters<D> = {
```
stramel: I think this is the only thing left that I noticed. This can be `Record<StringKey<D>, string>`

index.d.ts
```diff
@@ -1,184 +1,289 @@
...
-    }
+  import { ReactNode, useState } from 'react'
 
-    export interface Row<D> {
-        index: number;
-        cells: Cell<D>[];
-        getRowProps: () => any;
-        original: any;
-    }
+  type StringKey<D> = Extract<keyof D, string>
```
stramel: Nice

stramel:
This is definitely a big improvement and moves us one step closer. Thanks for the PR and working with me on the changes. Other than that one change it LGTM.

index.d.ts
ggascoigne: actually this didn't quite work since the key for the filter should be anything that can be an ID (same with sort), and that can include string.

Either way, I pulled out an IdType and then used that.

I'm still not particularly keen on allowing string in that type, it pretty much removes most of the explicit type checking we're jumping through hoops to get, but I also don't see how to avoid it.  It's pretty pervasive.

ggascoigne:
well that one last change was a bit more than just one line :)  But yes, it was a pleasure, thank you.

index.d.ts
```diff
@@ -1,184 +1,288 @@
...
-    export function useColumns<D>(props: TableOptions<D>): TableOptions<D> & UseColumnsValues<D>;
+  export interface UsePaginationState {
+    pageIndex: number
+    pageSize: number
+    pageCount: number
+    rowCount: number
+  }
 
-    export function useRows<D>(props: TableOptions<D>): TableOptions<D> & UseRowsValues<D>;
+  export interface UsePaginationValues<D = {}> {
```
fxOne: Shouldn't this option also be on the `TableInstance` as shown in this [excample ](https://github.com/tannerlinsley/react-table/blob/master/docs/api.md#example-5)?

janhartmann:
With these typings, how do we get typings for plugins? For instance, `usePagination`?

By doing this, all pagination related properties are not typed:

```const tableState = useTableState<TableState2>({ pageIndex: 2 });
const {
    getTableProps,
    headerGroups,
    prepareRow,
    pageOptions,
    page,
    pageCount,
    pageSize,
    pageIndex,
    gotoPage,
    canNextPage,
    canPreviousPage,
    previousPage,
    nextPage,
    setPageSize
  } = useTable(
    {
      columns,
      data,
      state: tableState,
      loading: false
    },
    usePagination
  );
```

index.d.ts
```diff
@@ -1,184 +1,288 @@
...
-        setFilter: () => any;
-        setAllFilters: () => any;
-    }
+  export interface HeaderColumn<D> {
+    /**
+     * This string/function is used to build the data model for your column.
+     */
+    accessor: IdType<D> | AccessorFn<D>
+    Header?: ReactNode | ((props: TableInstance<D>) => ReactNode)
+    Filter?: ReactNode | ((props: TableInstance<D>) => ReactNode)
```
8enSmith: In the [filter example](https://codesandbox.io/s/github/tannerlinsley/react-table/tree/master/examples/filtering) there are two filter properties on the HeadColumn: filter and Filter e.g.

{
  Header: 'Status',
  accessor: 'status',
  Filter: SelectColumnFilter,
  filter: 'includes',
}

I can see that the TS definition for Filter is present, but the typing for filter appears to be missing. Is it going to be added as part of this PR?

I've raised this observation here, as opposed to creating another issue, as when I previously raised a TS issue I was told that this is the PR for the big TS rework exercise(!).

stramel:
For those concerned, I'm not saying this is THE finalized typings but a step towards the correct and finalized typings. Currently, typings aren't published in this package and the typings included in this package were valid as of alpha.2, so quite outdated.

My vote is we get this in and continue to iterate on it instead of trying to make the typings perfect in one pass.

index.d.ts
stramel: @8enSmith I believe that is a custom property that was added for demo purposes.

stramel: You're correct, it is missing that. However, I think we could merge these and then tackle covering 100% of the API.

8enSmith: The sooner these changes are merged the better. Using this library with TypeScript without accurate typings can cause confusion.

janhartmann:
I've been playing around with these and here is my current observations. :-)

* `canSort` is missing on the `HeaderColumn<T>`, but it has the `canSortBy` which is not present in the API
* No support for `useRowSelect` hook: Missing types
* No support for grouped headings, for instance:
```ts
    {
        Header: 'Name',
        columns: [
          {
            Header: 'First Name',
            accessor: 'firstName'
          },
          {
            Header: 'Last Name',
            accessor: 'lastName',
 
          },
        ]
      }
```
* There is `width`, `maxWidth` etc present on the `HeaderColumn` which the current API does not support.

I can help by updating these typings to the current API: https://github.com/tannerlinsley/react-table/blob/master/docs/api.md#column-properties-1 but should I branch off this one or wait for the merge and then fork?

stramel:
@tannerlinsley Are you good with merging this as is and iterating on it? It still isn't being published but would at least give us a more accurate representation of the current API.

tannerlinsley:
Yep. I’m good to merge.

tannerlinsley:
Keep working on them!

stramel:
Thanks to @ggascoigne for the hard work on this PR! 👏

8enSmith:
So now that these TypeScript changes have been merged, will an update to the v7 beta branch be released? I am having to heavily work around the deficiencies in the current definition file so an update is eagerly awaited.

stramel:
@8enSmith You should be able to uninstall the `@types/react-table` package, then provide the types manually. Typically, this is what I do:

1. Create a `custom_typings` folder
2. Modify `tsconfig.json` to include the newly created `custom_typings` folder.  `"include": ["custom_typings/*.d.ts", "src"],`
3. Create a file with the same name as the package, `react-table.d.ts`.
4. Insert the typings into that file.

8enSmith:
Thanks for the prompt advice @stramel . I'll do that for now and I'll keep my eyes peeled for updates to the library.

The more I use (and <3 ) TypeScript, it's a wonder that complicated libraries like this don't opt to use TS from the start! I guess using vanilla JS gives more flexibility but at a cost i.e. maintenance pains! O_o

8enSmith:
> @8enSmith You should be able to uninstall the `@types/react-table` package, then provide the types manually. Typically, this is what I do:
> 
> 1. Create a `custom_typings` folder
> 2. Modify `tsconfig.json` to include the newly created `custom_typings` folder.  `"include": ["custom_typings/*.d.ts", "src"],`
> 3. Create a file with the same name as the package, `react-table.d.ts`.
> 4. Insert the typings into that file.
> 
> I'm hesitant to publish the types as they are still not complete and there is an easy workaround to using what we currently have.

FYI, I added the following to my tsconfig and its working nicely:

"include": [
    "custom_typings/*.d.ts",
    "src"
  ],
  "paths": {
    "react-table": [
      "custom_typings/react-table.d.ts"
    ]
  }

I added the "paths" entry to override the module definition as suggested [here](https://stackoverflow.com/a/49153507/203371). Thanks again @stramel

index.d.ts
ggascoigne: No. This is discussed in the description of this PR.  I really need to put together a typescript readme and a couple of examples so that it's clearer how to compose usable types.

ggascoigne:
@janhartmann re. "With these typings, how do we get typings for plugins? For instance, usePagination?"  The short answer is do what I suggest in the description of this PR.  But I suspect that I need to describe this more clearly.

I'll put together an example and link it.

ggascoigne:
FYI there's a simple example in https://github.com/tannerlinsley/react-table/pull/1534

index.d.ts
```diff
@@ -1,184 +1,288 @@
...
+    getRowProps: (userProps?: any) => any
+    prepareRow: (row: Row<D>) => any
+  }
+
+  export interface TableOptions<D = {}> {
+    data: D[]
+    columns: HeaderColumn<D>[]
+    state: State<D>
+    debug?: boolean
+    loading: boolean
```
xiaoyu-tamu: Is there a loading prop?

stramel: There doesn't appear to be a `loading` prop anymore. See the latest typings that are in progress. #1535

xiaoyu-tamu: Thanks

PR #1522, opened 2019-09-16
useAbsoluteLayout: Enable react-table to build with divs

gargroh:
Inspired by `useFlexLayout`.
Enable react-table to build with divs
Help in integrating virtual scrolling 

Disclaimer: May require some fixes(to integrate well with column order/pinning)


tannerlinsley:
Very cool! Let's get a few other things going so we can follow along better:

- [ ] A codesandbox demonstrating the plugin (you can upload it as the example, too)
- [ ] Documentation
- [ ] At the very least, some snapshot tests (this can come last)

tannerlinsley:
It's also worth noting that a `resizeColumns` hook would technically work better with an absolute layout like this, instead of with a table.

gargroh:
@tannerlinsley , good to merge form my side, have a look when you get a chance.

tannerlinsley:
This is looking really really good. One thing that I think we could expect from a plugin like this is that rows would also be absolutely positioned. What do you think?

I know that would required that `rowHeights` be defined and everything. And obviously, if it wasn't we could support a relative row position like it is right now.

Thoughts?

gargroh:
My thoughts are - absolute positioned rows will be useful in case of virtual scrolling only, which if anybody wants(which most of them will) can use any virtualization library(like react-window etc) for it and provide the inputs needed by it(whether the grid is of variable row heights), putting that functionality in this hook will be a redundant one in my opinion.

tannerlinsley:
Yeah, that's a good point. Okay, I think I'm ready to merge this.

stramel:
IMO this seems like an unnecessary plugin to pull into this library. I thought one of the benefits to this plugin system, was you could write your own plugin hook without having to get it into this lib. Seems like it could be published as a 3rd party plugin and could be listed as an option somewhere on the README?

Also, @tannerlinsley, this seems to be referring to a plugin that doesn't exist? `usePinColumns`?

tannerlinsley:
The way I see it, this hook is similar in spirit to the much needed `useFlexLayout` hook, which is not done yet. But, both of these provide some good ground work for a few things that are currently not possible without divs:

- Row virtualization
- Resizable columns
- Sticky columns

This hook doesn't alter the core in any way, is entirely tree-shakeable and should be extremely   simple to maintain with the rest of the library. If it presents problems in the future, we can always move it out to its own import or repo, but I don't want to make that decision before it's needed.

tannerlinsley:
`usePinColumns` is still a WIP hook (has a PR opened currently) that may or may not make it into the core. It may even be repurposed as a `useStickyColumns` hook. We'll have to see.

tannerlinsley:
I'd actually like to revisit this idea and possibly make a few changes:

- Internalize the logic for `width` etc measurements into the core and support them at the column level as first-class options
- Change this hook's responsibility to simply applying those measurements as absolute layout

This is in light of other layout hooks that could be built using the same technique as this one. Mainly, I'm thinking about the `useFlexLayout` hook..

PR #1535, opened 2019-09-22
Updating types

stramel:
So some of this is still broken. I attempted to read through all the [API Docs](https://github.com/tannerlinsley/react-table/blob/master/docs/api.md) and through most of the source to rid the typings of `any` and anything missing in the typings.

Looking for feedback on how I have broken this stuff down and ways we could improve it.

I have broken all hooks down to:

- The hook function
- Options
- Hooks
- State
- Column Options
- Instance Props
- Column Props
- HeaderGroup Props
- Row Props
- Cell Props

I'm still iterating on some ways to make the plugin/composable nature easier to work with on a per-table basis.

/cc @ggascoigne


janhartmann:
Amazing, this is so close! I actually tried myself a few times to create typings for this. But the nature of the API makes it a bit hard. This, however, looks much better than what I could do.

index.d.ts
```diff
@@ -1,288 +1,473 @@
...
 
-  export function useGroupBy<D = {}>(
-    props: TableOptions<D>
-  ): TableOptions<D> & { rows: Row<D>[] }
+  type UseRowUpdater<T = unknown> = T | ((prev: T) => T)
+  type UseRowStateLocalState<D extends object, T = unknown> = Record<IdType<D>, T>
+  /* #endregion */
+  
+  /* #region useSortBy */
+  export interface useSortBy<D extends object = {}> extends PluginHook<D> {}
```
janhartmann: This should probably be `export function useSortBy<D extends object = {}>(): PluginHook<D>;` for value usage? Same or other hooks.

stramel: Good catch, fixed.

index.d.ts
```diff
@@ -1,288 +1,472 @@
...
+
+
+  /* #region useTable */
+  export function useTable<D extends object = {}>(
+    options: TableOptions<D>,
+    ...plugins: PluginHook<D>[] 
+  ): TableInstance<D>
+  
+  export type UseTableOptions<D extends object> = Partial<{
+    // columns: Column<D>[] // FIXME
```
ggascoigne: I think that this is why the old version had that weird split between Column and EnhancedColumn.

Column is that small subset of what is available to the user when defining columns, but EnhancedColumn contains all of the additional information that the library adds to columns that are passed through the rest of the api.

I do think that you are correct that it shouldn't extend TableInstance, that was my mistake.

index.d.ts
```diff
@@ -1,288 +1,522 @@
...
-  export interface UseExpandedRow<D = {}> {
-    subRows?: D[]
-    groupByID?: string | number
-    toggleExpanded?: () => any
-    isExpanded?: boolean
-    isAggregated?: boolean
+  export interface UseTableCellProps<D extends object> {
+    column: Column<D>
+    row: Row<D>
+    value: unknown
```
ggascoigne: I think that value should be replaced with:

```ts
cell: { value: any }
```

ggascoigne:
BTW as I'm going through this, I have to say that I love the organization here, this is so much cleaner that what you started with.

index.d.ts
stramel: Right, I broke it into ColumnOptions and ColumnProps. Still trying to figure out how all these pieces go together easily for the composition.

index.d.ts
ggascoigne: I also think that `UseTableCellProps` should extends TableInstance.

It's documented as
```txt
Receives the table instance and cell model as props
```

index.d.ts
ggascoigne: yeah, there's a lot to unpack in your changes :)

index.d.ts
stramel: Based on the docs, this should be `value`. https://github.com/tannerlinsley/react-table/blob/master/docs/api.md#cell-properties

And the code indicates that this is the value directly. https://github.com/tannerlinsley/react-table/blob/master/src/hooks/useTable.js#L328

This interface is used by `Cell`

stramel:
Thanks @ggascoigne! I wanted to organize it so we could easily modify and add plugins.

It wouldn't be as accurate but, for now, we could have the main interfaces extend all plugins. Which would give us a working type definition for all plugins. Then after, we could work on making it more configurable based on which plugin(s) you use.

index.d.ts
stramel: Oh, I think we're talking about 2 separate things. I'm talking about the cell properties, you're referring to the props that the Cell option on column gets.

ggascoigne: https://github.com/tannerlinsley/react-table/blob/master/src/hooks/useTable.js#L355

this calls the cell renderer passing {...instance, cell, column, row }.

ggascoigne: ah, right.

ggascoigne:
I'll admit that most of my reservations to this come from not using several of the stock plugins and using similar but different in-house ones.  It's easy to merge declarations, but I don't know how to un-merge them.

I'm trying to balance my desire to not have to deal with that, vs. wanting to be a good team player :)

We could just put all of the declaration that extends everything in a separate file?  Then there's an easy option for those that want it, but it would be easy to ignore for those that don't?

And I'll admit that we might just have to go that route in the long run, I've got no clue how to make this self configuring at this point.

index.d.ts
```diff
@@ -1,288 +1,522 @@
...
+    getCellProps: ((cell: Cell<D>, instance: TableInstance<D>) => object)[]
   }
 
-  export interface Row<D = {}> {
-    index: number
+  export type UseTableColumnOptions<D extends object> = Accessor<D> &
+    Partial<{
+      columns: Column<D>
+      show: boolean | ((instance: TableInstance<D>) => boolean)
+      Header: ComponentType<TableInstance<D>> | ReactNode
```
ggascoigne: Header is more like:

```tsx
Header: ComponentType<TableInstance<D> & { column: Column<D>}> | ReactNode
 ```

Comment (ggascoigne) [reply index.d.ts]: actually it's a bit more complex since the header.render function has userProps that also get spread here too, but I suspect that since the code that defines those props and then the code that uses them will be in the user scope, they can be defined there directly.

Comment (stramel) [reply index.d.ts]: You're correct.

Comment (stramel) [reply index.d.ts]: I suppose we could extend with `any` to account for that?

File: index.d.ts [hunk 5]
```diff
@@ -1,288 +1,523 @@
...
-  export interface Row<D = {}> {
-    index: number
+  export type UseTableColumnOptions<D extends object> = Accessor<D> &
+    Partial<{
+      columns: Column<D>
+      show: boolean | ((instance: TableInstance<D>) => boolean)
+      Header:
+        | ComponentType<TableInstance<D> & { column: Column<D> }>
+        | ReactNode
+      Cell: ComponentType<TableInstance<D> & { cell: Cell<D> }> | ReactNode
```
ggascoigne: The Cell one is weird, columns and row gets repeated everywhere.

```tsx
Cell: ComponentType<TableInstance<D> & { cell: Cell<D>, column: Column<D>, row: Row<D> }> | ReactNode
```

ggascoigne: maybe extract `TableInstance<D> & { column: Column<D> }`, it gets passed to every renderer except the cell one, and also gets passed through to the filter calls (UseFiltersColumnOptions.Filter for example).

index.d.ts
```diff
@@ -1,288 +1,523 @@
...
+    ) => void
   }
 
-  export interface UseFiltersState<D> {
-    filters?: Filters<D>
+  export interface UseFiltersColumnProps<D extends object> {
+    canFilter: boolean
+    setFilter: (filterValue: FilterType<D>) => void
+    filterValue: FilterType<D>
+    preFilteredColumnRows: Row<D>[]
```
ggascoigne: Pretty sure that this is a bug in the docs, there's no reference to this in the docs, and the example code all uses `preFilteredRows`

stramel: You're correct.

index.d.ts
ggascoigne: It would be nice to refactor these a bit so that implementation of cell or header renderers can have a tight props definition, maybe HeaderProps and CellProps, or leave the external definition as Header and Cell, but call the internal version something else since the internal ones are generally needed to cleanly define chunks of instance, and aren't very likely to need to be typed by the user.

index.d.ts
stramel: Updated the docs as well

index.d.ts
```diff
@@ -95,10 +95,8 @@ declare module 'react-table' {
     Partial<{
       columns: Column<D>
       show: boolean | ((instance: TableInstance<D>) => boolean)
-      Header:
-        | ComponentType<TableInstance<D> & { column: Column<D> }>
-        | ReactNode
-      Cell: ComponentType<TableInstance<D> & { cell: Cell<D> }> | ReactNode
+      Header: ComponentType<HeaderProps<D>> | ReactNode
+      Cell: ComponentType<Cell<D>> | ReactNode
```
ggascoigne: shouldn't this be

```ts
Cell: ComponentType< CellProps <D>> | ReactNode
```

stramel: yeah autocomplete typo

index.d.ts
ggascoigne: related to this, I think that an awful lot of the Column references in here should actually be ColumnInstance instead.  At least staring with the definition in UseTableInstanceProps - Still working though the others - it's pretty much the replacement for EnhancedColumn, and that was correctly used all over the place.

index.d.ts
```diff
@@ -1,288 +1,530 @@
...
+  // UseSortByHooks<D> &
+  // UseGroupByHooks<D> &
+  // UseExpandedHooks<D> &
+  // UseRowSelectHooks<D>
 
-  export type SortByFn = (a: any, b: any, desc: boolean) => 0 | 1 | -1
+  export type Cell<D extends object = {}> = UseTableCellProps<D>
 
-  export type Filters<D> = Record<IdType<D>, string>
+  export type Column<D extends object = {}> = UseTableColumnOptions<D>
```
ggascoigne: Column and ColumnInstance need to be interfaces so that you can do declaration merging on them, it isn't supported on types.

ggascoigne: probably Hooks does too.

ggascoigne:
I don't want to push a revision again your branch, that seems rude, but I've been working on using our new types with ported version of a couple of the original examples from this repo.

Specifically, I ported the kitchen sync example to typescript, there's a somewhat modified version of this index.d.ts file at https://github.com/ggascoigne/react-table/commit/8ad8a06888575452705f195c1b151c237d8e0aff#diff-b462adda0ffcd9c8cefed0ad03060f25

I'm not sure how much of the changes you want to incorporate, the biggest one was around switching types over to interfaces since you can't do declaration merging on types.

There was also some messiness around filters, what I have works for this example, but I wouldn't say it's the general solution.

stramel:
@ggascoigne I appreciate all the time you have been taking to review and test these changes. 

I took some of the changes you had implemented in the kitchen sink example. However, a few things didn't line up with the documentation or code so I let those out. The main thing I haven't yet changed is switching to interfaces, I plan to do this.

I have an idea for making it generic per instance but it will take some playing around. It would be using generics. Something like:

```
useTable<MyDataType, UseSortByPlugin>({ columns, data, manualSorting: true }, useSortBy)
```

where `UseSortByPlugin` would be a map of all types necessary for the plugin.

```
type UseSortByPlugin = {
  options: UseSortByOptions,
  hooks: UseSortByHooks,
  state: UseSortByState,
  columnOptions: UseSortByColumnOptions,
  instanceProps: UseSortByInstanceProps,
  columnProps: UseSortByColumnProps,
}
```

Then the `useTable` would be able to apply that type to the appropriate place. 

```
type UseTableOptions<D, P1> = {
  columns: UseTableColumnOptions<D> & P1['columnOptions']<D>
  // ...
}
```

This would obviously make the type definition pretty ugly as we would have to have  a minimum of `n + 1` generics where `n` equals the number of plugins. Then if you wanted to use more than the number of plugins that the library offers, you would then have to deal with the nasty-ness of interface merging. But this would hopefully allow you to manage each table's plugins independently instead of globally. 

Interested to hear your thoughts. I haven't implemented this, just given it a bit of thought.

ggascoigne:
That's an interesting idea, and seems like it might be a way forward, and a variant could be used for useTableState which has very similar issues.

Aside from the maintainability issue I'm not too  worried about how ugly the types end out being as long as they end out being easy to actually use - of course, that's the trick :(

I wonder if there's some way to derive the types from the list of plugins themselves, but the options available to getting the different associated types are rather limited, so some sort of key type per hook (such as the one in your example) might well be the key to getting it working.

stramel:
@ggascoigne 

> I wonder if there's some way to derive the types from the list of plugins themselves

I have had the same thought and have been brainstorming some ways around doing so... Will continue to investigate.

janhartmann:
@stramel how far off to do you think we are now? I am eager to try this out in public ;-)

stramel:
@janhartmann I'm actively working on it 😅 I modified it locally on one of my projects and it only had a bit missing. I'm working on checking the types and still playing around with the plugin architecture. 

Hope to have it ready to go by Monday if I can find some time this weekend.

janhartmann:
Awesome, nice work! 😎

ggascoigne:
I'll be away from the computer for most of the weekend I think, back on Sunday, but I can try and get another look at the latest before Friday.  One thing that is still needed is switching the type definitions over to interfaces.

index.d.ts
```diff
@@ -1,288 +1,559 @@
...
+
+  export interface UseSortByState<D extends object> {
+    sortBy: SortingRule<D>[]
+  }
+
+  export type UseSortByColumnOptions<D extends object> = Partial<{
+    disableSorting: boolean
+    sortDescFirst: boolean
+    sortInverted: boolean
+    sortType: string | Function // FIXME
```
simPod: You can use something like this https://github.com/tannerlinsley/react-table/pull/1550/files#diff-b52768974e6bc0faccb7d4b75b162c99R75

```suggestion
    sortType: string | ((rowA: Row<D>, rowB: Row<D>, columnId: IdType<D>) => number)
```

stramel:
Going to take the current state of these types and run it through a TS port of the kitchen sink example. Hopefully, that will shed any light on invalid typings. 

Once we're satisfied with the typings, I'm thinking we enable all plugins or leave it requiring interface merging so that we can release. Then we can look into making the typings configurable per usage.

index.d.ts
```diff
@@ -1,288 +1,620 @@
...
+      | ((
+          originalRow: D,
+          index: number,
+          sub: {
+            subRows: D[]
+            depth: number
+            data: D[]
+          }
+        ) => string)
     id?: IdType<D>
```
ggascoigne: I think that this is problematic.

When defining columns, you provide either, an id or an accessor.  If the accessor is a function then you also must provide an id.  However when the extended column definition is passed on after construction, regardless of the accessor type, there is an id provided.  So outside of the initial column definition, the id shouldn't be optional.

index.d.ts
```diff
@@ -1,288 +1,620 @@
...
+  type HeaderProps<D extends object> = TableInstance<D> & {
+    column: ColumnInstance<D>
+  } & Record<string, any>
+  type CellProps<D extends object> = TableInstance<D> & {
+    column: ColumnInstance<D>
+    row: Row<D>
+    cell: Cell<D>
+  } & Record<string, any>
+
+  // NOTE: Atleast one of (id | accessor | Header as string) required
```
ggascoigne: typo: atleast -> at least

index.d.ts
```diff
@@ -1,288 +1,620 @@
...
-
-  type StringKey<D> = Extract<keyof D, string>
-  type IdType<D> = StringKey<D> | string
+// TypeScript Version: 3.0
 
-  type SortingRule<D> = {
-    id: IdType<D>
-    desc: boolean
+declare module 'react-table' {
+  import { ReactNode, useState, ComponentType, MouseEvent, Props } from 'react'
```
ggascoigne: Props is unused.

index.d.ts
ggascoigne: perhaps add `id: IdType<D>` to the ColumnInstance ?

index.d.ts
```diff
@@ -1,288 +1,620 @@
...
+
+  /* #region useTable */
+  export function useTable<D extends object = {}>(
+    options: TableOptions<D>,
+    ...plugins: PluginHook<D>[]
+  ): TableInstance<D>
 
-  export type SortByFn = (a: any, b: any, desc: boolean) => 0 | 1 | -1
+  export type UseTableOptions<D extends object> = Partial<{
+    // columns: Column<D>[] // FIXME
```
ggascoigne: Pretty sure that these two need to be uncommented

index.d.ts
stramel: You're correct. I modified the `ColumnInstance` to not use the definition of `id` in `Column`

index.d.ts
stramel: Yeah, not sure how it got there. Had already removed, forgot to push

index.d.ts
stramel: Yeah, I'm going to be getting columns and state fixed while working on the kitchen sink example.

stramel:
@ggascoigne Keep bringing those comments/feedback, loving it! 👍 💯

ggascoigne:
yeah, it's nice not working in a vacuum.

stramel:
Anyone wanting to see where I'm at on the "kitchen sink" example. Can see the code here. https://github.com/stramel/react-table-example-ts

index.d.ts
```diff
@@ -1,288 +1,615 @@
...
-    headers: HeaderGroup<D>[]
-    getTableProps: (userProps?: any) => any
-    getRowProps: (userProps?: any) => any
-    prepareRow: (row: Row<D>) => any
+  ) => AggregatedValue
+  type Aggregator<D extends object> =
+    | AggregatorFn<D>
+    | DefaultAggregators
+    | string // TODO: Check that DefaultAggregators works
+  type AggregatedValue = unknown
```
ggascoigne: I'm unsure about using unknown in this and a couple of other places, on one hand it is technically correct, but it sure causes some ugly casts all over the places.  Making it be any instead makes using this much easier.

stramel: Yeah, I did notice that happening which is why some things I have switched over so far. (ie. `CellValue` and `FilterValue`) I will keep playing with the example repo and see if it becomes and issue.

stramel:
I didn't have much time to work on it this weekend. However, I did get some research into my idea for enabling per implementation type definitions. It turns out that it isn't possible currently, but hopefully will be with https://github.com/microsoft/TypeScript/issues/31843

stramel:
Alright, I _think_ we're really close. Could anyone try out the type definitions in their projects? 

Looking for feedback on the naming conventions and definitions.

index.d.ts
```diff
@@ -1,288 +1,600 @@
...
+    pluginName: string
+  }
+
+  export type SetState<D extends object> = (
+    updater: (old: TableState<D>) => TableState<D>,
+    type: keyof typeof actions
+  ) => void
+
+  export type TableStateTuple<D extends object> = [TableState<D>, SetState<D>]
+}
```
ChristianMurphy:
```suggestion
}

```

index.d.ts
```diff
@@ -1,288 +1,600 @@
...
+  /**
+   * The empty definitions of below provides a base definition for the parts used by useTable, that can then be extended in the users code.
+   *
+   * @example
+   *  export interface TableOptions<D extends object = {}}>
+   *    extends
+   *      UseExpandedOptions<D>,
+   *      UseFiltersOptions<D> {}
+   */
+  export interface TableOptions<D extends object> extends UseTableOptions<D> {}
```
ChristianMurphy: is `export` needed here?
Aren't all items in a module namespace automatically exported?

stramel: I don't believe that the items are automatically exported. I can't find any documentation referring to it  and their module example explicitly is exporting items from within the `declare module` block.

8enSmith:
> Alright, I _think_ we're really close. Could anyone try out the type definitions in their projects?
> 
> Looking for feedback on the naming conventions and definitions.
> 
> @ggascoigne

Nice work! I'll try these typings out with a substantial project tomorrow. I'll let you know how I get on.

index.d.ts
ChristianMurphy: that may be, I re-targeted #1545 at this branch, it flagged

```
ERROR: 3:1     no-single-declare-module         File has only 1 module declaration — write it as an external module. See: https://github.com/Microsoft/dtslint/blob/master/docs/no-single-declare-module.md
ERROR: 15:3    strict-export-declare-modifiers  'export' keyword is redundant here because all declarations in this module are exported automatically. If you have a good reason to export some declarations and not others, add 'export {}' to the module to shut off automatic exporting. See: https://github.com/Microsoft/dtslint/blob/master/docs/strict-export-declare-modifiers.md
```

as priority items to look at.

index.d.ts
```diff
@@ -1,288 +1,600 @@
...
+      | IdType<D>
+      | ((
+          originalRow: D,
+          index: number,
+          sub: {
+            subRows: D[]
+            depth: number
+            data: D[]
+          }
+        ) => string)
```
simPod: Docs says:

> accessor: String | Function

not sure what's correct tho.

index.d.ts
```diff
@@ -1,288 +1,600 @@
...
+      column: Column<D>,
+      instance: TableInstance<D>
+    ) => object)[]
+  }
+
+  export interface UseSortByState<D extends object> {
+    sortBy: SortingRule<D>[]
+  }
+
+  export type UseSortByColumnOptions<D extends object> = Partial<{
```
simPod: I don't know how to extend the Column type by this 🤔

index.d.ts
stramel: `accessor` is a string or a function that returns a string.

simPod: Misread the definition, cool 👍

index.d.ts
stramel: So the way to use plugins currently is by using interface merging (aka declaration merging). At the top of this file, you will find all the main types we use, and will have to extend those with each plugins' types. You will end up with something like this. https://github.com/stramel/react-table-example-ts/blob/307d0360d0756afd123ca7c8ae68eedf49ca6769/custom_typings/react-table.d.ts#L6-L64

index.d.ts
stramel: Cleaned up the majority of the linting things. I ignored some of the generic stuff.

index.d.ts
```diff
@@ -1,288 +1,591 @@
...
+
+export type AggregatorFn<D extends object> = (
+  columnValues: CellValue[],
+  rows: Row<D>[]
+) => AggregatedValue
+export type Aggregator<D extends object> =
+  | AggregatorFn<D>
+  | DefaultAggregators
+  | string // TODO: Check that DefaultAggregators works
+export type AggregatedValue = unknown
```
ggascoigne: I think that this wants to be `any` rather than `unknown`.  As it is, it propagates that unknown through `row.values[id]` accesses.

index.d.ts
```diff
@@ -1,288 +1,591 @@
...
-    debug?: boolean
-    loading: boolean
-    defaultColumn?: Partial<Column<D>>
-  }
+export interface UseGroupByColumnProps<D extends object> {
+  canGroupBy: boolean
+  isGrouped: boolean
+  groupedIndex: number
+  toggleGroupBy: () => void
+}
```
ggascoigne: I think that `getGroupByToggleProps: (props?: object) => object` also wants to be on the UseGroupByColumnProps interface since it appears to be in both places.

index.d.ts
```diff
@@ -1,288 +1,591 @@
...
+  values: Record<IdType<D>, AggregatedValue>
+  subRows: Row<D>[]
+  depth: number
+  path: IdType<D>[]
+  index: number
+}
+
+export interface UseGroupByCellProps<D extends object> {
+  isGrouped: boolean
+  isRepeatedValued: boolean
```
ggascoigne: this looks like a typo, should be `isRepeatedValue` I think.

index.d.ts
```diff
@@ -1,288 +1,591 @@
...
+  preFilteredRows: Row<D>[]
+}
 
-  export interface UseGroupByState {
-    groupBy: string[]
-    isAggregated?: boolean
-  }
+export type FilterProps<D extends object> = HeaderProps<D>
+export type FilterValue = any
+export type Filters<D extends object> = Record<IdType<D>, FilterValue>
```
ggascoigne: should this be something like

```tsx
export type Filters<D extends object> = Record<IdType<D>, FilterValue> | {}
```

since it's valid to pass in an empty object to unset all filters, in fact I'm not sure how else to unset them all.

index.d.ts
```diff
@@ -1,288 +1,591 @@
...
-      reducer?: (
-        oldState: TableState<D>,
-        newState: TableState<D>,
-        type: string
-      ) => any
-      useState?: typeof useState
-    }
-  ): State<D>
+export const actions: Record<string, string>
+export function addActions(...actions: string[]): void
```
ggascoigne: can we add:

```tsx
  export const defaultState: any
```

index.d.ts
```diff
@@ -1,288 +1,591 @@
...
+export type AggregatedValue = unknown
+/* #endregion */
+
+/* #region usePagination */
+export function usePagination<D extends object = {}>(hooks: Hooks<D>): void
+export namespace usePagination {
+  const pluginName = 'usePagination'
+}
+
+export type UsePaginationOptions<D extends object> = Partial<{
```
ggascoigne: based on https://github.com/Microsoft/TypeScript/wiki/FAQ#why-is-astring-assignable-to-anumber-for-interface-at-- I wonder if we shouldn't be more careful about not using generic signatures on interfaces that aren't generic.  I know that it makes using the interface harder since you have to be careful about when you do and don't pass in generic options, but when I've got this wrong in the past I've ended out with complex interfaces that ended out allowing pretty much any old garbage through, i.e. they resolved to `any`.

index.d.ts
```diff
@@ -1,288 +1,591 @@
...
+export function useExpanded<D extends object = {}>(hooks: Hooks<D>): void
+export namespace useExpanded {
+  const pluginName = 'useExpanded'
+}
 
-  export interface HeaderGroup<D = {}> {
-    headers: EnhancedColumn<D>[]
-    getHeaderGroupProps: (userProps?: any) => any
-  }
+export type UseExpandedOptions<D extends object> = Partial<{
```
ggascoigne: needs to be an interface

ggascoigne: I was wrong about this one, I got tricked by the Record on UseTableOptions.

index.d.ts
```diff
@@ -1,288 +1,591 @@
...
+  columns: Column<D>[]
+  state: TableStateTuple<D>
+  data: D[]
+  defaultColumn: Partial<Column<D>>
+  initialRowStateKey: IdType<D>
+  getSubRows: (row: Row<D>, relativeIndex: number) => Row<D>[]
+  getRowID: (row: Row<D>, relativeIndex: number) => string
+  debug: boolean
+}> &
+  Record<string, any>
```
ggascoigne: I'm not a big fan of the `Record` here, it effectively means that you can pass any value in to this with no easy way to tell that it's junk, and it negates the ability to extend `TableOptions`

index.d.ts
stramel: I'm unable to find anything supporting that this is on the column instance

index.d.ts
stramel: On a `Record` you can pass an empty object. I don't see a benefit in adding the union.
https://www.typescriptlang.org/play/#code/C4TwDgpgBAYglgG2BATgZygXigJQgYwHsUATAHjWBTgDsBzAGigEMaQA+AKG6JspYBcsRMnRYoAbwC+QA

index.d.ts
stramel: I removed it. It has no value publicly, AFAIK. Also, `defaultState` would be `{}` or some kind of `Record` seeing as that what is defined in the code.

index.d.ts
stramel: I still haven't ran into any issues with the `unknown` here. Can you point me to a reproducible case where this is causing problems?

index.d.ts
stramel: Other than some slight confusion of if they pass `UsePaginationOptions<Garbage>` to something they previously defined as `UsePaginationOptions<Person>`, I only see a downside in confusion as to when to use and not use the generics. It shouldn't ever resolve to `any` unless you are unioning it with `any`. 

In general, I wouldn't add the generic for cases where we don't need it. However, in our case, where the developers are having to do a bunch of interface merging, I think it makes sense to leave it until we can clean up the usage of these types.

index.d.ts
ggascoigne: There is usage like https://github.com/tannerlinsley/react-table/blob/master/examples/kitchen-sink/src/App.js#L326 in several of the examples.

index.d.ts
ggascoigne: you can, but without a cast you get:

```tsx
TS2345: Argument of type '{}' is not assignable to parameter of type 'Record<IdType<T>, any> | ((filters: Record<IdType<T>, any>) => Record<IdType<T>, any>)'.
  Type '{}' is not assignable to type '(filters: Record<IdType<T>, any>) => Record<IdType<T>, any>'.
    Type '{}' provides no match for the signature '(filters: Record<IdType<T>, any>): Record<IdType<T>, any>'.

100       setAllFilters({})
```

You can fix it with:

```tsx
      setAllFilters({} as Record<IdType<T>,any>)
```

But I thought that simply allowing the empty object was cleaner and didn't require the user to use casts.

index.d.ts
stramel: The example is out of date. The actual code doesn't add the prop to a column. It only adds it to the header.

stramel: As for the prepareRow returned a boolean part, I think they were just using the fact that it doesn't have a return to automatically fallback to the else case of the OR statement. I do this quite often even if it doesn't return a boolean.

index.d.ts
ggascoigne: It has value publicly if you are writing your own hooks.

index.d.ts
ggascoigne: that's as may be, but the function is actually on the column object (which I'll admit might actually be a header), and the code works.

I do admit that there's a bit of a conflict here between writing the interface to match what the documented interface says should be valid, vs deducing the interface based on how it's used.

index.d.ts
stramel: I'm not able to reproduce this. I dropped `setAllFilters({})` into my kitchen sink example and it works just fine. Can you provide a reproducible case?

index.d.ts
ggascoigne: So let people extend TableOptions by creating an interface that lists the extra fields that they want to add and typecheck them.

Once you add Record here, there's no way that someone who wants TableOptions to be typechecked to undo the impact of such a change.  I don't know if you can remove it again after the fact.

index.d.ts
ggascoigne: My concern is that I've seen cases where doing this just breaks the interface, and does so in rather hard to find ways.  When your type suddenly becomes an effective alias for any, your code still compiles, you've just lost the compilation changes that you thought you had.

I'll have to dig more and see if I can come up with a better example.  I'm all for simplicity, but it has to actually work, and when the docs say "don't do the thing", it seemed wise to consider what they say.

index.d.ts
stramel: It's hard to argue the code when it only specifies it on the header. 
https://github.com/tannerlinsley/react-table/blob/f70d38e061ca6bae328e141d4c3c8efa1e9c86cd/src/plugin-hooks/useGroupBy.js#L129

Maybe @tannerlinsley can shed some light into this issue?

ggascoigne:
On re-reading them, I realize that several of my comments this morning were rather snippy. Being tired and rushed is no excuse.  You are doing a ton of work on this and I really appreciate it, so I apologize for not phrasing things better.

stramel:
No worries! I realize you're just wanting to get these typings correct.

I think again, we can maybe push some of these changes off into issues where we can continue to have discussions around them. While in the meantime allowing for more progress and others to contribute.

We're getting really close but it sounds like there are going to be some more changes going in, in the near future.

index.d.ts
ggascoigne: BTW of all of the comments that I've made, this might be the one that is most important to me, the others can be fixed or modified by overrides or casts, I can't see how to undo this one if it's published like this.  I can see how it might be useful to folks, but there are other ways to achieve this effect.

index.d.ts
ggascoigne: I'd say that this is good as is.  We can change it later if there's a discovered problem.

index.d.ts
stramel: Well, at the moment, none of the typings are published. So until we do, this is revert-able.

ggascoigne: that could work, though that extra param would filter up the tree a bit.

index.d.ts
ggascoigne: huh, same problem here.  Error in one case but not in the other.  I'm guessing that this will come down to some difference in the tsconfig, my production test has several things switched off :(  and I've seen that cause issues in the past.

Since this works correctly with the default strict settings, I think that this is good as is.

index.d.ts
stramel: Unfortunately, my suggestion won't work. We'll have to revisit this

index.d.ts
```diff
@@ -1,288 +1,593 @@
...
+} & Record<string, any>
+export type CellProps<D extends object> = TableInstance<D> & {
+  column: ColumnInstance<D>
+  row: Row<D>
+  cell: Cell<D>
+} & Record<string, any>
+
+// NOTE: At least one of (id | accessor | Header as string) required
+export interface Accessor<D extends object> {
+  accessor?:
```
simPod: Before it was clear from these new typings I was using accessor to return an object `x`. Then in cell's `sortType` I was sorting using `x.a` but through `Cell` renderer showing `x.a`.

When cell `accessor` returns only string the table rows can be only sorted by that string that implies alphabetically. For example when I'm displaying formatted money I want to sort it by numerical value that said not by formatted value that servers for display purposes. It would be better to be able to format it in Cell.  
The original behaviour IMO should be kept.

My column example:

```ts
{
    Cell: ({ cell: { value } }) => formatMoney(value),
    accessor: row => somehowExtractMoneyObjectFromRow(row),
    sortType: (a, b, columnId) => {
        const a: Money = a.values[columnId];
        const b: Money = b.values[columnId];

        return Math.sign(a.amount - b.amount) as 1 | 0 | -1;
    }
}

Comment (stramel) [reply index.d.ts]: If you could provide a quick codesandbox, that would be really helpful! 

Or alternatively, providing an example of your data structure and what you would expect to be the final result vs what you're getting?

File: index.d.ts (hunk 20 above)
Comment (ggascoigne) [reply index.d.ts]: FYI I think that the issue might stem from the fact that in the main table code, the terms header and column are used somewhat interchangeably.  useGroupBy adds `getGroupByToggleProps` to an object that it calls `header` that it pulls from `flatHeaders`.  In other parts of the code `flatHeaders` is iterated as a set of `columns`.  It makes keeping track of these things in the code tricky and a bit misleading.

Comment (stramel) [reply index.d.ts]: Agreed, I found that headers are essentially forked from columns. I'm not sure if the distinction needs to be clear or not.

Comment (ggascoigne) [reply index.d.ts]: yeah, I'm not clear on that either :(

File: index.d.ts (hunk 26 above)
Comment (ggascoigne) [reply index.d.ts]: I should add that removing the Record here exposes the Partial nature of TableOptions in TableInstance and you start running into issues where you can't destructure state inline because it might be undefined -which I think that it can't be.


I think that it might be more accurate to define TableInstance as

```tsx
export interface TableInstance<D extends object = {}>
  extends Required<Omit<TableOptions<D>, 'columns'>>,
    UseTableInstanceProps<D> {}
```

stramel: Alternatively, moving the `Partial` up to where it is used in the options.

ggascoigne: The only reason that I can come up with re. choosing between adding Required or moving Parial is that if we add Required then the Required/Optional code won't need to be repeated when extending TableInstance, if we move the Partial (in more than just UseTableOptions, which we might for consistency), then that would need to be replicated when extending the interfaces for other plugins.

evark:
It is totally impossible to work with the current typings as they are in latest beta versions. 
I understand that types are WIP but I think that must be removed until we have something that works at least partially. 
IMO we need more relaxed types in the beginning and add more strict types later.

index.d.ts
simPod: Here https://codesandbox.io/s/tannerlinsleyreact-table-basic-3o9yy

I display composite value of `firstName` and `age` but sort column only by `age`

ggascoigne:
@evark do you care to be a bit more specific? I know several folks who are using these types (or at least something very close to them) quite successfully, so a blanket, “these don’t work”, whilst expressing frustration, doesn’t really help us improve things. 

I realize that we need some docs for these, since there are definitely some things here that aren’t obvious, I’ll see what I can add today.

evark:
It was not my intention to offend anyone or to throw a blanket that this is not working. Actually, types in this generic library is not an easy task.
I am trying to run a very simple example for two days, I have read the conversation in this topic trying to understand how it is possible to use the types, but many times I have to use //@ts-ignore to overcome the error messages of TS.

From basic example in useTable hook in js:
`  const { getTableProps, getTableBodyProps, headerGroups, rows, prepareRow } = useTable({
    columns,
    data,
  });`

 useTables requires two more options, state and loading. For loading any value is accepted, and for state I am using useTableState({}).
Now seems that getTableBodyProps does not exists in TableInstance.  I can remove the getTableBodyProps from my code or to bypass the problem with @ts-ignore 

This code in TS is working:
` const state = useTableState({ });
  const {
    getTableProps,
    //@ts-ignore
    getTableBodyProps,
    headerGroups,
    rows,
    prepareRow,
  } = useTable({ columns, data, state, loading: false });` 

Now if we add any plugin in useTable, how the additional values will extent the table instance? I end up to use //@ts-ignore. For example if I add usePagination, all related returning values (page, pageCount, gotoPage, etc) are not defined in TableInstance. 
I am searching this conversation and the forks of commiters to find something but for now I am using the //@ts-ignore to have my program running.

tannerlinsley:
I think we're all on the same page here, but just to clarify and re-set the stage:

- React Table is not natively typed, thus any typings you use are never going to be "perfect". Tread lightly and contribute.
- Do not rely on any types in this library as a 100% replacement for the very thorough documentation I have rigorously and exhaustively written.
- If you are not satisfied with the WIP types for React Table, then please contribute to the types or as a worst-case scenario, disable/override them.

Keep going guys! And remember to keep the spirit of the contributors guidelines and code of conduct. We're all on the same team here.

ggascoigne:
@evark Since `getTableBodyProps` was only added to the api 22 hours ago, I can't say that I'm surprised it's not in the types yet.

As for how to correctly extend types so you can have them match the plugins that you are using, that's what needs writing.

ggascoigne:
FYI https://github.com/tannerlinsley/react-table/pull/1564 - a starting place for docs.

evark:
Thanks @ggascoigne, this is something to start, for removing all these ugly //@ts-ignore lines.
By the way, which is the best method to use a local version of types in source tree? Currently, I have a copy of original index.d.ts as react-table.d.ts in my src folder and in tsconfig.json a line like this: `"exclude": ["node_modules/react-table"]`

ggascoigne:
@evark That's pretty much what I do.  Once the latest types are merged this will be comes easier, but for now you are on the right path.

index.d.ts
stramel: @simPod I was able to achieve what you wanted with the string accessor.


```js
{
  Header: "Data",
  accessor: 'age',
  Cell: ({ cell: { value }, row: { original: { firstName } }}) => {
    return firstName + " " + value;
  },
  sortType: (aRow, bRow, columnId) => {          
    const a = aRow.values[columnId];
    const b = bRow.values[columnId];
  
    return Math.sign(a - b);
  }
}
```

ggascoigne: When accessor is a function it's defined as returning the value of the cell, in the same way that row[id] returns the value of a cell.  So while I'm not sure that it should really return an object, I do think that it could reasonably return pretty much any other thing that could be a valid cell value, say number or boolean.

ggascoigne:
@stramel Where are we on this, there seem to be a few unresolved issues still here?  Would you like me to open a PR against you fork with a couple of changes to push into this PR?  I'm not sure how else to try and help.

stramel:
@ggascoigne What are the unresolved issues left that we have a consensus on how to move forward with?

index.d.ts
simPod: Thanks, using CellValue now 👍

ggascoigne:
@stramel There only two (and I suppose that the key bit there is about consensus :) ).

*  `UseTableOptions` extending `Record`, this defeats the type checking on creating tables and is easily handled by creating a type for the additional params that you want to pass in to the options.  I really care about this since it's very awkward to undo if it's in the standard types.  
* `getGroupByToggleProps` on `UseGroupByColumnProps`, I agree that this one seems odd
 but suspect that it's a sign that our differentiation between Header and Column is more strict than actually exists in the code.  This can always be added after the fact though so whilst I think that this is probably correct it's easy to work around.

Other than these two, these types work in my prod app and in the samples that I've ported to TypeScript.

tannerlinsley:

ggascoigne:
I also noticed that if you don't delete the index.d.ts from the react-table package its types also get used which masks the optional nature of state in the TableInstance. If you do delete node_modules/react-table/index.d.ts you suddenly find that you can't destructure state.  I was wondering why I was seeing this and you weren't.  I thought that since the types weren't explicitly listed in the package.json file, that they weren't exposed to typescript, that's apparently not the case.

In other words, we should fix that one as well.

stramel:
This now has the latest changes with

removal:
- `useTableState`

additions:
- `useAbsoluteLayout`
- `useBlockLayout`
- `useResizeColumns`

ggascoigne:
@stramel I'll update the docs, thank you.

stramel:
Whoops! Didn't mean to close this! 😅 

 @tannerlinsley Do you feel comfortable moving forward with this and allowing us to continue to iterate on this as a base?

tannerlinsley:
Yep. I think this is a good start. Good to merge?

stramel:
Yes, I believe everyone is on the same page with getting this merged.

tannerlinsley:
Released as a new beta! Good work guys. Seriously this is soooo cool.

PR #1536, opened 2019-09-22
fixes deprecated lifecycle method warnings in v6

pallymore:
hi @tannerlinsley  - thanks for your great work on this fantastic table component lib.

we recently upgraded our app to 16.9 and noticed the lifecycle deprecation warnings - we look forward to upgrading our tables to v7 however before it is production ready we'd like to eliminate those deprecated methods from the app so here's my attempt to fix them.

I didn't find any test suites in the repo - however I did test the changes locally in our app and didn't notice any issues. However we don't really utilize the full potential of `react-table` please let me know if there is a better way to test the changes.

### DIFF
--- a/src/lifecycle.js
+++ b/src/lifecycle.js
@@ -1,16 +1,13 @@
 export default Base =>
   class extends Base {
-    UNSAFE_componentWillMount () {
-      this.setStateWithData(this.getDataModel(this.getResolvedState(), true))
-    }
-
     componentDidMount () {
+      this.setStateWithData(this.getDataModel(this.getResolvedState(), true))
       this.fireFetchData()
     }
 
-    UNSAFE_componentWillReceiveProps (nextProps, nextState) {
-      const oldState = this.getResolvedState()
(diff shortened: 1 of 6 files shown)


src/lifecycle.js
```diff
@@ -1,16 +1,18 @@
 export default Base =>
   class extends Base {
-    UNSAFE_componentWillMount () {
+    constructor(props) {
+      super(props);
+
       this.setStateWithData(this.getDataModel(this.getResolvedState(), true))
```
RazerM: This doesn't work for me. columns is undefined in getDataModel. If I change `super()` in index.js to `super(props)`, it fails when trying to call setState in the constructor.

pallymore: 🤔 could you post the steps to reproduce this. 

I didn't realize we are calling `setState` here - which is definitely not right. I can add a function to calculate proper `state` object and use it here. (and make `this. setStateWithData` to use that to get the state)

pallymore: https://github.com/tannerlinsley/react-table/pull/1621

PR #1545, opened 2019-09-26
[WIP] test: add dtslint, and add typings test file

ChristianMurphy:
:warning: work in progress :warning: 

This
* adds [dtslint](https://github.com/Microsoft/dtslint), a tool to test typescript definitions.
* documents minimum typescript version needed to use typings
* adds a typings tests file based off the simple example.
* adds doc comments based off API docs
* tests typings on Travis CI
* updates Travis CI to test on Node 10 and 12 LTS versions

TODO
- [ ] fix typings to support simple example
- [ ] resolve dtslint errors

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -1,7 +1,10 @@
 language: node_js
 node_js:
   - "10"
+  - "12"
 
 cache: yarn
 
-script: npm test
+script:
+  - yarn test
+  - yarn dtslint
--- a/index.d.ts
+++ b/index.d.ts
@@ -1,3 +1,5 @@
+// TypeScript Version: 3.0
+
 declare module 'react-table' {
   import { ReactNode, useState } from 'react'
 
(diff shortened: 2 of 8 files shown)


stramel:
Check #1535 for updated types.

ChristianMurphy:
> I think adding the testing around it would be great

Agreed

> once we get that merged which will put us significantly closer to the actual API definition.

I trust that progress is being made, having tests can be a nice approach to measuring progress and ensuring no regressions crop up.

> As for docs in the file, I don't see a benefit to maintaining/synchronizing the docs between the API.md file and the index.d.ts. I was actually looking for a way to generate the API docs automatically based on JSDocs using index.d.ts file. I think that would be preferred.

That's the same idea I'm working toward.
Moving the docs into the typescript definition.
Then having one of:
* https://github.com/microsoft/tsdoc
* https://typedoc.org
* https://api-extractor.com/pages/setup/generating_docs

convert it into markdown, html, or any other format desired.

PR #1550, opened 2019-09-26
Fix TS types of HeaderColumn sorting options as mentioned in docs

### LINKED ISSUE #1543: Custom sorting for column does not work
I cannot set custom sorting function for column.

In types, there is defined `sortByFn`

```ts
  export interface HeaderColumn<D> {
    /**
     * This string/function is used to build the data model for your column.
     */
    accessor: IdType<D> | AccessorFn<D>
    Header?: ReactNode | ((props: TableInstance<D>) => ReactNode)
    Filter?: ReactNode | ((props: TableInstance<D>) => ReactNode)
    Cell?: ReactNode | ((cell: Cell<D>) => ReactNode)
    /**
     * This is the unique ID for the column. It is used by reference in things like sorting, grouping, filtering etc.
     */
    id?: IdType<D>
    minWidth?: string | number
    maxWidth?: string | number
    width?: string | number
    disableSorting?: boolean
    canSortBy?: boolean
    sortByFn?: SortByFn
    defaultSortDesc?: boolean
    isAggregated?: any
  }
```

However, it is not used anywhere by this library. The only way I found to implement custom sorting was adding a new `sortTypes` and then referencing it through `sortType` property on column. That is not defined on the `HeaderColumn` type though.

I'm not sure what's more updated, whether API docs or ts types.

simPod:
Fixes #1543

### DIFF
--- a/index.d.ts
+++ b/index.d.ts
@@ -59,20 +59,20 @@ declare module 'react-table' {
      * This string/function is used to build the data model for your column.
      */
     accessor: IdType<D> | AccessorFn<D>
-    Header?: ReactNode | ((props: TableInstance<D>) => ReactNode)
-    Filter?: ReactNode | ((props: TableInstance<D>) => ReactNode)
-    Cell?: ReactNode | ((cell: Cell<D>) => ReactNode)
     /**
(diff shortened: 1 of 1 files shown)


stramel:
Please see #1535

PR #1562, opened 2019-10-02
feat(useexpanded, usepagination): omit subRows from rows

huboneo:
Added useExpanded 'expandSubRows' option

- Allows consumer to opt out of adding subrows to regular rows when expanding
- Also updated usePagination to respect this flag
- Updated README

### DIFF
--- a/package-lock.json
+++ b/package-lock.json
@@ -1,6 +1,6 @@
 {
   "name": "react-table",
-  "version": "7.0.0-alpha.29",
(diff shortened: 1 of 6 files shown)


package-lock.json
```diff
@@ -1,6 +1,6 @@
 {
```
huboneo: I had to push this to get the tests to pass

tannerlinsley:
Looks good to me!

PR #1564, opened 2019-10-02
Add typescript docs

ggascoigne:
This is tied to the type implementation found in https://github.com/tannerlinsley/react-table/pull/1535.

I'm more than happy if this gets merged into that branch before merging, or whatever's convenient :)

### DIFF
--- a/react-table-config.d.ts
+++ b/react-table-config.d.ts
@@ -1,0 +1,50 @@
+declare module 'react-table' {
+  export interface TableOptions<D extends object>
+    extends UseExpandedOptions<D>,
+      UseFiltersOptions<D>,
+      UseGroupByOptions<D>,
+      UsePaginationOptions<D>,
+      UseRowSelectOptions<D>,
+      UseSortByOptions<D>,
+      UseFiltersOptions<D> {}
+
+  export interface TableInstance<D extends object = {}>
+    extends UseColumnOrderInstanceProps<D>,
+      UseExpandedInstanceProps<D>,
+      UseFiltersInstanceProps<D>,
+      UseGroupByInstanceProps<D>,
+      UsePaginationInstanceProps<D>,
+      UseRowSelectInstanceProps<D>,
+      UseRowStateInstanceProps<D>,
+      UseSortByInstanceProps<D> {}
+
+  export interface TableState<D extends object = {}>
+    extends UseColumnOrderState<D>,
+      UseExpandedState<D>,
+      UseFiltersState<D>,
+      UseGroupByState<D>,
+      UsePaginationState<D>,
+      UseRowSelectState<D>,
+      UseSortByState<D> {}
+
+  export interface Column<D extends object = {}>
+    extends UseFiltersColumnOptions<D>,
+      UseGroupByColumnOptions<D>,
+      UseSortByColumnOptions<D> {}
+
+  export interface ColumnInstance<D extends object = {}>
+    extends UseFiltersColumnProps<D>,
+      UseGroupByColumnProps<D>,
+      UseSortByColumnProps<D> {}
+
+  export interface Cell<D extends object = {}>
+    extends UseTableCellProps<D>,
+      UseGroupByCellProps<D>,
+      UseRowStateCellProps<D> {}
+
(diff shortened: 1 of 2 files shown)


simPod:
When defining

```ts
export interface TableOptions<D extends object>
  extends UsePaginationOptions<D>,
    UseSortByOptions<D> {}
```

How to use it then?

Should I also redefine 

```ts
  export function useTable<D extends object = {}>(
    options: TableOptions<D>,
    ...plugins: PluginHook<D>[]
  ): TableInstance<D>;
``` 
?

ggascoigne:
@simPod You don't have to do anything.

I know, it took me a while to get my head around what's happening.  All of the type definitions are global, so by extending the definitions of say `TableOptions`, we've updated them everywhere, including in the pre-existing definitions.

simPod:
Right, makes sense!

But what if... I use different tables with different plugins? Eg. table A uses sorting and table B uses pagination. Then I should type them separately

```ts
export interface TableOptionsSorted<D extends object> extends UseSortByOptions<D> {} // table A
```

```ts
export interface TableOptionsPaginated<D extends object> extends UsePaginationOptions<D> {} // table B
```

ggascoigne:
I mention this in the caveat section, right now you just have to define a
superset type using all of the plugins that you are using.

We just haven’t found a better solution that works yet.

simPod:
I was wondering that we could somehow allow to pass those custom types into `useTable`.  
Eg. similar as in React's `type FC<P = {}> = FunctionComponent<P>;` where you can pass custom component props.

Not sure if doable tho, don't have any experience in designing such complex types.

```ts
export interface TableComposite<D extends object = {}> {
  value: D;
  options: TableOptions<D>;
  instance: TableInstance<D>;
  state: TableState<D>;
  column: Column<D>;
  columnInstance: ColumnInstance<D>;
  cell: Cell<D>;
  row: Row<D>;
}

...

useTable<TableComposite<MyValueType>>(...)
```

stramel:
@simPod This has already been discussed and explored. 
https://github.com/tannerlinsley/react-table/pull/1535#issuecomment-534152134
https://github.com/tannerlinsley/react-table/pull/1535#issuecomment-536673912

tannerlinsley:

react-table-config.d.ts
```diff
@@ -0,0 +1,50 @@
...
+    extends UseFiltersColumnProps<D>,
+      UseGroupByColumnProps<D>,
+      UseSortByColumnProps<D> {}
+
+  export interface Cell<D extends object = {}>
+    extends UseTableCellProps<D>,
+      UseGroupByCellProps<D>,
+      UseRowStateCellProps<D> {}
+
+  export interface Row<D extends object = {}>
```
simPod: I had to add `Omit<UseTableRowProps<D>, 'state'>` for I'm using `getRowProps` and `cells` on Row. Also had to omit `state: object` because it's already in `UseRowStateRowProps` as `state: UseRowStateLocalState<D>`.

```suggestion
  export interface Row<D extends object = {}> extends UseExpandedRowProps<D>,
    UseGroupByRowProps<D>,
    UseRowSelectRowProps<D>,
    Omit<UseTableRowProps<D>, 'state'>,
    UseRowStateRowProps<D> {
  }
```

ggascoigne: I noticed this as well, and raised this on https://github.com/tannerlinsley/react-table/pull/1535, though I'm failing to find the comment right now.  I'll add it to the list again.

simPod: Cool. Everything else works for me (though I have only one table for v7). Thanks!

stramel: `state` only appears to be on `row` if you use `useRowState` plugin hook. In which, it forks the key `rowState` out of `state`.

ggascoigne:
@stramel I think that this addresses your comment from the other PR.

ggascoigne:
@stramel @tannerlinsley - is it worth merging this before the next release?

stramel:
I think so

tannerlinsley:
Okie dokie!

tannerlinsley:
Released!

PR #1568, opened 2019-10-05
add canSort and canFilter column properties support

plxel:
Added support for `canSort` and `canFilter` properties for columns

This should cover this issue https://github.com/tannerlinsley/react-table/issues/1538

### DIFF
--- a/src/plugin-hooks/useFilters.js
+++ b/src/plugin-hooks/useFilters.js
@@ -107,16 +107,23 @@ function useMain(instance) {
   }
 
   flatColumns.forEach(column => {
-    const { id, accessor, disableFilters: columnDisableFilters } = column
+    const {
+      id,
+      accessor,
+      canFilter,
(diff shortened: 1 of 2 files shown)


tannerlinsley:
I think this is a good start, but now I think there is a conflict of interest between the `canThing` and `disableThing` options existing in the same API. I think we need to consolidate these to a single `canThing`. I'll reply with more info soon.

plxel:
@tannerlinsley any news?

tannerlinsley:
Still working on this. Please stay patient. I’ll also be on vacation until November. So it won’t be until I get back.

tannerlinsley:
I've found a really good solution to this issue. Will release today.

tannerlinsley:
https://github.com/tannerlinsley/react-table/blob/master/CHANGELOG.md#700-beta13

PR #1577, opened 2019-10-08
Update typings

dbertella:
Running the basic example in a typescript file I have two issue so far
1. loading property is required when useTable() is called
```
Property 'loading' is missing in type '{ columns: Column<any>[]; data: any[]; }' but required in type 'TableOptions<any>'.
```
That's why I made it optional

2. getTableBodyProps is not added in typings
```
Property 'getTableBodyProps' does not exist on type 'TableInstance<any>'.
```


dbertella:
Given the movement about typings I guess this pr is not needed might try to contribute to other prs even if it's not really easy to understand which pr will be the final one

PR #1579, opened 2019-10-09
Remove package-lock.json

stramel:
Removes the `package-lock.json` in favor of `yarn.lock`.
Prevent generation / addition of `package-lock.json` in the future.


tannerlinsley:
Nice :)

PR #1581, opened 2019-10-09
test: validate typings using dtslint

ChristianMurphy:


tsconfig.json
```diff
@@ -0,0 +1,14 @@
+{
+  "compilerOptions": {
+    "module": "commonjs",
+    "lib": ["es6"],
+    "noImplicitAny": true,
```
stramel: This (`noImplicitAny`) to `strictNullChecks` are covered by `strict`

ChristianMurphy: good idea, refactor to use `strict`

PR #1582, opened 2019-10-09
fix react-table-config.d.ts

ggascoigne:
It turns out that pulling the types directly from a package slightly
changes how visible they are to the consuming application.  This makes
sense, previously we explicitly adding the react-table module to the
global namespace an so didn't need explicit imports, but now that's not
the case.  So add explicit imports.


dbertella:
Hey guys (@ggascoigne) wonderful job about adding the typings, is the documentation in tyescript.md outdated already after this? I'm still struggling to understand what's the right way to use this lib along typescript, also I guess it's worthy to link the typescript doc (if updated) in the main readme, what do you think?

ggascoigne:
@dbertella having seen your PR I think that you've got it all sorted out, but just in case I updated my simple example too: https://github.com/tannerlinsley/react-table/pull/1534

PR #1583, opened 2019-10-10
test: add basic typings test based off basic example

ChristianMurphy:


index.d.ts
```diff
@@ -111,7 +111,7 @@ export interface UseTableInstanceProps<D extends object> {
   rows: Row<D>[]
   getTableProps: (props?: object) => object
   getTableBodyProps: (props?: object) => object
-  prepareRow: (row: Row<D>) => void
+  prepareRow: (row: Row<D>) => unknown
```
ChristianMurphy: Maybe this could be narrowed?
The docs on what can be return leave this pretty open.
https://github.com/tannerlinsley/react-table/blob/master/docs/api.md#instance-properties

stramel: This was discussed in my PR. Instead of changing this, the example should change its code.

```ts
(row, i) => {
  prepareRow(row)
  return (
    <tr {...row.getRowProps()}>
      {row.cells.map(cell => (
        <td {...cell.getCellProps()}>{cell.render('Cell')}</td>
      ))}
    </tr>
  )
}
```

ChristianMurphy: for reference, discussion was at https://github.com/tannerlinsley/react-table/pull/1535#discussion_r330128358

> the example should change its code

I'll revert 94bd828 and apply an update to the example code.

> Or modify the example's config to not throw errors on that

When it comes to a testing with strict vs loose/flexible config, I'd prefer to keep it strict.
It makes it easier for downstream adopters to use the test code as a reference for their own code.

dbertella:
I was following the new type definitions and I came up with a bit more complex example let me know if you feel like it's worthy to include it in the repo right now is here https://github.com/dbertella/react-table-playground I used few plugins there, not super polished but it can be a starting point

ChristianMurphy:
This PR is the first in a series turning the examples into typescript test cases.
If your advanced repo is one of the examples convered to typescript, or covers cases not covered in the examples it could be a welcome addition. 👍

dbertella:
Is not exactly one of the examples but I wanted to build a use case that I need in which I combine different plugins altogether. It covers 
```
useFilters
useSortBy
useBlockLayout
useResizeColumns
```
Plus I added some custom components using styled components. What I was interested in was to understand how to use it with typescript, for instance to made it run I needed to cast column to have all the properties from the included plugins, do you think it's worthy to create a issue with this and possibly discuss in there not to pollute this pr?

ChristianMurphy:
@dbertella sounds good, feel free to open a PR.

PR #1590, opened 2019-10-11
Rahul/render row

rahul3103:


tannerlinsley:
Not sure what's going on here. But I definitely need more info to consider reading through these changes.

PR #1599, opened 2019-10-15
Added index.d.ts to the package.json

kizu:
Oops, forgot to put the definition file into the list of included files inside a package :)

### DIFF
--- a/package.json
+++ b/package.json
@@ -20,6 +20,7 @@
     "src/",
     "es/",
     "lib/",
+    "index.d.ts",
     "react-table.js",
     "react-table.min.js",
     "react-table.css",


stramel:
Thanks for the PR but v6 type definitions are hosted on DefinitelyTyped and v7 type definitions are not quite ready yet.

kizu:
This is a follow-up to the #1584 

TL;DR: if we'd like to have both v6 and v7 in one app during the migration process, we would need to have one of them renamed to something else, and in this case the typings for it won't be taken from the DefinitlyTyped, so if we would have a separate package for the v6, it would be much more convenient to have them alongside the code.

kizu:
@tannerlinsley Hey, hope you got a good vacation :) Do you have an ETA for looking at this PR and the issue with publishing of v6 I've mentioned in https://spectrum.chat/react-table/general/additional-package-just-for-the-v7~7115661c-7e60-4a61-a16f-4be7e6796349

Not in hurry, but it would be nice to know the ETA so we could plan things (we really want to start using the v7!)

tannerlinsley:
Got back from vacay a day or two ago. Still getting back up to speed, so a release is coming soon on the v6 parallel to v7 issue.

kizu:
@tannerlinsley I have noticed the https://www.npmjs.com/package/react-table-v6, but it seems that it was published not by you, but by someone else, hmm. Do you know who they are?

tklarryonline:
Thanks @kizu for your package. The https://www.npmjs.com/package/react-table-v6 does miss the `index.d.ts` file so it seems to be published before your PR.

PR #1601, opened 2019-10-15
remove Record from HeaderProps and CellProps

ggascoigne:
These definitions mask errors in the other definitions and there are better
ways of extending these.

See https://github.com/tannerlinsley/react-table/pull/1597 for an
example of an existing prototype that was missing but harder to find
because of these definitions.


stramel:
Ping @tannerlinsley

PR #1609, opened 2019-10-21
v6: prefer components in normalizeComponent

henryqdineen:
We have been using v6.10.0 and one of our cell components was changed to use `React.memo` and the table unexpectedly stopped working. It looks like other have users reported issues with `normalizeComponent` in various versions and some fixes have been made. 

Unfortunately some users have code that assumes the "components" are just a element factory and not a true React component. When these React components are defined inline this can cause  unmounts when the parent component is re-rendered. This branch will not fix that issue but I think instead users should avoid creating inline components and just use React context if you absolutely need to provide anything to these components.

The fix for us was to use `isValidElementType` from `react-is` for component detection and allow us to use `React.memo`, `React.forwardRef`, etc. We think this solution is cleaner and providers a more predictable API. We weren't comfortable with the existing `String(component).includes('.createElement')` detection and have a lot of examples where that fails.

I understand that this probably won't be merged but it could be helpful for others that might fork this lib. Thanks!

Other changes
- updated `prop-types` peer dep to `^15.7.0` because that is when `elementType` was introduced
- add `react-is` peer dep
- update column "renderer" prop types to more closely reflect the new logic in `normalizeComponent`

### DIFF
--- a/package.json
+++ b/package.json
@@ -46,9 +46,10 @@
     "classnames": "^2.2.5"
   },
   "peerDependencies": {
-    "prop-types": "^15.5.0",
+    "prop-types": "^15.7.0",
     "react": "^16.x.x",
-    "react-dom": "^16.x.x"
+    "react-dom": "^16.x.x",
+    "react-is": "^16.8.1"
   },
   "devDependencies": {
     "autoprefixer": "^6.7.0",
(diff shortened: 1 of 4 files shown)


leeyeh:
Just running into the same issue with `React.memo` and `React.forwardRef`wrapped component.

Hope this PR got landed soon.

tannerlinsley:
Make react-is a direct dependency and you've got yourself a deal.

henryqdineen:
Done! Thanks

PR #1610, opened 2019-10-22
fix: memory leak fix

### LINKED ISSUE #1608: Memory leak due to methods such as toggleGroupBy and toggleSortBy
# Using v7? Yes

Thanks for using the beta version of React Table v7! We're very excited about it.

**Describe the bug**
Memory is not released every time the hooks are run.

**To Reproduce**
https://codesandbox.io/s/tannerlinsleyreact-table-kitchen-sink-8yde5
This sandbox was modified from the kitchen sink. The only modification I made is to add a setInterval that changes the data every second.

Using chrome devtools, you can see the memory usage keeps increasing.

If you take a heap snapshot in the memory tab, expand the "Objects" tree, sort by distance, then select the object with the highest distance, you will see the following.

Screen Shot 2019-10-21 at 19 19 04
Screen Shot 2019-10-21 at 19 19 15
Screen Shot 2019-10-21 at 19 19 29

On the 3rd screenshot, you can see that there is a chain of references by toggleGroupBy. The repeated "e" variable is actually the table instance object.

What I suspect is that the method toggleGroupBy holds a reference to the previous instance object by closure, which led to a memory leak.

The same thing happens to other function variables being applied to the table instance, such as toggleSortBy and toggleExpandedByPath.

Not sure how to solve this.

jasonlaw-dev:
The instance object is shared across functions due to JS closures, causing memory leak

fix #1608


src/plugin-hooks/useSortBy.js
```diff
@@ -161,62 +161,70 @@ function useMain(instance) {
...
-          const newSortBy = sortBy.filter(d => d.id !== column.id)
-          return {
-            ...old,
-            sortBy: newSortBy,
-          }
-        }, actions.sortByChange)
+  flatHeaders.forEach(
+    // avoid sharing instance in contexts of other functions
+    (instance => column => {
+      const { accessor, disableSorting: columnDisableSorting, id } = column
```
jasonlaw-dev: I haven't thought of a better way to handle this. Wrapping the function seems to prevent memory leak.

jasonlaw-dev:
I think this PR would be a breaking change since the timing of applying custom hooks is changed.

Would love to hear suggestions about other ways to handle memory leak.

tannerlinsley:
Hmm, what an interesting problem. I honestly wish I knew more about how to compile this away without resorting to some weird trickery. Let's keep looking into this...

tannerlinsley:
Would you be willing to chat asap about how you found these? I would love to learn your jedi ways.

jasonlaw-dev:
Do you mean how I found the cause of the memory leaks?

The steps to reproduce is at #1608. The heap snapshot shows that the table instance is in the context of every function. If you put a breakpoint around those functions, you will notice the closure object includes indeed the instance object.

Screenshot 2019-11-21 at 00 06 22

The memory leak caused by closure is demonstrated in this SO answer. See the line "originalThing is used in the closure and hence ends up in the lexical environment shared by all closures in that scope".
https://stackoverflow.com/a/22146841

In our case, the cause is that we are referencing the instance which gets saved in the functions. Every time those lines are run, getGroupByToggleProps is referencing an instance object which has the old getGroupByToggleProps as property which references an older instance object.

In my PR, I avoided referring to the instance object in the inner functions like getGroupByToggleProps. Also, I had to wrap some functions in a higher-order function so that the instance object cannot be shared among other closures.

I currently don't have a better way to solve the problem. My initial idea would be passing instanceRef instead of the instance object itself to applyPropHooks. That may prevent memory leak.

tannerlinsley:
Yeah, I think I'm getting it, but still wrapping my head around it fully. I would love to come up with an easy pattern here to help me (and others) avoid creating this problem in the future.

jasonlaw-dev:
@tannerlinsley I managed to fix the memory leak issue by using React.useRef(). This way the old instance object won't persist in the closure of affected functions.

This should be an acceptable pattern to avoid memory leaks in the future.

tannerlinsley:
Oh yeah! This is much better. So nice. One stylistic thing (that I need to turn into an eslint rule): can we do React.useRef as opposed to the Import?

jasonlaw-dev:
Done!

androsj:
Great solution indeed. Good job :)

Hey @tannerlinsley I'm curious... why do you prefer calling React.something instead of direct imports?

tannerlinsley:
• It makes it clear that they are built ins
• Less imports, less shuffling


Tanner Linsley

PR #1629, opened 2019-11-04
Remove void from flow control in examples

dcchuck:
This PR would allow developers using typescript to copy/paste the examples without seeing a compiler error.

Use of void in a control flow construct - i.e. `() => void || nextFunction()` - is disallowed by typescript see: [issue](https://github.com/microsoft/TypeScript/issues/26262). As the project is in `js`, no big deal. But as a satisfied consumer of the `index.d.ts` file I figured I'd extend the thanks if you have interest in changing the syntax in the examples.

Thanks!

### DIFF
--- a/examples/absolute-layout/src/App.js
+++ b/examples/absolute-layout/src/App.js
@@ -94,7 +94,7 @@ function Table({ columns, data }) {
       <div className="rows" {...getTableBodyProps()}>
         {rows.map(
           (row, i) =>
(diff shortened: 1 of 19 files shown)


tannerlinsley:
Many are not familiar with the comma operator. What about just a regular function With a call to prepareRow and a return?

dcchuck:
Thanks! I will re-approach.

PR #1631, opened 2019-11-05
Update syntax in examples

dcchuck:
Updated previous PR to use a function call & return statement

Previous PR for reference: [link](https://github.com/tannerlinsley/react-table/pull/1629)


tannerlinsley:
Awesome :)

PR #1634, opened 2019-11-07
Adding useColumnVisibility - for column visibility toggling 

gargroh:
This provides us the capability to have 
`toggleColumnVisibility` - to toggle column visibility, available at column level
`setColumnVisibility` - set column visibility, available at column level
`setColumnVisibility` - available at instance level, expects array of column ids which are to be hidden

Can enable lib user's to implement `show all`/`hide all` easily using above API.

As other hooks takes initial state, column level `show` flag will not be applicable anymore.

### DIFF
--- a/src/hooks/useTable.js
+++ b/src/hooks/useTable.js
@@ -8,8 +8,7 @@ import {
   flexRender,
   decorateColumnTree,
   makeHeaderGroups,
-  flattenBy,
-  determineHeaderVisibility,
+  flattenBy
 } from '../utils'
 
 const propTypes = {
@@ -229,9 +228,6 @@ export const useTable = (props, ...plugins) => {
 
   instanceRef.current.rows = rows
   instanceRef.current.flatRows = flatRows
-
-  // Determine column visibility
-  determineHeaderVisibility(instanceRef.current)
 
   // Provide a flat header list for utilities
   instanceRef.current.flatHeaders = headerGroups.reduce(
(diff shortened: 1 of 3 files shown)


ivanov-v:
@gargroh any updates for this PR?)

gargroh:
This is done from my side, it needs to be reviewed, you can visit https://github.com/tannerlinsley/react-table/issues/1600 for conversation over this.

tannerlinsley:
This is going to need to be refactored to use the new reducer setup from beta.20.

src/hooks/useTable.js
```diff
@@ -230,9 +229,6 @@ export const useTable = (props, ...plugins) => {
   instanceRef.current.rows = rows
   instanceRef.current.flatRows = flatRows
 
-  // Determine column visibility
-  determineHeaderVisibility(instanceRef.current)
-
```
tannerlinsley: It almost seems like we need to add a new hook here, so we can maintain the original order of ops here. I know it's important, still trying to remember why.

gargroh: Yeah...`a new hook` will help in keeping things clear.
`determineHeaderVisibility` enables us to calculate - `Column Visibility` and `visible child columns` which is used in forming `group headers`, `width` calculation in `useAbsoluteLayout`/`useBlockLayout`, this is also re-calculated when columns are re-ordered....

tannerlinsley:
I've added on to this PR here: https://github.com/tannerlinsley/react-table/pull/1700. Let's continue discussing there.

PR #1636, opened 2019-11-10
[WIP] Pagination Typescript Example

ggascoigne:
** For discussion - not for merging **

This feels like it's a step in the right direction, but it also **doesn't
work** :(

FYI all of these changes are in the examples/pagination-ts folder and
use a local declaration file since the one in the library conflicts with
this.

I took the approach of extending the hook functions defined by the
react-table library and adding additional data to the types so that
typeof could then be used to retrieve info about the other related
extensions provided by the library.

e.g. for `usePagination`, we get:

```ts
  type usePagination<
    D extends object,
    P extends Plugins = TUsePagination
  > = PluginConfig & {
    (hooks: Hooks<D, P>): void
    options: nsUsePagination.UsePaginationOptions<D>
    state: nsUsePagination.UsePaginationState<D>
    instance: nsUsePagination.UsePaginationInstanceProps<D, P>
  }

  export const usePagination: usePagination<D, P>
```

Then, that information can be extracted on a case by case basis to
enhance the resulting types:

```ts
  type UnionToIntersection<U> = (U extends any
    ? (k: U) => void
    : never) extends ((k: infer I) => void)
    ? I
    : never

  type PluginConfig = {
    options?: any
    instance?: any
    state?: any
    cell?: any
    column?: any
    columnInstance?: any
    headerGroup?: any
    row?: any
  }

  type Plugins = PluginConfig[]

  type ArrayPick<T extends Plugins, K extends keyof T[number]> = T[number][K]

  type PluginEnhancements<
    P extends Plugins,
    K extends keyof P[number]
  > = UnionToIntersection<NonNullable<ArrayPick<P, K>>>

  export type TableOptions<
    D extends object,
    P extends Plugins
  > = UseTableOptions<D, P> & PluginEnhancements<P, 'options'>
```

I.e. a single declaration of `TableOptions` can be built up of all of the
options added by the various plugins.

That sounds great, and it works to a point.  The problem that I ran
into, and the point at which it all fell apart, was realizing that many
of the extensions return `Row` or `Column` etc., and those in turn
require knowing about the plugin definitions used.  See the `instance`
reference in `usePagination` for an example.

I **think** that the problem now is the recursive nature of the
declarations.

Anyway, for now I've run out of ideas and feel that I've hit a dead end.
I'd really appreciate some more eyes on this, I'm sure that I've missed
something and perhaps some chunk of this and a slightly different
approach can get things unblocked.

N.B.  There's a fair bit of unrelated stuff in this PR, mostly around the typescript port of the pagination example, it helps to have an example of how to use this, but as I said, this **doesn't work** so it's mostly there to help with ideas.


ggascoigne:
FYI @stramel I'd be very interested in your feedback, I'll admit that I'm hoping that something here fills a hole in something that you've been thinking about :)

ggascoigne:
@stramel did you get a chance to look at this?  Any thoughts?

stramel:
@ggascoigne Sorry for not taking a look yet. Been pretty swamped with work. Will try to take a look today or tomorrow.

stramel:
Sorry for the delay in review. I had attempted something similar to this but also couldn't get it to work. I'm not sure there currently is a dynamic solution that works. There needs to be either changes in TypeScript or a restructure of the code to be more type friendly.

tannerlinsley:
I'm currently planning a restructure of the API that will be easier to add types for hopefully. Also, types are being moved out of this repo and into Definitely Typed.

Closing this for now. Can reopen if needed.

paolostyle:
Sorry for somewhat off topic question, but where can I find some discussion or details regarding moving types to DT? I'll be working on a table in a new project (I've used this library in another project but that one still uses beta.0 and inhouse typedefs - had no time to update it), I actually installed beta.23 already yesterday but it seems like you removed the types altogether in .24 and the types for v7 are not on DefinitelyTyped, there's no PR or issue there, so I'm not sure what to do now. @ggascoigne and @stramel were doing a great job at writing type definitions (sorry if I missed someone!) as far as I can tell, are you guys going to manage them at DT or how is that going to work?

I'm also a bit concerned about your comment @tannerlinsley, I know it's still beta but I'd appreciate if the breaking changes in the API were well documented with clear migration steps, I think quite a lot of people are already using this library but keeping up with the changes is quite challenging (and not keeping up and then trying to update the code when the release is stable would probably be even more difficult).

tannerlinsley:
Since the React Table is not native TS, and I don't plan on maintaining the types myself, I've decided to outsource the types to the community and similarly host them in a more community-controlled location (DefinitelyTyped). All of the existing type work is still available in history: https://github.com/tannerlinsley/react-table/tree/d7da8b631d1e8886e7e612f161872115822aea7a and can absolutely be reused. I also think @jaredpalmer was talking on Twitter about some changes to the types that would potentially make them much better.

The difficulty in waiting for things to become stable before upgrading is certainly objective. If you wait for things to be stable, you can know that things won't change and invest yourself more into learning the API and becoming acquainted with the docs.

As for breaking changes happening during beta, I started doing my best to track them and document them in the changelog around beta.13. I realize that a lot of people were using beta already, and that's honestly what's driving most of the changes your seeing.

The work/time involved in keeping up with beta is something I think is inherent to using a beta, so I'm simply just doing my best to keep people up to date. All I'm focused on right now is getting things stable, tested and released as soon as possible.

tannerlinsley:
Also, for further discussion, we've been using this thread: https://github.com/tannerlinsley/react-table/issues/1591

PR #1646, opened 2019-11-18
Refactoring aggregations.js

IKatsuba:


tannerlinsley:
This would produce a `NaN` result if either array is empty.

IKatsuba:
```javascript
Math.min(...[]) // -Infinity
Math.max(...[]) // Infinity
```

PR #1651, opened 2019-11-21
Fix types of getRowID & getSubRows

### LINKED ISSUE #1628: Incorrect type for getRowID
The index.d.ts file lists the type of the `getRowID` function as:
`(row: Row<D>, relativeIndex: number) => string`

However, this isn't correct. It should be `(row: D, relativeIndex: number) => string`.
In the useTable hook, the `accessRow` function calls `getRowID` with a row object from the raw data passed into a hook, which is of type D. The react-table row object, which is of type `Row<D>`, is constructed below this call.

I'm happy to submit a PR to fix this issue. Thanks!

androsj:
Fixes #1628 

- These 2 methods receive the original row object (`<D>`)
- getSubRows returns an array of original rows (`Array<D>`)
- Changed return type of getRowID to match type of Row.path (`Array<IdType<D>>`)

- **_[Please confirm]_** Removed getSubRows from UseExpandedOptions as it doesn't appear to be used.

### DIFF
--- a/index.d.ts
+++ b/index.d.ts
@@ -60,8 +60,8 @@ export type UseTableOptions<D extends object> = {
   ) => TableState<D>
   defaultColumn: Partial<Column<D>>
   initialRowStateKey: IdType<D>
-  getSubRows: (row: Row<D>, relativeIndex: number) => Array<Row<D>>
-  getRowID: (row: Row<D>, relativeIndex: number) => string
+  getSubRows: (originalRow: D, relativeIndex: number) => Array<D>
+  getRowID: (originalRow: D, relativeIndex: number) => IdType<D>
   debug: boolean
 }>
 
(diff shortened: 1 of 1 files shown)


index.d.ts
```diff
@@ -210,7 +210,6 @@ export namespace useExpanded {
 }
 
 export type UseExpandedOptions<D extends object> = Partial<{
-  getSubRows: (row: Row<D>, relativeIndex: number) => Array<Row<D>>
```
androsj: Is this okay? I only saw uses of getSubRows inside useTable.

index.d.ts
```diff
@@ -60,8 +60,8 @@ export type UseTableOptions<D extends object> = {
   ) => TableState<D>
   defaultColumn: Partial<Column<D>>
   initialRowStateKey: IdType<D>
-  getSubRows: (row: Row<D>, relativeIndex: number) => Array<Row<D>>
-  getRowID: (row: Row<D>, relativeIndex: number) => string
+  getSubRows: (originalRow: D, relativeIndex: number) => Array<D>
+  getRowID: (originalRow: D, relativeIndex: number) => IdType<D>
```
androsj: Changed return type to match type of Row.path `(Array<IdType<D>>)`

PR #1677, opened 2019-12-02
setup size-plugin-bot

kuldeepkeshwar:
Items:
- [x] Add rollup plugin `rollup-plugin-size`
- [x] Add configuration file(`.github/size-plugin.yml`) for [size-plugin 🤖](https://github.com/kuldeepkeshwar/size-plugin-bot)
- [x] Install [Github App](https://github.com/apps/size-plugin)  👈@tannerlinsley

### DIFF
--- a/.github/size-plugin.yml
+++ b/.github/size-plugin.yml
@@ -1,0 +1,3 @@
+size-files:
+  - sizes-cjs.json
+  - sizes-es.json
--- a/package.json
+++ b/package.json
@@ -23,7 +23,7 @@
     "test:ci": "yarn test:jest && yarn test:types",
     "test:jest": "jest",
     "test:types": "echo '\n****\nTODO: Re-enable dtslint\n****\n'",
-    "build": "rollup -c",
+    "build": "NODE_ENV=production rollup -c",
     "start": "rollup -c -w",
     "prepare": "yarn build",
     "release": "yarn publish",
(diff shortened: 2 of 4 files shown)


tannerlinsley:
This was failing on the windows test, so I added a `cross-env` prefix. Could you pull this down and install `cross-env` as a dev dep?

tannerlinsley:
Those sizes seem... off. Is it just me?

tannerlinsley:
Screen Shot 2019-12-02 at 9 40 12 AM

kuldeepkeshwar:
@tannerlinsley this `20.3 kB` is the gzipped size. The file is not minified. bundlephobia shows `11.2
kB` (MINIFIED + GZIPPED)

stramel:
I think showing the minified + gzipped sizes would be more accurate to what we need. If I went in and added a bunch of code documentation, this would inaccurately reflect a large bump in size when in reality, none of that size is making it to the end user.

tannerlinsley:
Yep. Minzipped.

Tanner Linsley

kuldeepkeshwar:
@tannerlinsley @stramel that makes sense. currently, it is limited to `gzipped/brotli` size. 

We do have plans to add more flexibility to how sizes are calculated.

just to give a glance of how it may look like
```
    file ⏤  brotli  gzip   uncompressed 
index.js ⏤  20kB    25kB    80 kB

```

PR #1685, opened 2019-12-03
Remove tooltip if columnDisableSortBy

Maddin-619:
When sorting is disabled there is no onClick on this column, so we also did not need the tooltip.


tannerlinsley:
You’ll need to update the sorting test snapshot.

tannerlinsley:
Nice!

PR #1700, opened 2019-12-04
Add useColumnVisibility as core hook

tannerlinsley:


tannerlinsley:
Since visibility is core to the entire library, this treats it as a required plugin. What do you think? See any issues?

tannerlinsley:
@gargroh

gargroh:
It looks super cool, treating `useColumnVisibilty` as a separate hook and at the same time directly using  it in `useTable`.

This pattern is so cool, we can follow for future plugins also if those plugins play with something which are mandatory(i think that was the same reason you integrated `useTableState` initially)

Coming to file changes, two thing to mentioned -
1. How can we have bulk update on this i.e. show/hide All columns
2. It seems `Toggle column visibility` requires some more stitching

tannerlinsley:
- Not sure what else would need stitching on that part
- Show/Hide might be as simple as making a toggleAll method and spreading in the ids from the flatheaders. Thoughts?

src/hooks/useColumnVisibility.js
```diff
@@ -0,0 +1,113 @@
...
+    column.setVisibility = value => {
+      dispatch({
+        action: actions.setColumnVisibility,
+        columnId: column.id,
+        value,
+      })
+    }
+
+    column.toggleColumnVisibility = () =>
+      dispatch({ type: actions.toggleColumnVisibility, columnId: column.id })
```
gargroh: > Not sure what else would need stitching on that part

Action for `toggleColumnVisibility` in `reducerHandlers[pluginName]` does not exists...

gargroh:
> Show/Hide might be as simple as making a toggleAll method and spreading in the ids from the flatheaders. Thoughts?

That sounds good to me...

PR #1714, opened 2019-12-05
[getResetDeps] fix instance param lag

Maddin-619:
I want to use the getResetFiltersDeps to detect the state for resetting the filters. I want to reset the filters, when the user deletes all filtered items. So I must check the instance rows length to be 0. But the instance.row param on the getResetFiltersDeps function lags one render behinde. To fix this i moved the effect function at the end of the useFilter function and pass the updated row into the instance param.
I tought that this fix could by applyed to other use functions (sort, pagination, etc.)


tannerlinsley:
Yes! Wonderful fix.

PR #1727, opened 2019-12-06
Made index.d.ts into a module itself (fix module name)

gcangussu:
Types got broken for those using the package `react-table@6.11.x`. It was working fine in previous releases with `@types/react-table`. The problems arise from the  new `index.d.ts` package provided file. It doesn't have to declare a named module. It can be a module itself and Typescript will use its name on the `package.json`. Thus avoiding the problem it introduced.

The diff of the files is not good. I just removed the first and last lines, and removed one identation level of all remaining lines.


gcangussu:
Using ignore space changes to view the diff makes it much better

tannerlinsley:
I am not sure about this, but wouldn't you just be able to change the module name back to `react-table` instead of `react-table-v6` (that's a mirror package for using v6 and v7 in the same project)

gcangussu:
Then it would fail to people using `react-table-v6`. If you look at other projects you gonna see that the norm is not declaring the module on a file, but declaring as the file is the module itself.

Here is the axios typings for an example of a wide usage project with its own provided typings:
https://github.com/axios/axios/blob/master/index.d.ts

tannerlinsley:
That makes sense. As with any PR's I merge into v6, we must make sure this will be a patch or minor semver update. If so, then I"m fine merging this as long as you're willing to own it (since I am not a TS user, I can't offer much more insight here)

gcangussu:
You could make me a codeowner and I would do it as best effort.

Other option, if you don't want to worry about TS, you could just remove the file and let people use the `@types/react-table` package. Or even make it a dependency of this package, so people don't have to manually install it and you can handle better it version. This is the strategy used by `react-testing-library`: https://github.com/testing-library/react-testing-library/blob/14670debd45236d2c5d0a61a83dadc72af1bef7c/package.json#L50

tannerlinsley:
Yeah that’s the plan. Most of them are already in DT for v7. But not for v6.

Tanner Linsley

ggascoigne:
FYI I've got a PR open on DT (https://github.com/DefinitelyTyped/DefinitelyTyped/pull/40816#issuecomment-562709432) to upgrade @types/react-table to v7, moving the v6 types into a hopefully correct version specific subdirectory.  Wouldn't it make more sense to either have both version included here in this repo, or both versions over in DefinitielyTyped?  BTW I've got no idea if the v6 types over there are the same as the ones in this branch, I never used v6 so I was a bit surprised to see types for it in both places.

tannerlinsley:
Same here. I didn't even know they were over there. Obviously I would love to have them in the same place, but since v6 already had them, there is now an expectation that they will always be there without a breaking change. We should probably just backwards-compatibly update the types in the v6 branch, then sync them over to the DT types for v6 (not sure how versions work on DT though)

ggascoigne:
I'm not even sure how having both works.  If the types are in the package then adding the ones in @types will merge with them, not replace them.  Which makes me wonder if anyone was actually using the ones over on DT anyway.  There's docs there about how you are meant to obsolete and delete the ones in DT if the types move into the main package, but it looks like that was never done.

tannerlinsley:
Looks good to me. I'll merge and release a patch with all of your approvals

gcangussu:
So lets make things clearer. I was using for more than a year react-table + @types/react-table. But with release 6.11.x it broke my build. Because this package introduced an invalid `index.d.ts`.

So, the first 6.11.x release was the breaking change release. But there is hope! This is the solution to not break anything any more and fix who is broken: make @types/react-table a dependency of react-table.

This way people who is expecting types provided by this package will get it. And people using @types/react-table will get its build working again, as they already have an explicit dependency on @types/react-table.

Allright, I think this is the best approach. Types will be maintained by DefinetlyTypes as they were maintained since I started using react-table (+1 year ago) and fixes the breaking change.

Also, the types here are identical to @types/react-table. Someone probably just copied them over here.

gcangussu:
> Looks good to me. I'll merge and release a patch with all of your approvals

Great, thank you for keeping this package!

tannerlinsley:
Yes, add these DT types as a v6 dependency.

ggascoigne:
Specifically it will want to depend on the latest 6.x version of the types
otherwise it will break again when the upgraded types get merged.

tannerlinsley:
Agreed.

tannerlinsley:
This has been release as version 6.11.5

PR #1738, opened 2019-12-09
Avoid headerGroups deps change, using same by ref

gargroh:


gargroh:
By bad, this is not solution, but i hope you got what issue is here.

Problem is - these deps changes everytime, for e.g. you select any row then also these are executed and column and row calc occurs, which are not needed in this case.

tannerlinsley:
I think what your talking about is unrelated to the header deps changing all the time and more that the column visibility plugin needs a hook that is later in the lifecycle, closer to header group generation.

I’ve got a solution I’m working on. Stay tuned.

Tanner Linsley

tannerlinsley:
This should be fixed in the latest release. Thanks!

gargroh:
Hi Tanner, thanks for looking into it, but it still happens.

Problem  - I will take `useRowSelect` as example, whenever i select row, all the things in `useTable` are re-calculated, straight from `columns` hook to `accessRow`, which increases operation time to a high extent. I used to work fine with `"version": "7.0.0-rc.1"` i.e. it was not re-calculated for that.

Let me know in case you need more details on it.

tannerlinsley:
If `columns` is firing, then it has to be one of these dependencies changing: Screen Shot 2019-12-10 at 6 30 54 AM

I am not able to reproduce this yet in this codesandbox: https://codesandbox.io/s/crazy-maxwell-hl7gi.

gargroh:
This is fixed with latest version

PR #1742, opened 2019-12-09
feat: make decorateColumn and getBy public

kizu:
What: export `decorateColumn` and `getBy` from `publicUtils.js`.

Why: I have used those for my custom plugin-hooks for adding custom columns, they were exported from `utils` before, but after `7.0.0-rc.2` they're no longer available in exports.

### DIFF
--- a/src/publicUtils.js
+++ b/src/publicUtils.js
@@ -178,3 +178,107 @@ export function useConsumeHookGetter(hooks, hookName) {
   hooks[hookName] = undefined
   return getter
 }
+
+export function decorateColumn(
+  column,
+  userDefaultColumn,
+  parent,
+  depth,
+  index
+) {
+  // Apply the userDefaultColumn
+  column = { ...defaultColumn, ...userDefaultColumn, ...column }
+
+  // First check for string accessor
+  let { id, accessor, Header } = column
+
+  if (typeof accessor === 'string') {
+    id = id || accessor
+    const accessorPath = accessor.split('.')
+    accessor = row => getBy(row, accessorPath)
+  }
+
+  if (!id && typeof Header === 'string' && Header) {
+    id = Header
+  }
+
+  if (!id && column.columns) {
+    console.error(column)
+    throw new Error('A column ID (or unique "Header" value) is required!')
+  }
+
+  if (!id) {
+    console.error(column)
+    throw new Error('A column ID (or string accessor) is required!')
+  }
+
+  column = {
+    // Make sure there is a fallback header, just in case
+    Header: () => <>&nbsp;</>,
+    Footer: () => <>&nbsp;</>,
+    ...column,
+    // Materialize and override this stuff
+    id,
+    accessor,
+    parent,
+    depth,
+    index,
+  }
+
+  return column
+}
+
+const pathObjCache = new Map()
+
+export function getBy(obj, path, def) {
+  if (!path) {
+    return obj
+  }
+  const cacheKey = typeof path === 'function' ? path : JSON.stringify(path)
+
+  const pathObj =
+    pathObjCache.get(cacheKey) ||
+    (() => {
+      const pathObj = makePathArray(path)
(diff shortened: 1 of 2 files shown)


tannerlinsley:
Yeah, I remembered there were one or two people relying on these. So what I'd like to do from here is take a look at your plugin and see what we can do to reduce your reliance on these utilities. This might mean adding more hooks to suit your use case, or restructuring your plugin to better utilize some things. Could you perhaps post your plugin here so I can take a look?

kizu:
Ah, yes, maybe adding a proper extra hook would make sense there :)

I'm not sure I could post the whole plugin at the moment, it has a lot of things going on inside, but here is a minimal example of how I'm using the `decorateColumn`:

```jsx
const columnsBeforeHeaderGroups = (flatColumns, instance) => {
    const { defaultColumn = {} } = instance;

    flatColumns.push(
        utils.decorateColumn(
            {
                // my column options
            },
            defaultColumn
        )
    );

    return flatColumns;
};
```

What I basically want: for the plugin to preprocess the columns that I get from the TableOptions, in this case my plugin would add an extra column with some stuff inside. So there I need to have a new ColumnInstance, and using `decorateColumn` seems to be the only way.

In that case I don't really need a flat list of columns, so if there'd be a hook that would allow adding new columns or modify the initial columns from a plugin, before they're decorated, then I guess we won't need the public `decorateColumn`.

(About `getBy`: I don't need it yet, so it could be omitted from export, I have moved it to public utils as it is used by the `decorateColumn`).

tannerlinsley:
Here's my notes for the next release:

## 7.0.0-rc.6

- Added `columns` and `columnsDeps` hooks, which, similar to `flatColumns` and `headerGroups`, allow you to decorate (and trigger) the memoized column generation/decoration.
- Renamed `columnsBeforeHeaderGroups` and `columnsBeforeHeaderGroupsDeps` hooks to `flatColumns` and `flatColumnsDeps` respectively, which better reflects what they are used for, rather than their order, which can remain implicit.
- Added `headerGroups` and `headerGroupDeps` hooks, which, similar to `columns` and `flatColumns`, allow you to decorate (and trigger) the memoized header group generation.
- The new hook order is as follows: `columns/columnsDeps` => `flatColumns/flatColumnsDeps` => `headerGroups/headerGroupsDeps`

What do you think?

kizu:
Sounds good!

tannerlinsley:
Alright, that new api is now available in the latest release. Thanks!

kizu:
Thank you as well, would see if this would work for me tomorrow!

kizu:
@tannerlinsley I've used the `columns` hook, and I think there are two problems currently: it seems that they're not decorated properly:

1. The render methods are not applied, so if you push a new column without a `Header` and/or `Cell`, you get the `renderer error`.

2. It seems if you push to `columns` a nested structure, the columns inside do not get their parents attached properly, in the resulting flatColumns those nested columns has no `parent` field.

If you need, I could create a codesandbox with a test plugin example which would show this, but if you'll know what could be the problem from the get-go — the better (as I'll need to strip what I have now to the bare minimum to show the reduced case, haha).

tannerlinsley:
No need to create a sandbox. But could you post your code here? Even if some of it is redacted.

kizu:
Here is an extensive example: https://codesandbox.io/s/mystifying-wescoff-qxqo9

Basically, the pushed columns work very differently from those from the ones defined in the initial `columns`:

1. The groups do not work: you can see how the “Inner Age” column doesn't have a header, but if we'd move the same column inside the original columns definition, it would be there, and looking at the inner structure of the columns inside the instance there wouldn't be a `parent` on this column.
2. Header and Cell are basically required on columns. Header is not required on the group, but probably just because it is not rendered because of the first point. Removing any of those render an error.
3. Accessor can be only a function, string one doesn't work and renders an error.

tannerlinsley:
Got it! That was a perfect example. I found the issue. Fix will be out soon.

PR #1764, opened 2019-12-12
fix: render components should include memo/forwardRef

pengx17:
User should be able to pass render components wrapped by `React.memo` and `React.forwardRef` in column definitions.
Though `React.forwardRef` is not so valid here since the render function will only pass the props to the render, I think it should be safely included.

See the bugged example: https://codesandbox.io/s/fervent-night-6v86g, which should be fixed by this pr.


tannerlinsley:
Nice work :)

PR #1796, opened 2019-12-20
fix Manual Pagination - useAsyncDebounce fn not a function

ShareefAlawneh:
this pr fixes the console error that shown when using useAsyncDebounce function and causing table break, related issue: https://github.com/tannerlinsley/react-table/issues/1794


tannerlinsley:
There is a fix for this and also deeper problems with the useAsyncDebounce function shipping in the next version. Closing this in favor of those fixes. Thank you!

martindines:
Hi @tannerlinsley Has the above fix made it in to master?

PR #1810, opened 2019-12-25
Add fixed width column support to useFlexLayout

ggascoigne:
*  Allow useFlexLayout to honor canResize by calculating flex width separately from total width.
*  Update example to show the selection checkbox since that's a common fixed width use case.
*  Add example for right aligning columns.
*  Tweaked the styles for the table to better align the resize handles (since it made verifying the rest easier when they weren't misaligned by the scroll bar width)

Note that the resize behavior is still rather strange, but that's a separate problem that this change didn't really effect.

### DIFF
--- a/examples/full-width-resizable-table/package.json
+++ b/examples/full-width-resizable-table/package.json
@@ -11,7 +11,7 @@
     "react": "^16.8.6",
     "react-dom": "^16.8.6",
     "react-scripts": "3.3.0",
-    "react-table": "^7.0.0-rc.15",
+    "react-table": "latest",
     "styled-components": "^4.3.2"
   },
   "devDependencies": {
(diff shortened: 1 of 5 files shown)


examples/full-width-resizable-table/package.json
```diff
@@ -10,8 +10,8 @@
     "namor": "^1.1.2",
     "react": "^16.8.6",
     "react-dom": "^16.8.6",
-    "react-scripts": "3.0.1",
-    "react-table": "latest",
+    "react-scripts": "3.3.0",
+    "react-table": "^7.0.0-rc.15",
```
tannerlinsley: Let's try to keep the example repos on react-table@latest

PR #1842, opened 2020-01-09
fix(usefilters): fix userFilterTypes alias in setFilter action

### LINKED ISSUE #1831: useFilters - Filter autoremove is not being called
Using the latest react-table v7.

**Describe the bug**
The `autoremove` method on a custom filter type is not being called when filter values are changed. This only occurs when passing the filter type using the main options object passed to `useTable`.

**To Reproduce**
1. Define a custom filter type with an `autoremove` method. For debugging purposes, make this method do something obvious (e.g make an alert).
2. Pass this filter to react-table using the main options object passed to `useTable`.
3. Define a column that uses this filter type.
4. Edit the filter value for this column. The `autoremove` method will not be called.

**Expected behavior**
The `autoremove` method should be called.

**Codesandbox!**
I have reproduced the issue [here](https://codesandbox.io/s/white-frog-o9hn5) using the example codesandbox for `useFilters`. `fuzzyTextFilterFn.autoremove` has been updated to alert to the screen. You will see when you edit the last name column (which uses this filter) nothing happens because `autoremove` is not being called.

**Desktop**

- OS: macOS Catalina
- Browser: chrome
- Version: 78.0.3904.108

**Potential solution**
I had a look through the code for `userFilters` and I think the problem lies [here](https://github.com/tannerlinsley/react-table/blob/master/src/plugin-hooks/useFilters.js#L43). The line:

`const { flatColumns, userFilterTypes } = instance`

should be (as it is in other places in this file):

`const { flatColumns, filterTypes: userFilterTypes } = instance`

As the key passed to `useTable` is `filterTypes`, not `userFilterTypes`.

ta-anders:
Fix to correctly alias filterTypes to userFilterTypes in set filter action, so that the filter
autoremove method is called correctly

fix #1831

### DIFF
--- a/.size-snapshot.json
+++ b/.size-snapshot.json
@@ -1,13 +1,13 @@
 {
   "dist/index.js": {
-    "bundled": 113159,
-    "minified": 52747,
-    "gzipped": 13786
+    "bundled": 113233,
+    "minified": 52524,
+    "gzipped": 13836
   },
   "dist/index.es.js": {
-    "bundled": 111809,
-    "minified": 51501,
-    "gzipped": 13598,
+    "bundled": 112296,
+    "minified": 51688,
+    "gzipped": 13670,
     "treeshaked": {
       "rollup": {
         "code": 80,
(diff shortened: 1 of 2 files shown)


tannerlinsley:
Nice!

PR #1844, opened 2020-01-10
docs: link to the original blog post instead of medium

lukasbuenger:
This is a very small and opinionated suggestion regarding the «Concept» section of the docs.

Just link to the original blog post and ditch Medium, especially given the fact that you already have it online somewhere else. I really don't care about their business model, but in terms of UX/UI Medium is an abomination. I tried to read the thing on my mobile on a train ride earlier today. Ugh.
Then I found the link at the top (I know, right) that points to the original post – much cleaner, no more fighting modals and whatever shitshow is going on over there – I mean, just *readable* fcs. 

I might be unaware of some advantages you see by promoting Medium over your own, imho much better approach. If there are any, just don't mind me. 

Thanks for the project and have a nice weekend!

Regards
Lukas

Edit: Just realised, it's probably not you (the lib author) who wrote the blog post. Still, I'd rather have the link point to the original thing than Medium.

### DIFF
--- a/docs/concepts.md
+++ b/docs/concepts.md
@@ -2,7 +2,7 @@
 
 ## React Table is a "headless" UI library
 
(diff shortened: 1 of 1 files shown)


tannerlinsley:
Absolutely!

PR #1846, opened 2020-01-14
Fix useFlexLayout table min-width

mjchang:
Current implementation of useFlexLayout has an issue where styles are not applied beyond the viewport. As a result, the header borders are not stretched to the entire width of the table in the screenshot below.
 
Screen Shot 2020-01-13 at 4 23 24 PM

You can see this happening in the [full-width-resizable-table codesandbox](https://codesandbox.io/s/github/tannerlinsley/react-table/tree/master/examples/full-width-resizable-table).

The cause is because the min-width is not set for the table. It is set for the body only which is why this issue doesn't affect the it. Fixed by moving the min-width style to getTableProps and removed the now-redundant style in getTableBodyProps.

### DIFF
--- a/.size-snapshot.json
+++ b/.size-snapshot.json
@@ -1,13 +1,13 @@
 {
   "dist/index.js": {
-    "bundled": 113233,
-    "minified": 52524,
-    "gzipped": 13836
+    "bundled": 113228,
+    "minified": 52519,
+    "gzipped": 13842
   },
   "dist/index.es.js": {
-    "bundled": 112296,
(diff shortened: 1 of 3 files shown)


tannerlinsley:
This issue may be fixed with the latest master commits (coming from unrelated refactors/PRs). Please check and reopen this PR if this is still an issue :)

mjchang:
v7.0.0-rc.16 still exhibits this issue so recreated this PR in https://github.com/tannerlinsley/react-table/pull/1924

PR #1858, opened 2020-01-18
Update App.js

ovieh:
Fixed typo.


examples/editable-data/src/App.js
```diff
@@ -63,7 +63,7 @@ const EditableCell = ({
     updateMyData(index, id, value)
   }
 
-  // If the initialValue is changed externall, sync it up with our state
+  // If the initialValue is changed external, sync it up with our state
```
joaquinox: I think this actually needs to be `externally` and not `external`

PR #1864, opened 2020-01-23
Adds default value to filter

ryan-sandy:
In a production build of a private application I get an undefined error. This error is thrown from line 99 of the `useFilter` plugin. The `filters` variable is an empty string and `filter` is not a function. Hence the error.

I am having trouble identifying the root cause of this bug. It only occurs in my production minified application. My development build is fine. Worse, I cannot reproduce the issue using the example in this library. I will try to isolate the bug in the future but debugging in a production-minified file is hard. Defaulting `filters` to an array resolves the issue.

**tl;dr**: Please allow this default value so I don't have to fix my broken code.

### DIFF
--- a/.size-snapshot.json
+++ b/.size-snapshot.json
@@ -1,13 +1,13 @@
 {
   "dist/index.js": {
-    "bundled": 113233,
-    "minified": 52524,
-    "gzipped": 13836
+    "bundled": 113250,
+    "minified": 52535,
+    "gzipped": 13844
   },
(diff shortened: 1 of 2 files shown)


tannerlinsley:
This is not a bug. When calling `setAllFilters`, you must call it with an array of filters, even if that array is empty.

ryan-sandy:
@tannerlinsley Thank you for answering the question! I have submitted a PR that adds this to the documentation.

PR #1874, opened 2020-01-29
wrap instance dispatchers in useCallback

ggascoigne:
addresses https://github.com/tannerlinsley/react-table/issues/1808

Several of the instance methods added by plugins wrap their state update
functions in useCallback, the ones that didn't do this can lead to problems
where when they are used in a component, for instance in a hook, their
required presence in the dependency array causes an infinite loop as
executing the function triggers an instance update and a new function
generation.  This PR addresses this.


VanTanev:
@ggascoigne Hey, just wanted to mention, you missed the column instance `setFilter` here:
https://github.com/tannerlinsley/react-table/pull/1874/files#diff-8fcb4d3631040704a13e857438931e27R166

Although, I'm kind of unsure if `React.useCallback` can be used there, we might need a custom way keep the function identity the same.

tannerlinsley:
@VanTanev If you need the column level `setFilter` function to be stable, you may need to wrap it in a `useRef` in the render function where you are using it and just keep it up to date manually until we can figure out a better way. I would love to see how you're using it too.

VanTanev:
@tannerlinsley I have a Filter like this, where I manually set the filterValue to `undefined`. It might not actually be necessary:
```typescript
function CheckboxListFilter<D extends DataItem>({
    column,
    preFilteredRows,
}: FilterProps<D>): React.ReactElement | null {
    const { filterValue, setFilter } = column
    // TODO instead of clearing the filters when table data changes,
    // we should create an intersection of the new and old values, maybe?
    React.useEffect(() => {
        setFilter(undefined)
        // https://github.com/tannerlinsley/react-table/issues/1808
        // eslint-disable-next-line react-hooks/exhaustive-deps
    }, [preFilteredRows])

    // ...
}
```

This is a dropdown multi-select filter built from the values in the column. Right now we reset it to empty when the table data changes, but I want to make it keep the intersection of the new table data and the old filter values in the future, for which I would still need the same `React.useEffect()` that uses `setFilter()` inside its body.

PR #1879, opened 2020-01-31
Add option to allow toggleRowSelect to not select subRows

ggascoigne:
If you are using manualGrouping, it's convenient to be able
to select the group without selecting all of the children.

This adds an instance option `selectChildRows`, which defaults to true,
that can be set to false to allow selection of the group without it
selecting all of the children.


tannerlinsley:
What a fantastic addition :)

PR #1894, opened 2020-02-09
Fix server-side pagination

Andrew-Kang-G:
When the last page from new props is smaller than the current page in the page box, the current page needs to be the last page. This happens when new datasets from the server side get smaller than previous ones by using search boxes.

image


briboles:
Is there any plan for when this update will be published?

cuellarj:
Bumping as it's been a while. 
@tannerlinsley are you planning to publish this fix or will there be no new releases of the v6 branch?

PR #1906, opened 2020-02-14
Add `Column` data in `autoRemove` of Filters

gargroh:
To auto remove `filters` using `autoRemove` fn on `ColumnFilter`, there may be some use-cases which require `Column` metadata to be available for making decisions.


tannerlinsley:
Nice Addition :)

PR #1907, opened 2020-02-14
feat(aggreations): add new uniqueCountNotEmpty and countNotEmpty aggr…

rturk:
Adds new aggregations for nom empty values

### DIFF
--- a/src/aggregations.js
+++ b/src/aggregations.js
@@ -18,6 +18,16 @@ export function uniqueCount(values) {
   return new Set(values).size
 }
 
+export function uniqueCountNotEmpty (values) {
+  const filtred = values.filter(el => !!el);
+  return new Set(filtred).size;
+}
+
 export function count(values) {
   return values.length
 }
+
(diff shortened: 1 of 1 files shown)


rturk:
This PR adds two new aggregators, similar to existing ones that will 

- `uniqueCount` extended aggregator `uniqueCountNotEmpty`
- `count` extended aggregator `uniqueCountNotEmpty`

tannerlinsley:
I love these! However, I don't think they need to be available in the main package like the other standard ones. They can easily be added to the wiki for reference and then added to a project as needed. Also, take note that the aggregation function signature has changed in the latest commits to master. Take a look at the changelog and the `src/aggregations.js` file for more info.

rturk:
@tannerlinsley happy to create a separated repo or export them as standalone items. 

IMO some filters are tricky to build and test. Allocate them in a single source where the community can review and enhance seems to be beneficial to all users

PR #1910, opened 2020-02-14
Fixed example

pmorgan3:
The CodeSandbox example didn't work because hooks.allColumns doesn't exist. I changed it to hooks.columns and badabing badaboom it runs perfectly


tannerlinsley:
All columns is actually correct, it just doesn’t exist yet. A release needs to be made first.

pmorgan3:
@tannerlinsley my mistake! Thanks for the quick response.

PR #1918, opened 2020-02-18
useRowSelect: remove redundant checks

kizu:
The `shouldExist` is enough to check if we should add or remove a row from the list of selected rows: 

https://github.com/tannerlinsley/react-table/blob/10a5a9226027149004ffed7cf3d3dd26b41afbac/src/plugin-hooks/useRowSelect.js#L125-L127

And for the case when the value is the same as before, we already return from the function: 

https://github.com/tannerlinsley/react-table/blob/10a5a9226027149004ffed7cf3d3dd26b41afbac/src/plugin-hooks/useRowSelect.js#L129-L131

I had an issue where for some reasons my table had `undefined` for the `isSelected` there, but couldn't reproduce it on CodeSandbox. This issue seemed to impact only these conditions, making selected rows not uncheckable, removing this unnecessary check fixes it for me.

If someone knows why the `isSelected` can be `undefined` in the first place — I couldn't find this out easily — maybe it would be worth investigating, but even then I think those checks are redundant.

### DIFF
--- a/.size-snapshot.json
+++ b/.size-snapshot.json
@@ -1,13 +1,13 @@
 {
   "dist/index.js": {
-    "bundled": 130976,
-    "minified": 61378,
-    "gzipped": 15745
+    "bundled": 130947,
+    "minified": 61370,
+    "gzipped": 15740
   },
   "dist/index.es.js": {
-    "bundled": 130034,
-    "minified": 60537,
-    "gzipped": 15571,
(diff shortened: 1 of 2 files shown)


tannerlinsley:
`isSelected` may be `undefined` if the user did not indicate a `true` or `false` state for the row selection and merely just want to invert or "toggle" the state (eg. `old => !old`)

PR #1920, opened 2020-02-18
useTable: Fixes memoizing of headerGroups

kizu:
When the headerGroups were defined at first, they were memoized and then attached to `getInstance()`, but later `getInstance().headerGroups` was redefined using a filter, removing the memoizing.

That PR fixes it by properly memoizing the new filtered `headerGroups`.

[Diff without whitespace changes](https://github.com/tannerlinsley/react-table/pull/1920/files?utf8=%E2%9C%93&diff=split&w=1)

### DIFF
--- a/.size-snapshot.json
+++ b/.size-snapshot.json
@@ -1,13 +1,13 @@
 {
   "dist/index.js": {
-    "bundled": 130976,
-    "minified": 61378,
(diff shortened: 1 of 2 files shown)


tannerlinsley:
Nice find :)

Issue #1901, opened 2020-02-11
I think toggleAllRowsSelected is doing the wrong thing.


boatcoder:
# Using v7
## Selecting/Deselecting All that are filtered should not affect non-filtered rows
Currently the toggleAllRowsSelected always clears the selection and creates a new one.  That shouldn't happen, it should add to or remove from the selection, not completely reset it and restart with an empty selection.

## To Reproduce
Using this sandbox:  https://codesandbox.io/s/table-with-checkbox-rb9mk
* Sort by something and then select a few rows
* Filter by something that causes those rows to disappear and a small subset to display
* Use the header to select all
* Remove the filter and the first 2 rows are unchecked (may need to re-sort to see them, but that is another issue with the example)

## Expected behavior
I can add to and remove from the global selection while a filter is in place without affecting things that are not filtered

**Codesandbox!**
https://codesandbox.io/s/table-with-checkbox-rb9mk

React-Table Version 7.0.0-rc.15

Browser versions don't matter

*Additional context**
Here is a better reducer for useRowSelect and how I had to patch it in my code because replacing the reducer is almost impossible without this kind of hack:
```
  const useFixedRowSelect = (hooks) => {
    useRowSelect(hooks);
    hooks.stateReducers.pop();  // Remove the damaged reducer

    function reducer(state, action, previousState, instance) {
      if (action.type === actions.init) {
        return {
          selectedRowIds: {},
          ...state,
        };
      }

      if (action.type === actions.resetSelectedRows) {
        return {
          ...state,
          selectedRowIds: instance.initialState.selectedRowIds || {},
        };
      }

      if (action.type === actions.toggleAllRowsSelected) {
        const { selected, } = action;
        const { isAllRowsSelected, flatRowsById, } = instance;

        const selectAll =
      typeof selected !== 'undefined' ? selected : !isAllRowsSelected;

        // Only remove/add the rows that are visible on the screen
        //  Leave all the other rows that are selected alone.
        const selectedRowIds = Object.assign({}, state.selectedRowIds);

        if (selectAll) {
          Object.keys(flatRowsById).forEach((rowId) => {
            selectedRowIds[rowId] = true;
          });

        } else {
          Object.keys(flatRowsById).forEach((rowId) => {
            delete selectedRowIds[rowId];
          });
        }

        return {
          ...state,
          selectedRowIds,
        };
      }

      if (action.type === actions.toggleRowSelected) {
        const { id, selected, } = action;
[...]

PR #1925, opened 2020-02-19
Fixes the RowSelect Reducer #1901

boatcoder:
Built a whole new test case that combines `useRowSelect` and `useFilters` then went thru all the scenarios of selection with 

1.  filtering
2.  selecting all . (singular row select is handled in useRowSelect.test)
3.  deselecting all
4.  changing the filter
5.  selecting all
6.  removing the filter
7.  selecting all
8. deselecting all

`yarn format` produces changes outside of my code, so I did not commit the examples or some of the hooks that prettier changed, maybe that should be a separate PR....


tannerlinsley:
Wow, this looks great so far! Can you please remove the snapshot file? I am pretty sure that's an artifact after I moved away from snapshot tests.

PR #1928, opened 2020-02-19
Update Documentation to reflect behavior

pmorgan3:
The accessor property is only necessary when using the useSortBy hook, correct?


pmorgan3:
I was mistaken

PR #1932, opened 2020-02-20
WIP: Enhancement in adding additional classname on resizing

### LINKED ISSUE #1916: Resize: Finding or Adding Class to current Resizing column
I have been using react-table just few days ago only and been stuck with this queries

Is there a way to know that current `column` in a table is being resized currently ?
Do we have any `props` to add a `className` to the column which is resized currently ?

The reason behind this is to highlight the column header which is getting resized.

Version: `6.11.5`

RaghulXander:
**PR Summary:**
- Added a class `rt-resizing` to the header which is getting resized.
- Solves the problem of adding additional styles ⚡️ to resizing header column in ease.
- Added support only to v6 react-table.

Happy to take any suggestion or changes 👍 

resolves #1916


tannerlinsley:
There are too many files being committed with this PR. I'll gladly take a look at this, but first, you'll need to make sure you are only committing necessary changes to the minimum amount of files. If you need to, check out a fresh version of v6 before applying your changes, then open a new PR.

Issue #1900, opened 2020-02-11
v7 Row selection not working


evgeniylovejs:
Using v7

**Describe the bug**
Found bug in my project then checks on your kitchen sink example and there it is also reproduced.
When user apply some filters or globalFilter there is no way to select (or in some cases unselect) rows.

**Steps To Reproduce**

1. Apply any column filter
2. Press several times on row selection checkbox
3. Checkbox are locking

**Codesandbox!**
https://codesandbox.io/s/github/tannerlinsley/react-table/tree/master/examples/kitchen-sink

- OS: Windows
- Browser Chrome
- Version 7.0.0-rc.15

boatcoder:
Checkbox are locking?  What does that mean?

Checkboxes aren't updating?  It is because you aren't triggering a re-render.  Click a couple of checkboxes and page down and back up.  When you do that, they are selected and it displays.  Need to make sure you trigger a re-render for them to show up.

evgeniylovejs:
> Checkbox are locking? What does that mean?
> 
> Checkboxes aren't updating? It is because you aren't triggering a re-render. Click a couple of checkboxes and page down and back up. When you do that, they are selected and it displays. Need to make sure you trigger a re-render for them to show up.
> 
> Check this sandbox . https://codesandbox.io/s/table-with-checkbox-rb9mk

In yours example:
1. Put "legs" string in Alias column filter.
2. Select any row
3. Try to unselect this row

Actual result - unselect not working.

boatcoder:
Interesting.  I see this happening in the sandbox.  I don't see it happening in my storybook though.

boatcoder:
I'm seeing `isSelected` always set to false in the useRowSelect reducer
```
      if (action.type === actions.toggleRowSelected) {
        const { id, selected, } = action;
        const { flatGroupedRowsById, } = instance;

        // Join the ids of deep rows
        // to make a key, then manage all of the keys
        // in a flat object
        const row = flatGroupedRowsById[id];
        const isSelected = row.isSelected;
        const shouldExist = typeof selected !== 'undefined' ? selected : !isSelected;

        if (isSelected === shouldExist) {
          return state;
        }
```

row.isSelected never seems to be true which causes this failure.  In the debugger, I see the UI uncheck and recheck after the reducer.

tannerlinsley:
Yep, looks like an inconsistency between the useFilters plugin and the useRowSelect plugin. Will dig in.

PR #1937, opened 2020-02-23
Row deselection fix

vinodsai-a:
On Filter, deselecting a row is not working. fixed that


Jaraujo6:
@tannerlinsley I'm running into the issue #1900 as well from now being able to deselect a row after a filter is applied. This simple line change is all that needed to correct the issue. Is it possible to approve this PR??

Issue #1923, opened 2020-02-18
Row DnD CodeSandbox Example not working:  Error: 


alexanderela:
**Describe the bug**
The React Table Row DnD example on CodeSandbox is not working.  I get console errors saying, "Cannot read property 'TARGET' of undefined."  The error appears to be on line 87, in the Row . component

**To Reproduce**

1. Go to 'https://codesandbox.io/s/github/tannerlinsley/react-table/tree/master/examples/row-dnd'
2. See error

Thanks in advance and thanks for the excellent React Table!

**Screenshots**
![Screen Shot 2020-02-18 at 9 44 43 AM](https://user-images.githubusercontent.com/30238776/74758885-312d0600-5235-11ea-942c-3e366e1d4845.png)
![Screen Shot 2020-02-18 at 9 45 16 AM](https://user-images.githubusercontent.com/30238776/74758899-34c08d00-5235-11ea-8064-9cf7c540ce55.png)

**Desktop (please complete the following information):**

- OS: MacOS Mojave 10.14.16
- Browser Chrome
- Version Updating Google Chrome
Version 79.0.3945.130 (Official Build) (64-bit)

alexanderela:
Found a fix.  I had to update the version of `react-dnd` in the package.json in the https://codesandbox.io/s/github/tannerlinsley/react-table/tree/master/examples/row-dnd example.

Replace the existing version for react-dnd in this package.json with the current version of react-dnd (in my case it was `"^10.0.2"`) and it should work fine.

tannerlinsley:
Would you submit a PR that applies this fix?

ironicbrew:
> Would you submit a PR that applies this fix?

Put together a pull for this https://github.com/tannerlinsley/react-table/pull/1939

alexsandr1992:
Now codesanbox example has access denied.
I see message "Request failed with status code 403"

ironicbrew:
> Now codesanbox example has access denied.
> I see message "Request failed with status code 403"

It's important to note that the url in the examples is hardcoded so If you're testing the pull by looking at the readme in my fork its going to send you to to [tannerlinsley's master branch](https://codesandbox.io/s/github/tannerlinsley/react-table/tree/master/examples/row-dnd)

To test my fix checking out [my pull on codesandbox](https://codesandbox.io/s/github/ironicbrew/react-table/tree/issue/1923/examples/row-dnd) will get you to the fixed branch

[...]

PR #1939, opened 2020-02-24
Updated react-dnd package

ironicbrew:
To Address #1923 

+ Prevents code sandbox from crashing the dnd example


michaelshiel:
The link in the readme for this example is wrong too, it has "dhd" instead of "dnd"

PR #1962, opened 2020-03-06
[WIP] useColumnPin

gargroh:
Added new react table hook for Column Pinning/Freezing. It supports pinning to right and left of table. 

Functionality illustration - column-pinning


tonilxm:
Really love this feature... I hope it can make it on next release ❤️ 😍

tannerlinsley:
Big question here is whether or not this is supported with grouped headers and how compatible is it with other plugins?

gargroh:
Yeah, it supports grouped headers well, but i doubt about its interaction
with plugins such as useAbsoluteLayout...

luis-carlos-campos:
This plugin looks really good...
@gargroh Is it working properly on Safari?
@tannerlinsley Would you recommend anything for freezing columns on current version?

gargroh:
I am planning to have different layout option also, like three different tables so as to have different horizontal scrollbars for each segment(left, center, right) this may require some more time, lets see how it goes.

sherbakov1:
@gargroh Great and useful feature! When are we going to see it live in production?

tannerlinsley:
This feature is awesome, though in v8, it will be implemented a bit differently. I used the general concepts here to build the new version (that ended up entirely different in v8's improved plugin system). I'm going to close this PR since there won't be any forward compatibility when v8 goes live.

luis-carlos-campos:
@tannerlinsley Are you tracking this feature somewhere else for V8 ?

tannerlinsley:
No, but all it would take is this PR to be reopened, pointed at the next branch and refactored to use the new plugin system.

tannerlinsley:
Sorry, got confused. This plugin is already implemented on the next branch. You can check it out here: https://github.com/tannerlinsley/react-table/blob/next/src/plugins/withColumnPinning.js

luis-carlos-campos:
That's awesome.
Thanks @tannerlinsley

amd-9:
> Sorry, got confused. This plugin is already implemented on the next branch. You can check it out here: https://github.com/tannerlinsley/react-table/blob/next/src/plugins/withColumnPinning.js

Seems that link not working anymore :( Sorry for bothering @tannerlinsley but is there any already implemented solution for column pining?

PR #2008, opened 2020-03-17
[New Plugin] useColumnSummary hook 

gargroh:
Added new hook, helps in showing column summary.

Something like - column-summary

76535091-56add800-64a0-11ea-89b6-ba6941440e5d 1


gargroh:
@tannerlinsley wanted to know your thoughts on it, then i can go ahead with documentation/examples.

src/aggregations.js
```diff
@@ -1,7 +1,7 @@
 export function sum(values, aggregatedValues) {
   // It's faster to just add the aggregations together instead of
   // process leaf nodes individually
-  return aggregatedValues.reduce(
+  return (aggregatedValues || values).reduce(
```
tannerlinsley: What's going on here?

tannerlinsley:
I'm confused why this is managing state and allowing to update state. Shouldn't a summary be something that is derived and calculated, not set?

gargroh:
@tannerlinsley, summary type(such as count or sum https://github.com/tannerlinsley/react-table/pull/2008/files/05725d689e46fa88a92c747d220f029de494eba4#diff-ee4a9fb75eba461aefc6a9e117f3d36aR37 ) is set in table state and its derived value is attached as column property(https://github.com/tannerlinsley/react-table/pull/2008/files/05725d689e46fa88a92c747d220f029de494eba4#diff-ee4a9fb75eba461aefc6a9e117f3d36aR97)

tannerlinsley:
I think it would be best to look into porting this to the upcoming v8. The plugin system is a bit different, but it will give you a good chance to get to know the new API and help me with some feedback.

PR #2013, opened 2020-03-17
New documentation site built with Docz

riceboyler:
So, this uses Docz (http://docz.site) to build a documentation web site for react-table. For the most part, it just uses the already built Markdown files in the repo, but I added a new Quick Start page, as well as splitting out the Examples into their own pages (Simple | Complex | Controlled | UI/Rendered). It's not perfect yet, some of the links will have to be updated to change from pointing to .md files to relative paths. But, it's probably 90% of the way there. If you like this, I'll go forward with making more updates and further enhancing the documentation.

Thanks!


riceboyler:
Crap, I just realized I forgot to add the .docz/.cache folder to the .gitignore. Let me add that so this isn't so bloated.

tannerlinsley:
So what would be the plan here for the existing documentation? Would it continue to work? If so, then we would probably want to change up the main README so that it only includes things like the overview, sponsors and a link to the docs. Then in the docs main readme, we can have mostly the same stuff, but it full format. I think for now I will just create a subdomain at react-table.tannerlinsley.com to host this on Netlify, then we can add the link at the top of the repo and in the readme to point to that.

Thoughts?

riceboyler:
My $.02: I think it would make sense to leave the main README.md in place, because duh, but I think it would point to the documentation site for the bulk of the documentation pieces. I think it makes sense to leave the code of conduct, changelog, contributing in Github as well, just so they're right there in the repo, but the actual "here's how you use this amazing thing" documentation could be just on the docz site and the main readme could point to it.

A question: I embedded the CodeSandboxes for some of the examples, but for the Simple page, it was just too many iframes and it slowed the page to an absolute crawl (as you might imagine). Do you like having the CodeSandbox embedded, or was that a step too far?

tannerlinsley:
Yeah, the existing readme is fine, as long as the links in it point to the documentation site where necessary and doesn't create duplicated docs anywhere.

riceboyler:
OK, I'll try to make those updates today, but I'm fighting a stomach bug, so it may be delayed...

riceboyler:
@tannerlinsley I was thinking about this over the weekend: What if you were to create a new repo that could house the documentation for each of your libraries? Do you think that would make more sense, or would it make more sense to just leave the documentation in each project and keep everything independent?

BTW, updated the links to point internally on this. If you want Docz advice on deploying to Netlify: https://www.docz.site/docs/deploying-your-docs#using-netlify

tannerlinsley:
From experience, it's **much** easier to keep the docs as close to the code as possible.

.docz/gatsby-config.js
```diff
@@ -0,0 +1,112 @@
...
+        'edit-branch': 'master',
+        config: '',
+        title: 'React Table Documentation',
+        description: 'Documentation for the react-table library built in Docz',
+        host: 'localhost',
+        port: 3002,
+        p: 3000,
+        separator: '-',
+        paths: {
+          root: '/Users/jasonclark/Source/react-table',
```
tannerlinsley: Why exactly are this vendored to your computer?

.docz/gatsby-node.js
```diff
@@ -0,0 +1,144 @@
+const get = require('lodash/get')
```
tannerlinsley: Just out of curiousity, what is this file doing exactly?

.docz/gatsby-config.js
riceboyler: That is an outstanding question. I think that file is autogenerated (as is everything in the .docz folder).

.docz/gatsby-node.js
riceboyler: Yeah, everything in `.docz` is autogenerated by Docz. So, I'm not 100% sure.

riceboyler:
@tannerlinsley If it weren't for needing to deploy to Netlify, I would say the `.docz` folder should be added to .gitignore. After all, it will be regenerated each time.

.docz/gatsby-config.js
tannerlinsley: Interesting! Well, let's hope it doesn't cause any issues :)

.docz/gatsby-node.js
tannerlinsley: I see. I thought this was some regular gatsby thing. So that's good.

tannerlinsley:
I would prefer that it is excluded for sure. But why does it need to be there to deploy to netlify?

riceboyler:
D'oh. Colossal brain fart. Netlify will just run the docz script, so it doesn't need to be included. 

I'll add it to `.gitignore` and remove it from git right now.

tannerlinsley:
Awesome. Then we just need to specify the right build script for netlify.

riceboyler:
OK, that's much cleaner. According to this: https://www.docz.site/docs/deploying-your-docs#using-netlify we should just need to point it to `yarn docz:build` and specify the proper folder.

tannerlinsley:
Good golly, gatsby is slow...

riceboyler:
> Good golly, gatsby is slow...

For the initial deploy or just in general? In my experience, once it's built, Gatsby is stupid fast. Getting it built does require a bit of patience.

tannerlinsley:
Yeah, the build process on Netlify is super chunky.

tannerlinsley:
Also, I wish this wasn't there on the github root: 
Screen Shot 2020-03-25 at 2 43 04 PM

riceboyler:
Yeah, I saw that. Headmatter...

After my current fight with wavesurfer.js has come to either me fixing what I need or committing harikari, I'll take a look at creating a GettingStarted file that Docz will use, and ignore the main Readme.md so we can remove the headmatter.

tannerlinsley:
Well, I'm open to any ideas. Also, can we make the font-size a bit smaller?

riceboyler:
Sure, we can customize the Gatsby Theme to our own liking. 

But don't you know that (according to marketing people) everybody using the internet is totally blind nowadays? How could anybody possibly be expected to read text smaller than 16px??? </sarcasm>

tannerlinsley:
Lol. Yeah, you may be right, but for developers, I'd like to match their expectations with sites like Github etc and do 14 px. Also, not all fonts are created equal. This could also be adding to the large-ness of it.

PR #2021, opened 2020-03-18
Upgrading rollup to fix bug with React external. Resolves systemjs/systemjs#2147

joeldenning:
See related systemjs/systemjs#2147 and https://github.com/rollup/rollup/pull/3420. There was a bug in rollup where it assumed incorrectly that external ES modules have the full Object.prototype available on them.

Here's what this changes inside of react-table.development.js:

Before:
image

After:
image

Note that calling `hasOwnProperty()` requires the object to have `Object.prototype` as its prototype, which is not true for ES modules.


csantos1113:
@tannerlinsley is it possible to get this merged soon 🙏? Thanks

PR #2073, opened 2020-03-31
Add cell’s value to its renderer back

kizu:
One thing that I wondered about and wasn't sure why it was changed from v6 to v7 is the Cell renderer API, where the value was only accessible inside the `cell`. From how I look at it, nothing stops us from returning it back, as there shouldn't be any collisions with the instance's properties and methods, and it would make both migrating from v6 much easier, and writing new Cell code much simpler as well.

We can still keep `cell` for stuff like in the `useRowState` and not to break anything for those who already use the v7's API.

If there was a reason to remove the `value` from the Cell's props, I'd be happy to talk about this in order to find a better solution for the underlying problem.


tannerlinsley:
image

PR #2117, opened 2020-04-07
Documentation links are just wrong.

boatcoder:
Fixed the links


gregolai:
Thanks!

tannerlinsley:
Does this work for both github browsing and react-table.js.org?

PR #2166, opened 2020-04-14
[WIP] V8

tannerlinsley:


gargroh:
Hi Tanner, thanks for the awesome library, its plugin and play architecture really helps in building customization on top of it, this is super helpful. 

Regarding this change, react-table was using `useState` pattern earlier and is migrated to useReducer pattern with https://github.com/tannerlinsley/react-table/blob/master/CHANGELOG.md#700-beta20 release. 

Both have their own pro and cons, going for the cons...it is easy to set state using `useState` pattern more in imperative style. And with `useReducer` pattern it is easy to define different actions and is useful in collecting feature usage stats/analytics. 

Apart from 3KB benefit and easy to read lib(which seems to be driven by https://github.com/tannerlinsley/react-table/discussions/2147 discussion), I wanted to check can we get best of both world i.e. what if with `useReducer` pattern only we expose a method/plugin/hook which helps us achieving the functionality of newly added `useState`.

tannerlinsley:
The main benefits actually haven't been shown yet as this is still a WIP. But for starters:

- Much less code, at least for our case. If you look at the migration, the logic inside of `useState` is exactly the same as useReducer in most cases, and also, since it lives inside of the context which it is in, the access to the instance (or `getInstance` makes more sense here.
- It's very probable that the concept of updating table state is about to change internally as we externalize some or most of the table's core to a non-react instance factory.
- Even though it's setState, it's a special flavor, one that is able to pass action types, action meta, etc, so aside from the syntax, the benefits at the end of the day are the same as a reducer, except it's more ad-hoc, which is perfectly suited for plugins and users IMO.
- Yes, we could expose an action that replicates `useState` in the reducer, but if you noticed, for every action we had, we were also exporting a companion proxy method on the instance that does the same thing and most times is way more reliable than calling the action itself. Getting rid of that layer makes things way more direct
- Again, this is all subject to change. This is the stage we're in...

image

tannerlinsley:
Oh, and also, this branch may not even be viable if my findings aren't worth it. So I wouldn't get married to any ideas here at all.

tannerlinsley:
This PR is likely v8. It's been refactored to something that is much easier to understand and maintain. More notable differences:

- What were "core" plugins are now native  to the `useTable` hook
- All of the new core functionality is now more tightly integrated, way more predictable. All core features are "always on", but not always active, making them easier to test and type together
- A new plugin-system style is being born, but is not yet complete. It will likely involve simply extending and composing the options object that is passed to the `useTable` hook
- The table instance has a much better and fully-formed api that should allow full control over most features of the table, including column, row, and even cell-level operations
- State is now much easier to control from fully-automatic, to partial opt-in, all the way up to fully controlled. Some things are still not controllable, but the system would be much easier to extend to make that happen now.

tannerlinsley:
- Better support for building dedicated expander and row selection columns
- Much of the API has been rearchitected to be lazy (but still memoized), meaning that basically everywhere, you'll see a lot of getter functions. This actually ends up improving render performance quite a bit, since we can set the getter functions really early on when we build the row/cell model even though the final state those getters will return isn't ready yet. That also means we don't have to re-loop over the rows and cells later on (which is a ton of extra wasted looping) to do it then.

tannerlinsley:
The kitchen-sink example is the one I've been using to develop this with. So use that if you want to try it out!

tannerlinsley:
Tracking in #2335 now.

PR #2229, opened 2020-04-25
Enhancement: Added Classname to Table Header Currently Getting Resized

### LINKED ISSUE #1916: Resize: Finding or Adding Class to current Resizing column

RaghulXander:
**PR Summary:**

- Added a class `rt-active-resizing-header` to the header which is getting resized.
- Solves the problem of adding additional styles ⚡️ to resizing header column in ease.
- Added support only to v6 react-table.
- Happy to take any suggestion or changes 👍

resolves #1916


RaghulXander:
@tannerlinsley i don't know whats wrong with deploy check the error states no `doc:build` found kind of.

tannerlinsley:
Sorry for the delay, lots of work to do! I'm not currently accepting PRs for v6. I would suggest forking v6 to make your change.

PR #2281, opened 2020-05-07
Fix RMWC links in examples documentation

marvinthepa:


marvinthepa:
Also fix comcastbusiness.com (which seems offline) to business.comcast.com to fix the build.
Hopefully uber.com works this time 😄 .

marvinthepa:
@tannerlinsley seems like uber is blocking the requests from github. Would it make sense to ignore them for the url checker?

tannerlinsley:
Yeah let’s ignore them

gargroh:
We can also include mdx files for watching broken urls in `Broken Link Checker` to catch broken links like this

.github/workflows/links-test.yml
```diff
@@ -20,7 +20,7 @@ jobs:
           subfolder: /github/workspace/
 
           # A comma-separated list of file types to cover in the URL checks
-          file_types: .md
+          file_types: .md, .mdx
```
gargroh, line 23: I think we need to remove `space` after `,` so as to make that valid.

marvinthepa: It did not complain 😃 .

Anyways, It is easier to change than to test if it works both ways.

gargroh: Thanks, actually i find in logs for '.mdx' files, it was not showing earlier, now it shows.

Thank you

PR #2282, opened 2020-05-07
complete docs for overriding Cell renderer

### LINKED ISSUE #2269: Documentation missing for overriding the Cell renderer

https://github.com/tannerlinsley/react-table/blob/master/docs/api/useTable.md#cell-properties

```
- `value: any`
  - The **resolved** value for this cell.
  - By default, this value is displayed on the table via the default `Cell` renderer. To override the way a cell displays
```

That sentence stops right in the middle, without explaining how to override the way the cell displays.
Opening as a bug instead of a feature request as it looks like the information was intended to be there.

marvinthepa:
Fixes #2269.

### DIFF
--- a/docs/api/useTable.md
+++ b/docs/api/useTable.md
@@ -95,7 +95,7 @@ The following options are supported on any column object you can pass to `column
   - Receives the table instance and column model as props
   - Must either be a **string or return valid JSX**
   - If a function/component is passed, it will be used for formatting the header value, eg. You can use a `Header` function to dynamically format the header using any table or column state.
-- `Cell: Function | React.Component => JSX`
+- `Cell: Function | React.Component => JSX` <a id="column-object-cell"></a>
   - Optional
   - Defaults to `({ value }) => String(value)`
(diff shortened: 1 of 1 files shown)


docs/api/useTable.md
```diff
@@ -95,7 +95,7 @@ The following options are supported on any column object you can pass to `column
   - Receives the table instance and column model as props
   - Must either be a **string or return valid JSX**
   - If a function/component is passed, it will be used for formatting the header value, eg. You can use a `Header` function to dynamically format the header using any table or column state.
-- `Cell: Function | React.Component => JSX`
+- `Cell: Function | React.Component => JSX` <a id="column-object-cell"></a>
```
tannerlinsley, line 98: What is this for?

marvinthepa: This produces the anchor that is referenced by `[Cell](#column-object-cell)`. I always find it helpful to have cross-references as clickable links..

If you know of a more markdown-esque syntax to produce this, I would be more than happy to know of it.

PR #2325, opened 2020-05-14
fix: detect browser support for passive events

### LINKED ISSUE #2302: useResizeColumns - IE11 issue
Initially opened as a discussion item, but further digging my end leads me to believe that this is an actual issue. `useResizeColumn`, specifically in IE does not work. It looks like mouse event listeners aren't being removed when a column is resized, causing the table to rerender based on mouse position. 

Unable to replicate in codesandbox since this is an IE11 issue, but I implemented the `useResizeColumns` [example ](https://codesandbox.io/s/github/tannerlinsley/react-table/tree/master/examples/column-resizing) provided in the docs.

react-table-v7-IE

1. Resize any column in IE11
2. See column snap to max width (if you have the console open, and are logging state from `useTable`, you should see the width of the resized column size as you mouse around)

**Expected behavior (recommended)**
Columns resize as they do in other browsers

**Desktop**
- OS: Windows 10
- Browser: Internet Explorer 
- Version: 11.778.18362.0

BrianMitchL:
This fixes #2302 and brings IE support to the useResizeColumns plugin. It adds a feature check function which performs a memoized check for using `{ passive: false }` in an event handler.

It seems that with IE 11, the event could be added with the unsupported option, but then the event could never be removed. This caused a move handler to be added on every mousedown on a resizer, and the rezising would never stop.

More info on the feature check:
- https://developer.mozilla.org/en-US/docs/Web/API/EventTarget/addEventListener#Safely_detecting_option_support
- https://developer.mozilla.org/en-US/docs/Web/API/EventTarget/addEventListener

### DIFF
--- a/src/plugin-hooks/useResizeColumns.js
+++ b/src/plugin-hooks/useResizeColumns.js
@@ -6,7 +6,7 @@ import {
   ensurePluginOrder,
 } from '../publicUtils'
 
-import { getFirstDefined } from '../utils'
+import { getFirstDefined, passiveEventSupported } from '../utils'
 
 // Default Column
 defaultColumn.canResize = true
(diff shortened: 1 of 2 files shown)


tannerlinsley:
Nice! Would you mind adding this to the Next branch too?

BrianMitchL:
Sure thing!

PR #2327, opened 2020-05-14
[Next] fix: detect browser support for passive events

BrianMitchL:
Apply #2325 to the next version work.


tannerlinsley:
Huzzah!

Issue #2174, opened 2020-04-15
version 7.0.4, filter: 'includes', not working as before


schoi8uber:
**Describe the bug (required)**
In version v7.0.4, I noticed that the select filter using below not working any more.
`filter: 'includes' `

This works fine in the previous version I had v7.0.0. I think it seems to perform AND operator instead of OR when there are multiple values selected from the dropdown.

**Provide an example via Codesandbox! (required)**
Use a new [react-table codesandbox](https://codesandbox.io/s/m5lxzzpz69) to reproduce the issue.

**Steps To Reproduce (required)**

**Expected behavior (Recommended)**
Select filter using filter: 'includes', should perform OR operator to filter out all rows including any of selected values.

**Screenshots**

**Desktop (please complete the following information):**

- OS: [e.g. iOS]: MacOS 10.15.3
- Browser [e.g. chrome, safari]: Chrome 
- Version [e.g. 22]: Version 81.0.4044.92 (Official Build) (64-bit)

**Smartphone (please complete the following information):**

**Additional context**

ericberens:
Related: https://github.com/tannerlinsley/react-table/pull/2001

PR #2329, opened 2020-05-15
Feature/Fix: Add filterTypes: includesSome & includesValue

nderscore:
This is sort of a fix for #2174.

In a recent release, #2001 changed the behavior of the `includes` filter to check for a single filter value within an array or string row value. This broke some existing usages of the previous `includes` filter, which expected the previous behavior (checking for a single row value within an array or string filter value).

`includes` is a very vague name/name-prefix to begin with (especially without any documentation yet around built-in filterTypes), but I expect that the new behavior is probably a more common use-case, especially scenarios where the row value is a string. From what I've seen, most of the codesandbox examples expect this behavior too.

So, instead of reverting #2001, I think it makes more sense to add a new filter to reintroduce the previous `includes` behavior, which I've implemented here as `includesValue`.

While I was in there, I also added another filter `includesSome`, which is a natural complement to the existing `includesAll` filter.

### DIFF
--- a/src/filterTypes.js
+++ b/src/filterTypes.js
@@ -64,6 +64,32 @@ export const includesAll = (rows, ids, filterValue) => {
 
 includesAll.autoRemove = val => !val || !val.length
 
+export const includesSome = (rows, ids, filterValue) => {
+  return rows.filter(row => {
+    return ids.some(id => {
+      const rowValue = row.values[id]
+      return (
+        rowValue &&
+        rowValue.length &&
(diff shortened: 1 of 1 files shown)


tannerlinsley:
Very nice :)

nderscore:
Thanks @tannerlinsley ! 👍

ashvin777:
If you release the documentation around it then it will super great :)

as-aschnabel:
Bump ^^

PR #2335, opened 2020-05-15
[WIP] V8

tannerlinsley:


GuptaSiddhant:
Will v8 package contain typescript types as they are present in react-virtual?

tannerlinsley:
Possibly. Then again, it’s possible I migrate it to Typescript too. No promises.

pedroeusebio:
Hi @tannerlinsley, do you know when this new version will be released? I saw some really good features (e.g. columnPinning) that I would like to use!
Please let me know if there is something that I could do to help

tannerlinsley:
It's going to be a few months at least. If you would like to help, I would suggest checking out the `next` branch and trying to get the basic example to work. That will help with catching any bugs, writing docs, writing examples, converting existing examples, etc.

GuptaSiddhant:
Now that in latest commit all files are using typescript, @tannerlinsley, are you adding types yourself? Or are you using @types/react-table as starting point? 
Since TS is throwing all kind of errors, and library cannot be compiled, the testing of examples had to be halted.

tannerlinsley:
I’m in the process of converting everything to TS. I will use a few of the existing types, but a majority will be home grown for the new plugin system.

OfficerHalf:
Yikes, just when I was about to attempt to migrate our codebase to v7 since it seems like the API has stabilized, I see this.

I'm definitely all for Typescript (it's been one the blockers for us migrating to v7, since it took a while for types to materialize), but this seems like a pretty quick turnaround for major version. The answer may be "I don't know" but are you expecting a smaller amount of breaking changes/easier migration between 7 and 8? Or is this another total re-write situation?

If there's a better place for discussion I'm happy to put this there, just didn't see anything.

tannerlinsley:
@OfficerHalf This is definitely nowhere near the changes found in v7. The external API in v8 will be almost identical (with breaking changes for only a couple of apis, mostly around controlled state). Other breaking changes will be focused around the plugin system as well, but only in boilerplate structure and types.

I'm currently trying has hard as I can to get the next version to be written in TS, but trust me, React Table is no small undertaking to type.

Regardless of what happens, it's going to be a smooth transition from v7 to v8. You are fine to keep going with the v7 migration, a vast majority of your work will transfer to v8.

Also, v8 is a long way out. It could even take a year or longer (v7 did).

OfficerHalf:
Okay, that's reassuring. I'll continue with the migration work then. 👍

goatfryed:
Hey @tannerlinsley,
I've used and followed your project for quite some time and like to offer my help as well.
I checked out the next-typescript version two weeks, fixed most tests and played around with the new plugin system.

I decided to give it a try and start sprinkling a bit of typescript. I noticed, that you made a new commit on friday that starts using typescript. Even though my work doesn't  build on it, since we made it in parallel, i hope the effort can still support your project and I'd like to make a pull request about it, at least as food for thought.

If you're open for it, I'd be glad to help translate the existing js to ts in an interim setup that supports both so you can focus more on adding and changing meaningful parts of the library.

Two thoughts I had looking at it:
Since most tests were broken due to the removal (?) of prepareRow and some issues with render('Cell'), I extracted a very small plugin that adds the render support as an optional plugin to play around with the new plugin system as a POC. It's not really dependent, but it might actually be nice to reduce the core to the bare minimum...
Maybe it would be nice to provide react-table as the bare core plus one or more presets of plugins. Other candidates that could be optional as well might be get[..]Props

From my previous use cases, I used other react hooks occasionally in custom plugins. In general, to my understanding, the plugins would be something that shouldn't change between calls to useTable, right? I often use currying to generale a local useTable with the plugins i need and then pass the options. Based on that, I think it's actually nicer to have the plugins as a separate method argument instead as part of the options object that needs to be merged. Maybe it's cleaner to keep it similar to the old api. Other than that, the new api is really nice so far!

tannerlinsley:
Yeah, TBH, it's all still up in the air. I've got a few other branches that are unpublished too that are trying new things. I hope to wrap up my R&D by the end of the year and start actual work on v8 in 2021.

goatfryed:
Ah, okay. I understand. If at one point you'd like help with migrations,
please let me know. I'd be glad to give something back to the project.

tannerlinsley:
Absolutely. I'll keep that in mind :)

tannerlinsley:
Moving this to the "beta" branch.

xiaoyu-tamu:
hi, @tannerlinsley.  Where can i see the beta branch?

tannerlinsley:
The V8 branch is currently private and has undergone some massive changes to migrate to typescript. When it’s ready, it will be available in alpha and beta to sponsors first, hopefully starting this spring, with an eventual rollout to general public and OSS over the rest of the year.

PR #2338, opened 2020-05-16
Basic Example: v8 try

gargroh:
Tanner i was trying to bring up v8 version with basic example, getting table as in screenshot

image

Is there anything i am missing regarding preparation of row data or something in context of rendering header/cell ?

### DIFF
--- a/examples/basic/src/App.js
+++ b/examples/basic/src/App.js
@@ -40,7 +40,7 @@ function Table({ columns, data }) {
     getTableBodyProps,
     headerGroups,
     rows,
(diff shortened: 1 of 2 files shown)


tannerlinsley:
Does the kitchen sink example work that is already in here?

gargroh:
ohh...was not aware of that, Thanks will check

gargroh:
Are we talking of this example 
https://github.com/tannerlinsley/react-table/blob/next/examples/kitchen-sink/src/App.js#L6 ?

PR #2362, opened 2020-05-24
docs: Document support for pageCount: -1

TimNZ:


TimNZ:
I didn't add any links, failing on existing

PR #2364, opened 2020-05-24
Fix incorrect documentation for column header accessor

samccone:
Currently as of react-table 7.x the accessor field of the columns used without id errors (based on the typings of @types/react-table)
In order for this to not error by default switch to `id` instead of `accessor`. 

https://github.com/DefinitelyTyped/DefinitelyTyped/blob/ac0327a3ddfe6dac85bf13b99363e58925ee56f2/types/react-table/index.d.ts#L78

### DIFF
--- a/docs/quickstart.md
+++ b/docs/quickstart.md
@@ -35,11 +35,11 @@ const columns = React.useMemo(
   () => [
     {
       Header: 'Column 1',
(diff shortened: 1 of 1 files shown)


samccone:
Dug more, looks like the typing is incorrect. closing. sorry for the noise!

PR #2410, opened 2020-06-04
fix: documentation for default Cell function

timkindberg:


tannerlinsley:
Value is the actual default, not cell.value is it not?

timkindberg:
I looked in the src and yeah it is... it was just weird because when I supplied my own custom `Cell` I did not see a `value` on the top level obj param, I had to dive into `cell.value`. I'll close I guess.

PR #2411, opened 2020-06-04
Clarify how to update hiddenColumns

jlei523:
I was really confused on how to toggle hide and show columns. This was the missing documentation I needed.


thomasmarr:
hey @jlei523 could you clarify what you meant by 
> Changing `hiddenColumns` directly won't cause the table to add the hidden columns back.

I took this to imply that it's not possible to useControlledState to set the hiddenColumns value directly. I don't think that's correct, so just wanted to check what the intention was.

PR #2443, opened 2020-06-12
fix(useSortBy): fix the bug that causes flatRows to be incorrect when sorting and using subRows

### LINKED ISSUE #2303: Nested SubRows Not Sorting
**Describe the bug**
SubRows do not sort beyond a parent subRow of length 1.

**Provide an example via Codesandbox!**
https://codesandbox.io/s/nice-surf-hzprc

1. Go to the sandbox
2. Click on any of the info columns to apply sorting
3. Notice that no sorting occurs

**Expected behavior**
SubRows should sort at any depth.

**Desktop / Smartphone:**

- OS: Any
- Browser: Any
- Version: Any

**Additional context**
The bug is caused by an early termination of subRow sorting, which occurs on this line:
https://github.com/tannerlinsley/react-table/blob/master/src/plugin-hooks/useSortBy.js#L324

**Proposed Solution**
Adjust the check for skipping subRow sorting so that it only skips subRows if the length of subRows is zero (i.e. change: `!row.SubRows || row.subRows.length <= 1` to `!row.SubRows || row.subRows.length === 0`).

aaryn101:
This should also fix #2303, though obviously I was noticing different symptoms.


dbertella:
What's the status of this pr? I'm having this exact issue, I was going to report, can we move forward or is there any blocker?

dbertella:
If someone else lend here, as a work around I'm sorting the data before pass them as parameter to useTable hook, but yeah it's not great. Anything we can help on this one? (of course this will work as initial sorting, not sure what happens if you update sorting manually)

aaryn101:
@tannerlinsley Would you please take a look at this PR?

Chase-Dickerson:
@aaryn101, can you take another look at #2303? This PR did not address that issue.

PR #2458, opened 2020-06-17
Allow user to specify default Header

tariqporter:
[Issue 2457](https://github.com/tannerlinsley/react-table/issues/2457)

The `Header` option in `defaultColumn` is ignored as `column` already has `Header` specified, meaning it will overwrite anything further down the chain

```
  Object.assign(column, {
    // Make sure there is a fallback header, just in case
    Header: () => <>&nbsp;</>,
    Footer: () => <>&nbsp;</>,
    ...defaultColumn,
    ...userDefaultColumn,
    ...column,
  })
```

### DIFF
--- a/src/utils.js
+++ b/src/utils.js
@@ -75,7 +75,6 @@ export function decorateColumn(column, userDefaultColumn) {
     Footer: () => <>&nbsp;</>,
     ...defaultColumn,
     ...userDefaultColumn,
-    ...column,
   })
   return column
 }
--- a/src/utils.js
+++ b/src/utils.js
@@ -75,7 +75,13 @@ export function decorateColumn(column, userDefaultColumn) {
     Footer: () => <>&nbsp;</>,
     ...defaultColumn,
     ...userDefaultColumn,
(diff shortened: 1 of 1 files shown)


tariqporter:
I see why the test is failing....

`column` already has property `Header` on it, but not `Cell`, so when doing

```
  Object.assign(column, {
    // Make sure there is a fallback header, just in case
    Header: () => <>&nbsp;</>,
    Footer: () => <>&nbsp;</>,
    ...defaultColumn,
    ...userDefaultColumn,
    ...column,
  })
```

Then the `Header` from `...column` is going to overwrite anything further down the chain

tariqporter:
This resolves it

```
  // column already has default Header property
  if (userDefaultColumn.Header) {
    column.Header = userDefaultColumn.Header;
  }
```

But for consistency, it would likely be better to have `Footer`, `Cell`, and `Header` all specified on the same object as defaults

tariqporter:
This was due to incorrect usage - Data should not be passed in as a key named `Header`

PR #2461, opened 2020-06-18
feat(useResizeColumns): add reset column resizing

gargroh:
Added functionality for resetting column resized width to the initial width


akraines:
Just a quick note:
1. The sandbox example doesn't work
2. The documentation needs updating
3. Also todo: Document the autoResetResize property - if you don't set it to false, then trying to restore the state using initial state doesn't work

PR #2464, opened 2020-06-18
fix(aggregations): update syntax in unique aggregation function

### LINKED ISSUE #2173: unique aggregation return [SetIterator] instead of array
Seems like the following `unique` premade aggregation not working well:
https://github.com/tannerlinsley/react-table/blob/master/src/aggregations.js#L70-L72

I have the following column:
```
{
    variety: 'x',
    Header: 'x',
    accessor: d => d.id,
    aggregate: 'unique',
    aggregateValue: 'unique',
    Aggregated: a => {
      const {
        cell: { value },
      } = a;
      console.log(value);
      return `${value}`;
    },
  },
```

when I check the log it prints [SetIterator] instead of the unique values as an array.

if I am doing the following it works as expected:

```
import { unique } from "react-table/src/aggregations";
...
{
    variety: 'x',
    Header: 'x',
    accessor: d => d.id,
    aggregate: unique,
    Aggregated: a => {
      const {
        cell: { value },
      } = a;
      console.log(value);
      return `${value}`;
    },
  },
```

https://codesandbox.io/s/laughing-pond-ieulg?file=/src/App.js

thomasmarr:
Update syntax in return line of unique aggregation function
Use Array.from() instead of the spread operator [...]
Motivation: prevent polyfilling to [].concat() to enable reference by string in column definitions

closes #2173

### DIFF
--- a/src/aggregations.js
+++ b/src/aggregations.js
@@ -68,7 +68,7 @@ export function median(values) {
 }
 
 export function unique(values) {
-  return [...new Set(values).values()]
+  return Array.from(new Set(values).values())
 }
 
 export function uniqueCount(values) {


tannerlinsley:
Noice!!!

PR #2508, opened 2020-07-06
fix: initalState sortBy fix, if the last sortBy column is clicked we …

sullyslegacy:
fix: initalState sortBy fix, if the last sortBy column is clicked we will replace the sort state, not toggle it


sullyslegacy:
This PR fixes the following issue

Go to https://codesandbox.io/s/thirsty-pine-gtzl1
Click on Cost (This will only toggle the sort)
Click on Cost (This will remove the sort by Cost)
Click on Cost (This will only sort by Cost, which is what's expected)

PR #2516, opened 2020-07-10
docs: fix material ui examples links

rvgpl:
The links for `material-ui` and `material-ui-kitchen-sink` in the docs are broken and throw a 404. 

The path has an uppercase `UI` term in the path. This PR changes it to lowercase so the links work as expected


omarmosid:
Hey! I think the kitchen sink url still throws a 404. 
image

PR #2522, opened 2020-07-10
Fix dead link in readme

pmorgan3:
The link to the API documentation went to a 404, so I switched it to where the API documentation is currently located.


tannerlinsley:
Could you link to the doc site instead?

pmorgan3:
@tannerlinsley updated

PR #2531, opened 2020-07-13
docs(useExpanded): sub-components-lazy example

BrianMitchL:
This adds an example for lazy sub-components. This lets you, for example, fetch data for a subRow once it's it expanded.

I'm not sure what's needed to get this created and synced to a CodeSandbox, let me know if I need to do anything else!


justin-tufte:
Testing this out for an alternative to subrows. It works great for most use cases I could find except for row selection. It seems the parent row instance gets cloned down to the child rows so one child checkbox selects all checkboxes in that "group" including the parent row. Still looking for a way around that via overriding props but havent found a solution yet

BrianMitchL:
This version essentially renders the subRows outside of the react-table state. It's using the same renderer and capitalizing on the expansion state, but then it's rendering a custom component, so hooking back into react-table features within a subRow, or handling it differently, would probably require a different solution. I'm hoping some of this can be re-worked and made easier with v8.

thegitfather:
Thanks @tannerlinsley @BrianMitchL for your work! I'm also trying to use some extra functionality (expand, filter) while loading async data in the sub-comp. Of course this isnt working probably because as you said in the comments you would have to hook back to use those table features for the sub rows. So my use case is: expandable rows get loaded from the beginning and if they dont have any more rows, I know I have to render the sub component where load the "leafs" or final rows via a fetch in a useEffect. Can somebody point me in to the direction how you would handle this all together? I have an modfified sub-components-lazy example here: https://codesandbox.io/s/sharp-voice-59jmf?file=/src/App.js 

Any hints how to approach this very much appreciated :)

BrianMitchL:
You should be able to pass filters or other table state to your `renderSubComponent` function and pass it as additional props to `SubRowAsync` in order to add that to your async function to load more data.

lianghx-319:
Any demo or example for v8?

PR #2534, opened 2020-07-14
Add useRowSpan plugin

liberza:
This PR adds a new plugin, useRowSpan. The plugin sets a rowSpan attribute on duplicate cells, and provides the spanned rows to the user via the cell object.

Not complete yet - needs test, example, & docs.


tannerlinsley:
I'd like to build this into the core in the `next` branch. You up for that?

liberza:
Sure! I believe the only thing I was still working through is getting it to
not crash on sort.

underfisk:
@liberza I'm curious if this plugins work fine with React-virtualized since validateDOM is always throwing errors due to the HoC being a div. 
Have you tested this with any kind of virtualization? I'm looking forward for row span but has to support virtualize

liberza:
Great question, that's actually a problem I haven't solved yet with
rowspan. Seems pretty tricky with react-window, but virtuoso may mostly
just work with this.

I had looked a bit into how Autodesk's table library does rowspan +
virtualization, but it appeared that they weren't doing "real" rowspan with
the html attribute.

underfisk:
@liberza Yup i have checked Autodek's too but i really prefer react-table API but this comes to a need having virtualization with row spans, now i can't use row span html attribute because my HoCs are divs and React naturally can't handle that nested DOM  validation.
I would really like to find a good solution to keep react-table but considering spanning

liberza:
> @liberza Yup i have checked Autodek's too but i really prefer react-table API but this comes to a need having virtualization with row spans, now i can't use row span html attribute because my HoCs are divs and React naturally can't handle that nested DOM validation.
> I would really like to find a good solution to keep react-table but considering spanning

Ah, yes for my use case I've relied on <tr>'s rowspan attribute do the heavy lifting with regard to table layout. At least that's how I've used it so far. However, you could use the cell.rowspan property created by this plugin to manually do the styling required for a rowspan on your HoC divs though, if that's what you're looking for.

underfisk:
@liberza I'm just looking for because virtualize does not work with natural td, in case you have thousands of rows like in my use case, somethings you need to do the heavy lifting of calculating what might be on the screen of not. Let's say you have a merge that has 400childs and the screen can only display 50, i will chunk it and not display everything otherwise the whole merge parent has to be rendered, and naturally, with HTML rowspan it's going to get rendered.
Btw what do you mean by this plugin row span, do you mean that even without td your plugin? The style in my use case is done via the virtualize combined with Cell measurer, it will be hard to add row spans because when you add row spans, the inner text cannot be measured anymore (by default)
Btw i checked your plugin and i think it would be better if you create a merged cells data structure in order to prevent being re-calculating each time it gets re-render, like caching the row spans for future re-renders

underfisk:
@tannerlinsley Any news on this?

tannerlinsley:
Again, I think it's going to be easier and better if we just slate this for v8. When things are ready, I'll ping here for the PR to be reopened, but pointed at the next branch.

gargroh:
@underfisk in case interested to make this plugin available for v7, feel free to PR it at https://github.com/gargroh/react-table-plugins, in this repo we are maintaining all non-core v7 plugins, till the time they are available in v8.

underfisk:
@gargroh Thanks for mentioning that package, unfortunately, i have decided to start creating my own plugin to solve my needs and so far its going pretty well which is a pity, i kinda liked how React table is organized because it is indeed scalable
I'll stay tune anyway to the further plugin updates

alesdrobysh:
@tannerlinsley  so is there a way to make rowspan with v8?

andregueva:
I'm interested to know if I can do rowspan with v8

PR #2541, opened 2020-07-16
Fix bug in sortData subRows

leyinetwork:
Fix sort subRows stop recursion when subRows length is one.
If row.subRows length is one, the subRows of row.subRows[0] will not sort in recursion

### DIFF
--- a/src/plugin-hooks/useSortBy.js
+++ b/src/plugin-hooks/useSortBy.js
@@ -321,7 +321,7 @@ function useInstance(instance) {
       // If there are sub-rows, sort them
       sortedData.forEach(row => {
         sortedFlatRows.push(row)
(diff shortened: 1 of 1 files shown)


tannerlinsley:
Why would you need to sort a single row?

leyinetwork:
Because I want to sort subRows data in recursion. If I have data[n].subRows[0].subRows[0].subRows[0] data set and data[0].subRows length is one, I can't sort the data in data[n].subRows[0] and subRows of the data in data[n].subRows[0]

You can find the case in https://codesandbox.io/embed/boring-sanne-d56h1?fontsize=14&hidenavigation=1&theme=dark

leyinetwork:
@tannerlinsley Can you take a look?

tannerlinsley:
Ah. I see. Good find.

Issue #2528, opened 2020-07-11
Toggle all row selected not working for subRow


pjshu:
**Describe the bug (required)**
Toggle all row selected not working for subRow

**Provide an example via Codesandbox! (required)**
codesandbox: https://codesandbox.io/s/crimson-river-ymw1p?file=/src/App.js

**Expected behavior (Recommended)**
I'm not sure if this is a bug,
when i click toggle all row checkbox,it is not working for subrow

**Screenshots**
![2020-07-11 14-47-49](https://user-images.githubusercontent.com/47631325/87218545-7093f480-c386-11ea-9b3b-2fee9ad4ce6e.png)

**Desktop (please complete the following information):**

- OS:ubuntu20.04
- chrome
- latest

07harish:
Hi @pjshu,

I think you should place useExpanded hook before useRowSelect. [here is a working demo](https://codesandbox.io/s/modern-framework-119cg?file=/src/App.js)

I just checked the souce code: `useRowSelect` hook must be placed after these hooks: ['useFilters', 'useGroupBy', 'useSortBy']. But there is no `useExpanded` included in the array. So mostly this should be handled from the library, But the above change will solve this issue for you.

pjshu:
> Hi @pjshu,
> 
> I think you should place useExpanded hook before useRowSelect. [here is a working demo](https://codesandbox.io/s/modern-framework-119cg?file=/src/App.js)
> 
> I just checked the souce code: `useRowSelect` hook must be placed after these hooks: ['useFilters', 'useGroupBy', 'useSortBy']. But there is no `useExpanded` included in the array. So mostly this should be handled from the library, But the above change I mentioned will solve this issue.

thanks!!!slove my problem

PR #2543, opened 2020-07-16
fix: (issue #2528) Unexpected behaviour useRowSelect while selecting subRows

07harish:
Fix issue #2528 

[Codesandox to check the issue](https://codesandbox.io/s/crimson-river-ymw1p?file=/src/App.js)


07harish:
Will create PR after writing tests

PR #2545, opened 2020-07-16
chore: remove process reference in umd builds

riptusk331:
update rollup config to replace 'process.env.NODE_ENV' occurrences for umd builds.
this would prevent the build from loading via the <script> tag since theres no process in browser

### DIFF
--- a/rollup.config.js
+++ b/rollup.config.js
@@ -22,7 +22,7 @@ export default [
     },
     external,
     plugins: [
-      replace({ 'process.env.NODE_ENV': `"production"`, delimiters: ['', ''] }),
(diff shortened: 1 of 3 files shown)


rollup.config.js
```diff
@@ -20,7 +21,11 @@ export default [
       globals,
     },
     external,
-    plugins: [babel(), externalDeps()],
+    plugins: [
+      replace({ 'process.env.NODE_ENV': `"production"`, delimiters: ['', ''] }),
```
tannerlinsley, line 25: Would this need to be `development` instead of `production`?

riptusk331: my mistake - correcting

PR #2546, opened 2020-07-17
docs: refactor search modal

MrLeebo:
Same changes as I applied to the react-query docs in https://github.com/tannerlinsley/react-query/pull/773


tannerlinsley:
Beautiful!

PR #2576, opened 2020-07-27
feat(useSortBy): set sortBy state via method at once

gky360:
It would be nice and consistent if we have something like `setAllSortBy`, which sets sortBy state at once, similar to `setAllFilters`.

### DIFF
--- a/src/plugin-hooks/useSortBy.js
+++ b/src/plugin-hooks/useSortBy.js
@@ -15,6 +15,7 @@ import * as sortTypes from '../sortTypes'
 
 // Actions
 actions.resetSortBy = 'resetSortBy'
+actions.setAllSortBy = 'setAllSortBy'
 actions.toggleSortBy = 'toggleSortBy'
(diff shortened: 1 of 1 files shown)


src/plugin-hooks/useSortBy.js
```diff
@@ -15,6 +15,7 @@ import * as sortTypes from '../sortTypes'
 
 // Actions
 actions.resetSortBy = 'resetSortBy'
+actions.setAllSortBy = 'setAllSortBy'
```
gargroh, line 18: How about renaming `setAllSortBy` to `setSortBy`

gargroh: This will be in sync with reset thing...also similar to other plugins (https://github.com/tannerlinsley/react-table/pull/2570/files)

gky360: Thank you for checking this pull request.
That would be nice. I renamed them.

Issue #2578, opened 2020-07-27
UseRowSelect isn't telling row is selected when useGroupBy and useRowExpanded is use together 


maxime4000:
**Describe the bug**
When you make a table with useRowSelect, useGroupBy and useRowExpanded, expanded Row can be selected, but row.isSelected is undefined even if the row is selected. it seem that isExpand is pass through sub row but is selected isn't. 

**Provide an example via Codesandbox! **
[Example CodeSandBox here](https://codesandbox.io/s/gifted-proskuriakova-v2ejb?file=/src/TestTable.tsx)

**Steps To Reproduce (required)**

1. Expand a Status
2. Expand a Age
3. Unselect the row with the Status
4. Clear console
5. Click on Status Row, Age row and FirstName row and see the console.
6. FirstName row doesn't have the property of Selected or Grouped, but has the expanded properties 

**Expected behavior (Recommended)**
When selecting a "Sub Row", the row should has isSelected working and isSomeSelected working in Group Row. While those aren't passdown, isExpanded is pass through Sub Row.

**Screenshots**
![image](https://user-images.githubusercontent.com/4070713/88564101-cca27e00-d000-11ea-944e-3e7802990b92.png)

**Additional context**
I think the code that is suppose to pass down those properties are in those files: 
-   useGroupBy.js see **groupUpRecursively** function
-   utils,js see **expandRows** function

maxime4000:
Ok so the hooks definition is important and useRowSelect should be after useExpanded is use...
## [Wrong :](https://codesandbox.io/s/gifted-proskuriakova-v2ejb?file=/src/TestTable.tsx) 
    useRowSelect, -- Error when before useExpanded
    useExpanded,
## [Right :](https://codesandbox.io/s/admiring-shannon-v4l1l?file=/src/TestTable.tsx)
    useExpanded,
    useRowSelect, -- Perfectly working

tannerlinsley:
This should be fixed now :) Thanks for the help!

bgilm:
I did a fork of this sandbox, I wonder why using the `paginateExpandedRows: false` makes the selection break, it looks like you can no longer select subRows from the parent and the preparedRow (subRow) has undefined selection props

https://codesandbox.io/s/blazing-pond-4dqpq?file=/src/TestTable.tsx

@tannerlinsley

gargroh:
>it looks like you can no longer select subRows from the parent and the preparedRow (subRow) has undefined selection props

[...]

PR #2581, opened 2020-07-28
Add useExpanded in the pluginOrder 

maxime4000:
Selection get broken when rows are expanded if the plugin order is wrong.
#2578


maxime4000:
https://github.com/tannerlinsley/react-table/issues/2578

PR #2584, opened 2020-07-30
fix: Use references to renderers instead of inline functions

kizu:
In order to be able to check if a column has a default renderer, we need them to be exposed constants instead of inline functions.

This way we could do, say:

```jsx
import { emptyRenderer } from 'react-table';
// …
if (column.Header === emptyRenderer) {
  // …
}
```

Right now these are inline functions that cannot be compared with anything, so having publicly available references to these could help a lot when developing plugin-hooks that work with columns.

### DIFF
--- a/src/publicUtils.js
+++ b/src/publicUtils.js
@@ -6,8 +6,11 @@ export const actions = {
   init: 'init',
 }
 
+export const defaultRenderer = ({ value = '' }) => value;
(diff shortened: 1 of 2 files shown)


nextglabs:
Very nice, but this seems like the `emptyRenderer ` function is missing from the Typescript types (tested with `@types/react-table@7.7.2`)

PR #2592, opened 2020-08-04
feat(withSelection): Improve integration with pagination

roginfarrer:
Copy of #2585 for the next branch


roginfarrer:
@tannerlinsley I'm not very familiar with the new v8 architecture, so let me know if I'm missing anything (esp when referring to `getInstance()` or `instance`)

tannerlinsley:
I'm pretty sure you got it! As you can see, it's easy to use instance/getInstance() to either watch or bypass effect dependencies. What do you think?

roginfarrer:
Ah makes sense! I like it.

It's all good on my end then 👍

PR #2633, opened 2020-08-13
fix: bug in sort subRows data and add test in useSortBy

leyinetwork:
Same as #2541 
Because this pull request #2443 changed nested subRows sort when subRows data length is one
This time add test case in useSortBy.test.js


leyinetwork:
@tannerlinsley Can you take a look?

src/plugin-hooks/useSortBy.js
```diff
@@ -337,10 +337,7 @@ function useInstance(instance) {
       // If there are sub-rows, sort them
       sortedData.forEach(row => {
         sortedFlatRows.push(row)
-        if (!row.subRows) {
-          return
-        } else if (row.subRows.length === 1) {
-          sortedFlatRows.push(row.subRows[0])
+        if (!row.subRows || row.subRows.length < 1) {
```
aaryn101, line 340:
```suggestion
        if (!row.subRows || row.subRows.length === 0) {
```

PR #2635, opened 2020-08-13
Update material-ui-components.md

toczeksiedlce:
when UI is written in small letters we got 404. Capitals works.


tannerlinsley:
This will be fixed in a different PR. Thanks!

PR #2639, opened 2020-08-13
fix(useFlexLayout): change table min width to be the sum of min width of all columns

Abdallatif:
Table minimum width should be the sum of minimum width of all columns so the table can be responsive and shrink to the minimum width possible.

### DIFF
--- a/src/plugin-hooks/useFlexLayout.js
+++ b/src/plugin-hooks/useFlexLayout.js
@@ -14,7 +14,7 @@ const getTableProps = (props, { instance }) => [
   props,
   {
     style: {
-      minWidth: `${instance.totalColumnsWidth}px`,
+      minWidth: `${instance.totalColumnsMinWidth}px`,
     },
   },
 ]


mariusespejo:
This change (or something else introduced in v7.6.0) seems to have introduced a major breaking change on a minor version bump, see #2843. @tannerlinsley Please consider reverting the change until it is fully tested.

tannerlinsley:
Fixes are not considered breaking if they alter the code to exhibit expected functionality as originally documented. So yes, it will change the outcome of your app, it’s changing it from incorrect to correct. If you prefer the buggy/incorrect implementation, feel free to lock your version to the previous patch version.

mariusespejo:
@tannerlinsley  thanks for you advice. I will do that for now. 

But your own documentation is NOT showing "expected functionality as originally documented". Visually it definitely doesn't look correct. Could you provide advice on how to fix this using the correct newer version? Will the docs be updated to reflect that? 

image

tannerlinsley:
Yeah I’ll fix the example sandbox to utilize the correct measurement.

marcolazzarini:
This "fix" causes columns not having a defined min-width to be cut-off the table container, that doesn't take all the available horizontal space like in previous versions. Can you provide an updated example to overcome this issue?

Abdallatif:
A simple fix would be to move `overflow` to be on the `table` itself https://codesandbox.io/s/magical-morning-pexwi?file=/src/App.js
You still need to fix the header to be sticky + cell styles

mariusespejo:
@Abdallatif I'm not sure I follow, your linked example is still broken:

image

Abdallatif:
That's what I meant by you still need to fix some style issues. It can be easily fixed by setting a border-bottom on td instead of tr. Check the code again https://codesandbox.io/s/magical-morning-pexwi?file=/src/App.js

AlexSav94:
This issue: https://github.com/tannerlinsley/react-table/issues/3095 seems to be connected with this changes

Sahero:
If you have the background or stripped table, this will not help. Check this one 
image

Codesandbox: https://codesandbox.io/s/wizardly-nash-owkng?file=/src/App.js

Any idea for this case?

PR #2666, opened 2020-08-24
fix(useRowSelect): verify row selection based on reducer state

centrodph:
Fix: 

- https://github.com/tannerlinsley/react-table/issues/2659
- https://github.com/tannerlinsley/react-table/issues/2171


pzi:
Thanks @centrodph, I was wondering why I was unable to select rows properly until I saw your comment about React StrictMode. Hopeful this gets merged 👍

ahmedrafayat:
After countless hours of scratching my head, your solution of removing strict mode helped. Thanks

PR #2669, opened 2020-08-24
Fix pagination on sort

anatha:
When navigating to any page that is not the first one, and then sorting - the paginator input is not updated.
This means it will show the previous page number from the state, instead of getting the updated page number (which is 1) from the props.


Omar-Trigui:
i have the version 6.11.5 but still have the old changes
Capture

Nabil-MEDIOUNI:
Hi @tannerlinsley 
I faced the same issue, I can jump back to the first page after filtering the table.
Are any solutions to this problem yet?

PR #2706, opened 2020-09-04
example: Remove irrelevant `groupByBoundary`

gargroh:


utkarshgupta137:
This line is present in https://github.com/gargroh/react-table/blob/master/examples/kitchen-sink/src/App.js#L326 as well. Any replacements for this feature?

PR #2715, opened 2020-09-08
Feat/spanning

alinauroz:


alinauroz:
This PR introduces rowpsan in block layout. Rowspanning doesn't affect affect virtualization. But this PR has some style limits.

Issue #2701, opened 2020-09-02
initialCellStateAccessor should be called with `Cell` object instead of row original


pavelpipus:
initialCellStateAccessor should be called with `Cell` object instead of row original
Documentation says that If this function is defined, it will be passed a Cell object, from which you can return a value to use as the initial state, eg. cell => cell.row.original.initialCellState[cell.column.id]

Issue #2102, opened 2020-04-05
`useRowState` seems to have a logical error


bradennapier:
Since I am clearly just showing in your code exactly where the issue is, skipping the template there.

At the moment using the `useRowState` seems to have some bugs in the `initialAccessor` logic for both row and cell states.

https://github.com/tannerlinsley/react-table/blob/master/src/plugin-hooks/useRowState.js#L156

Documentation says:

```
initialRowStateAccessor: Function(originalRow) => Object<any>
Optional
Defaults to: row => ({})
This function should return the initial state for a row.
If this function is defined, it will be passed a Row object, from which you can return a value to use as the initial state, eg. row => row.original.initialState
initialCellStateAccessor: Function(originalRow) => Object<any>
Optional
Defaults to: cell => ({})
This function should return the initial state for a cell.
If this function is defined, it will be passed a Cell object, from which you can return a value to use as the initial state, eg. cell => cell.row.original.initialCellState[cell.column.id]
```

So while the type shows `originalRow` - the description indicates it would give the row or cell object (which is probably what it should be since its fairly useless for every sell to get the original row without knowing any information about itself even anyway.  That should only be executed a single time if that is the case.  

Seems that both the row and cell should indeed receive the row or cell object as the docs indicate rather than the raw `originalRow` value being sent.  At this point I can not see any purpose to using initialCellStateAccessor

picosam:
Hello @tannerlinsley, is this a verified bug?

PR #2719, opened 2020-09-09
fix(useRowState): change argument for "initialRowStateAccessor" and "initialCellStateAccessor"

ivaskevych:
pass `row` / `cell` to `initialRowStateAccessor` / `initialCellStateAccessor` accordingly

closes(#2102, #2701)


tannerlinsley:
FYI, row state is either *probably* being removed or replaced in favor of other more practical approaches. More info to come.

PR #2724, opened 2020-09-11
feat(useSortBy): Add option to disable sorting of subrows

rctbusk:
Add flag to initial state called disableSubRowSortBy which, if true, will not sort the sub rows in a table. This is to increase performance for long tables where each row has sub rows. Sorting of sub rows is done even if the sub rows are not showing, which can lead to performance issues. The default value of this flag is false, so existing tables will not see any changes in how they work.


tannerlinsley:
We'll be handling this in v8, but in a different way that allows disabling at a specific nesting level.

PR #2733, opened 2020-09-17
Add RTL support - update column resizing example

dadwic:
RTL support is NOT related to CSS.
deltaX calculation is different in RTL direction
**useResizeColumns hook** [Line 183](https://github.com/dadwic/react-table/commit/230a75e8350ce7713b2241334c430a798fd91719#diff-d5222209c13f3ad3c77a75a493635e56R183)

### DIFF
--- a/examples/column-resizing/package.json
+++ b/examples/column-resizing/package.json
@@ -7,6 +7,7 @@
     "eject": "rescripts eject"
   },
   "dependencies": {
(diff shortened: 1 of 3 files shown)


tannerlinsley:
I think we're going to be handling RTL in v8 across the board. Would you mind taking a look at the next branch and applying this PR to that instead?

PR #2737, opened 2020-09-19
fix(useGroupBy): crash on selecting row with grouping

RaiVaibhav:
Fixes
- https://github.com/tannerlinsley/react-table/issues/2465

Reason -

rowsById was getting changed after grouping, but we only need update.


RaiVaibhav:
@tannerlinsley this will change in the v8? should I close this PR?

tannerlinsley:
Yes, this is changing in v8. Closing this for now as it would be a breaking change.

PR #2754, opened 2020-09-28
fix: do not crash when headerIdWidths is undefined

FezVrasta:
Since we upgraded React Table we started getting this error when we try to resize a column:

image

This fix should prevent the code from erroring.


FezVrasta:
Any chance this could be reviewed and eventually merged?

PR #2815, opened 2020-10-28
fix(useExpanded) (useRowSelect): Memoize some variables

gargroh:
Avoid expensive calculation of `isAllRowsExpanded` on re-renders

### DIFF
--- a/src/hooks/useTable.js
+++ b/src/hooks/useTable.js
@@ -396,6 +396,12 @@ export const useTable = (props, ...plugins) => {
 
   getInstance().prepareRow = React.useCallback(
     row => {
+      if (row.preparedWith === getInstance().state) {
+        return
+      }
+
+      row.preparedWith = getInstance().state
+      
       row.getRowProps = makePropGetter(getHooks().getRowProps, {
         instance: getInstance(),
         row,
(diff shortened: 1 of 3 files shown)


gargroh:
Reason for doing this - `Object.keys(obj).length` is a costly operation, when obj is having ~100K of keys, it takes around ~100-150ms time, so memoizing this can save time when these values are actually not required to be re-computed.

src/hooks/useTable.js
```diff
@@ -396,6 +396,12 @@ export const useTable = (props, ...plugins) => {
 
   getInstance().prepareRow = React.useCallback(
     row => {
+      if (row.preparedWith === getInstance().state) {
+        return
+      }
+
+      row.preparedWith = getInstance().state
```
gargroh, line 403: Helpful when virtualization is done, this helps in skipping prepareRow call if it has already been called once with same table state..

src/plugin-hooks/useRowSelect.js
```diff
@@ -256,9 +256,12 @@ function useInstance(instance) {
     return selectedFlatRows
   }, [rows, selectSubRows, selectedRowIds, getSubRows])
 
-  let isAllRowsSelected = Boolean(
-    Object.keys(nonGroupedRowsById).length && Object.keys(selectedRowIds).length
-  )
+  let isAllRowsSelected = React.useMemo(() => {
```
gargroh, line 259: Reason for doing this - Object.keys(obj).length is a costly operation, when obj is having ~100K of keys, it takes around ~100-150ms time, so memoizing this can save time when these values are actually not required to be re-computed.

gargroh:
@tannerlinsley please have a look at this, let me know if something else is required on this part.

PR #2820, opened 2020-10-29
Clarify acceptable return values for `sortType`

jclem:
Since the [implementation](https://github.com/jclem/react-table/blob/v7.6.1/src/plugin-hooks/useSortBy.js#L387) just uses `Array.prototype.sort`, I wonder if we can change the documented return type for the `sortType` function to reflect that. Likewise, `0` is presumably also an acceptable return value, and we should document that, as well.

The reason for what may seem like a rather pedantic change here is to avoid unnecessarily having to convert comparison return values from functions like [`localeCompare`](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/String/localeCompare) (which only specifies generally a negative, positive, or zero value) to `-1` or `1` in order to align with expectations set by react-table documentation.


jclem:
Closing this to clear it out of my personal backlog.

PR #2886, opened 2020-11-24
fix(useRowSelect): Correct way to fetch subRows

gargroh:
Currently in `useRowSelect`, `getSubRows(row)` is used to fetch child rows, where `row` is react-table's row structure, hence it will not work with custom `getSubRows`,  we can use `row.subRows` in place of that to correctly fetch child rows.


gargroh:
Minimal replication of bug is at https://codesandbox.io/s/useexpanded-userowselect-parent-indeterminate-bug-i0zoy?file=/src/App.js (check by selecting row, and observe parent checkbox is not changed to indeterminate)

karosowka13 (commented):
Would be nice if this was merged

PR #2966, opened 2021-01-05
fix(useRowSelect): fix getToggleAllRowsSelectedProps logic

utkarshgupta137:
- only return true when at least one of the selectedRows is present in data

Steps to reproduce the bug:
1. Create a table with row selection and autoResetSelectedRows: false and getRowId uniquely identifying the row.
2. Select a row and delete it.
3. intermediate is still true even though none of the rows in data is selected.

This is useful when implementing a row select and delete. When I click undo delete, I want the rows to still be selected. This helps with that.


utkarshgupta137:
@tannerlinsley I don't understand why the Vercel check is failing, but this is ready to review.

PR #2994, opened 2021-01-17
fix(useGridLayout): handle column.width & hidden columns

utkarshgupta137:
Steps to reproduce the bug:
1. Create a table with useGridLayout & add width attribute to at least one of the columns
2. => width attribute is not respected
3. Hide a column => The table is all messed up because it is still rendered with the default number of columns

CSS grids provide awesome ways of managing widths, being able to use that with react-table would be awesome.


golota60:
this fix is just what i need 😄 @tannerlinsley are there any plans on merging & releasing this?

probablyup:
@tannerlinsley ping

PR #3021, opened 2021-01-26
Fix bug of includes filter

aruprakshit:
Reject rows which is having falsy values like `null`, `undefined`.


Keshavdulal:
Would love your take on this @tannerlinsley whenever you are free.
I am also facing some issues with this one and [my initial workaround didn't last long](https://github.com/tannerlinsley/react-table/issues/3020#issuecomment-775180900).

sicarius97:
@tannerlinsley I feel as though this easy patch is actually a pretty big deal for a lot of projects. Can you perhaps take a look at this?

PR #3048, opened 2021-02-08
Document available filter functions

raq929:
The available filter functions were previously undocumented.


raq929:
Any chance this can get a review?

PR #3130, opened 2021-03-10
fix(useFilters): memoize column.setFilter

### LINKED ISSUE #3129: column.setFilter isn't memoized
`column.setFilter` isn't memoized (and leads to excessive update)

`API rate limit exceeded for app ID c07a89833b557afc7be2.`   
Providing a MR, fix is quite simple

1. Use useFilter
2. in a filter make use of column.setFilter in a useEffect that calls setFilter on change
3. React should say maximum depth exceeded

PaulCordo:
Close #3129


PaulCordo:
Damn, was too quick to fix, not as simple as imagined.
Will try to come with a solution but I'll use general and memoized setFilter

PR #3228, opened 2021-04-21
Fix bug in text filters where null is incorrectly matched

brianpursley:
Text filters (text, exactText, and exactTextCase) incorrectly match null cell values when filterValue is the string "null".
This change coerces null to empty string before comparing to the filterValue to avoid the unexpected match.


brianpursley:
Ping @tannerlinsley 

it’s ok if you don’t want to merge this, but just wanted to make you aware of this weird behavior when filtering on a column that has null values.

if you want to handle it a diff way, that’s fine and I can close this pr.

PR #3235, opened 2021-04-25
Add two new built in sort types 

goldflag:
Add two new build in sort types. Current sort types are unchanged.
Suggested in https://github.com/tannerlinsley/react-table/issues/3093

1.  `number`
Sorts only numerical values inside column.

2.  `string`
Does what is suggersted in https://github.com/tannerlinsley/react-table/issues/3137.
Sorts case-insensitively until there is a tie and then sorts case-sensitive.

### DIFF
--- a/src/sortTypes.js
+++ b/src/sortTypes.js
@@ -4,8 +4,8 @@ const reSplitAlphaNumeric = /([0-9]+)/gm
 // It handles numbers, mixed alphanumeric combinations, and even
 // null, undefined, and Infinity
 export const alphanumeric = (rowA, rowB, columnId) => {
-  let a = getRowValueByColumnID(rowA, columnId)
-  let b = getRowValueByColumnID(rowB, columnId)
+  let [a, b] = getRowValuesByColumnID(rowA, rowB, columnId)
+
   // Force to strings (or "" for unsupported types)
   a = toString(a)
   b = toString(b)
@@ -52,10 +52,8 @@ export const alphanumeric = (rowA, rowB, columnId) => {
 
   return a.length - b.length
 }
-
 export function datetime(rowA, rowB, columnId) {
-  let a = getRowValueByColumnID(rowA, columnId)
-  let b = getRowValueByColumnID(rowB, columnId)
+  let [a, b] = getRowValuesByColumnID(rowA, rowB, columnId)
 
   a = a.getTime()
   b = b.getTime()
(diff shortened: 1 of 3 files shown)


src/sortTypes.js
```diff
@@ -64,8 +62,51 @@ export function datetime(rowA, rowB, columnId) {
...
+  let [a, b] = getRowValuesByColumnID(rowA, rowB, columnId)
+
+  a = a.split('').filter(Boolean)
+  b = b.split('').filter(Boolean)
+
+  while (a.length && b.length) {
+    let aa = a.shift()
+    let bb = b.shift()
+
+    let alower = aa.toLowerCase()
```
ggascoigne, line 80: Shouldn't this use `localeCompare` and avoid all of the case conversions? This will give the wrong results in a lot of locales.

tannerlinsley: Yeah, it probably should. @goldflag would you please open a new PR for this?

src/sortTypes.js
```diff
@@ -64,8 +62,51 @@ export function datetime(rowA, rowB, columnId) {
...
-  let b = getRowValueByColumnID(rowB, columnId)
+  let [a, b] = getRowValuesByColumnID(rowA, rowB, columnId)
+
+  return compareBasic(a, b)
+}
+
+export function string(rowA, rowB, columnId) {
+  let [a, b] = getRowValuesByColumnID(rowA, rowB, columnId)
+
+  a = a.split('').filter(Boolean)
```
brianpursley, line 73: It seems this throws an exception if the value of `a` or `b` are null

brianpursley: Opened an issue to track this:
https://github.com/tannerlinsley/react-table/issues/3269

src/sortTypes.js
```diff
@@ -64,8 +62,51 @@ export function datetime(rowA, rowB, columnId) {
...
+
+  return a.length - b.length
+}
+
+export function number(rowA, rowB, columnId) {
+  let [a, b] = getRowValuesByColumnID(rowA, rowB, columnId)
+
+  const replaceNonNumeric = /[^0-9.]/gi
+
+  a = Number(String(a).replace(replaceNonNumeric, ''))
```
JulienLGRD, line 108: This creates an absolute value right? Is that really what we want here because it will break the sort if negative and positive values are compared?

PR #3247, opened 2021-04-29
Fix min-width style for rt-thead and rt-tbody

matiosfree:


matiosfree:
it's a fix for https://github.com/tannerlinsley/react-table/issues/3248

PR #3275, opened 2021-05-17
fix(useControlledState): memorize return value of useControlledState

huangbinjie:
The `reducerState` processed by useControlledState should be memorized, ant now it has its own state. I think it might be a bug.


huangbinjie:
seem i misunderstood something, i'll close it.

PR #3301, opened 2021-05-31
Enable multi-sort rows without a keyboard

thiagofernando:
- create a option to multi-sort columns without a shift key, just with mouse click;
- create a param enableMultiSortClick to flag this option;


gargroh:
You can use existing `isMultiSortEvent` cb for the required functionality,
returning `true` from the cb will enable multi sort with mouse click only.

PR #3310, opened 2021-06-07
fix(deps): include babel runtime in build

garand:
This updates the Rollup Babel plugin and also includes the Babel runtime in the final build, as recommended by Babel. The Rollup Babel plugin has babelHelpers set to "runtime", which is recommended for libraries in the plugin documentation.

https://www.npmjs.com/package/@rollup/plugin-babel

This is necessary for anyone using a tooling setup that does not include Babel out of the box. For instance Vite, etc.

## Reasoning

I was running into this error:

```
Uncaught ReferenceError: regeneratorRuntime is not defined
```

Specifically with useAsyncDebounce.

### DIFF
--- a/.babelrc
+++ b/.babelrc
@@ -9,6 +9,7 @@
     ],
     "@babel/react"
   ],
+  "plugins": ["@babel/plugin-transform-runtime"],
   "env": {
     "test": {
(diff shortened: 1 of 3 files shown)


drichar:
I'm getting this error as well on build with Next.js v10.2.3

PR #3359, opened 2021-07-01
Update useSortBy docs

Jeffdude:
According to `sortTypes.js` these are the new sortTypes.

Thanks all for your great work on this project.


Jeffdude:
Why was this abandoned? It actually updated an out of date doc. @tannerlinsley

PR #3406, opened 2021-07-27
add v6 peer dependency support for react 17

kochis:
This should allow `react-table-6` to be installed with React 17. Would it be possible to get this merged and distributed as a new version in the `react-table-6` lib?

### DIFF
--- a/package.json
+++ b/package.json
@@ -49,8 +49,8 @@
   },
   "peerDependencies": {
     "prop-types": "^15.7.0",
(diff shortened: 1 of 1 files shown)


evanseeds:
Is there anything holding this back? We're still installing with --legacy-peer-deps because of this

ayxos:
Any updates on this? this lib is fully compatible with react v17 right?

Issue #3071, opened 2021-02-16
react-table crashes using manualPagination, useRowSelect, and toggleRowSelected


anjunatl:
## Describe the bug (required)

When using react-table with row selection & server-side pagination configured, `toggleAllRowsSelected(false)` only affects visible rows and `toggleRowSelected(rowId, false)` crashes the page with the following error when it is called on a row that is no longer visible: `Uncaught TypeError: Cannot read property 'isGrouped' of undefined` at [`handleRowById` in `useRowSelect.js` on line 157.](https://github.com/tannerlinsley/react-table/blob/master/src/plugin-hooks/useRowSelect.js#L157)  

## Provide an example

Since this is a complicated bug (using server-side pagination, row selection, and batch row action behavior that resets the selected rows on completion) I've created a docker-based test environment to demonstrate the issue: 

https://github.com/anjunatl/react-table-bug-repro/ 

[Here's the usage of `react-table` in context.](https://github.com/anjunatl/react-table-bug-repro/blob/main/demo/laravel/resources/js/DemoGrid.jsx) [And here are more files of note for the example if it helps](https://github.com/anjunatl/react-table-bug-repro/#example-source-highlights)

I have a fix for this to enable `toggleRowSelected` to work on invisible rows without crashing, but I'm not sure if it both fully fixes the problem and does not also cause more trouble in concert with other plugin hooks - 
https://github.com/anjunatl/react-table-bug-repro#potential-fix 

### toggleAllRowSelected - not all rows get cleared

1. Clone the reproduction repo & run `docker-compose up` to start the example
2. Choose one to many rows on the first page
3. Change to another page
4. Choose another one to many rows
5. Click the "Do something" button

#### Expected behavior

The selected row counter should reset to 0.

#### Actual behavior

The selected row counter is not 0 & is the sum of selected rows on non-visible pages

### toggleRowSelected - crashes

1. Clone the reproduction repo & run `docker-compose up` to start the example
2. Click the "Mode" button to change it from "All" to "Row" to enable use of `toggleRowSelected`
3. Choose one to many rows on the first page
4. Change to another page
5. Choose another one to many rows
6. Click the "Do something" button

#### Expected behavior

The selected row counter should reset to 0.

#### Actual behavior

The page crashes

## Screenshots

[video screenshot](https://user-images.githubusercontent.com/157208/108133591-d2d27780-7082-11eb-85ca-64950c90694a.mp4)

## Desktop (please complete the following information):

[...]

Issue #2833, opened 2020-11-03
Cannot read property 'isGrouped' of undefined 


kareshi:
**Describe the bug (required)**

I am using all those hooks in react-table :
- useRowSelect
- useExpanded 
- useRowSelect 

After I group by a column and select a row, the below error will occur :

TypeError
Cannot read property 'isGrouped' of undefined
handleRowById
https://2u2zj.csb.app/node_modules/react-table/dist/react-table.development.js:3571:18
eval
https://2u2zj.csb.app/node_modules/react-table/dist/react-table.development.js:3581:20
handleRowById
https://2u2zj.csb.app/node_modules/react-table/dist/react-table.development.js:3580:34
eval
https://2u2zj.csb.app/node_modules/react-table/dist/react-table.development.js:3581:20
handleRowById
https://2u2zj.csb.app/node_modules/react-table/dist/react-table.development.js:3580:34
reducer$8
https://2u2zj.csb.app/node_modules/react-table/dist/react-table.development.js:3586:7
eval
https://2u2zj.csb.app/node_modules/react-table/dist/react-table.development.js:1055:16
eval
https://2u2zj.csb.app/node_modules/react-table/dist/react-table.development.js:1054:126
updateReducer
https://2u2zj.csb.app/node_modules/react-dom/cjs/react-dom.development.js:10083:26
Object.useReducer
https://2u2zj.csb.app/node_modules/react-dom/cjs/react-dom.development.js:10723:20
Object.useReducer
https://2u2zj.csb.app/node_modules/react/cjs/react.development.js:918:25
useTable
https://2u2zj.csb.app/node_modules/react-table/dist/react-table.development.js:1059:35
Table
/src/App.js:85:4
  82 |   selectedFlatRows,
  83 |   state: { pageIndex, pageSize, selectedRowIds }
  84 | } = useTable(
> 85 |   {
     |  ^
  86 |     columns,
  87 |     data
  88 |   },

Codesandbox :
https://codesandbox.io/s/react-table-userowselect-useexpanded-userowselect-issue-2u2zj

**Steps To Reproduce (required)**

1. Go to https://codesandbox.io/s/react-table-userowselect-useexpanded-userowselect-issue-2u2zj
2. Click on column "First Name"
3. Select first row by clicking on the check box at the beginning of the line

**Expected behavior (Recommended)**
No error is happening.

A first guess analysis is that subRows has been set manually with makeData. 
When grouping the data, the ids has not been generated for those rows.

[...]

PR #3422, opened 2021-08-12
fix: check for row existence in handleRowById

jschen2:
This will fix the errors referenced by #2833 and #3071. This will also fix a "TypeError: Cannot read property 'subRows' of undefined" error. These errors appear when a row does not exist, whether removed from the data or filtered out, but there is still a reference in the selected state.

### DIFF
--- a/src/plugin-hooks/useRowSelect.js
+++ b/src/plugin-hooks/useRowSelect.js
@@ -154,16 +154,18 @@ function reducer(state, action, previousState, instance) {
     const handleRowById = id => {
       const row = rowsById[id]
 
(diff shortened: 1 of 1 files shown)


theopak:
this looks great! i notice that this pr has been open for a long time, and now there's a similar pr that applies the same fix plus some unit tests: #3440

simcha90:
@jschen2 hi any updates here?

tannerlinsley:
Will merge this if we can get the conflicts resolved.

jschen2:
Changed to the correct base (v7) to remove conflicts

PR #3427, opened 2021-08-16
Fix typo

chuchuva:


examples/editable-data/src/App.js
```diff
@@ -254,7 +254,7 @@ function App() {
     )
   }
 
-  // After data chagnes, we turn the flag back off
+  // After data changes, we turn the flag back off
```
theopak, line 257: 📝 👍

theopak (commented):
i'm using react-table too, and i just wanted to comment that this is a cool one-line fix of an example :)

PR #3429, opened 2021-08-16
fix: table crash when data is null or undefined

paulizio:
This PR provides a fix for situation when tabledata is undefined or null,the table will crash with error 'TypeError: Cannot read property 'forEach' of undefined'

### DIFF
--- a/src/hooks/useTable.js
+++ b/src/hooks/useTable.js
@@ -588,7 +588,7 @@ function accessRowsForColumn({
     )
   }
 
-  data.forEach((originalRow, rowIndex) =>
+  data?.forEach((originalRow, rowIndex) =>
     accessRow(originalRow, rowIndex, 0, undefined, rows)
   )
 }


sicarius97:
@tannerlinsley This is actually a big problem that is a pretty simple fix. If there are any null values in a table right now, it crashes the whole component

sicarius97:
Since neither version 12 or the alpha have the fix merged, I have to maintain my own separate fork for my projects currently where the ability to handle null values is a requirement

sicarius97:
@tannerlinsley This still needs a fix

KevinVandy:
Sorry, why can't you just pass data: data ?? [] to useTable?

sicarius97:
@KevinVandy Its not that the whole table is null, just that the array of data has a particular value that is null. Then if you try to sort or filter the data, the whole component crashes

tannerlinsley:
Yeah, null or inaccessible row values should either be filtered out or remapped/transformed into a homogenous row shape.

sicarius97:
@tannerlinsley I mean I kinda disagree. First off, I'm pretty sure it says in the docs that a cell supports values of any valid JSON datatype, of which null is a perfectly valid JSON datatype. Sure I could do some transforms beforehand and cast all null values to strings, but of course `"null"` and `null` are not the same thing and shouldn't be treated the same. A row in a table should be able to have a column in which does not have a value for that particular row when data is incomplete, which it very often is in a lot of applications. It would be pretty silly for like microsoft excel to not allow you to have empty cells in a row and not very user friendly if the whole excel app froze and crashed when trying to generate a graph from an incomplete dataset or filter a column that had an empty cell

Secondly for like a table that might take user input to complete incomplete data, it is definitely more human readable as a literal absence of a value if there is no apparent value rendered in that row, like would happen for the datatype `null`, but of course if that was a string then to a human it might appear as though the row is complete in its data when it really isn't. Plus it is also possible for a user to manually enter a cell as blank, save it, and then if they try to filter it at that point, it will also crash.

Lastly, it just makes more sense, both for dev quality of life using the lib and for performance reasons. I understand having to remap or write transformations in order to ensure a homogenous row shape, which is normal for tables, but I think a lot of the advantages in using some of the hooks this library exposes are moot if I have to filter, sort, and/or transform normal json data types ahead of time before I am able to utilize the tables internal sorting, filtering, etc without my whole client crashing due to one table component (which it will literally do). 

Plus things are never perfect, an api might return homogenous perfect data 99.9% of the time and then that .1% of the time a null value just so happens to be included in the response, the users whole client crashes, and you shouldn't need to have a bunch of client side transforms running on every api request for table data to compensate for 1 in 1000 requests.

I hope that makes more sense, thanks for your response so far. Happy to discuss this further too and contribute as needed. Seeing as there are a couple of different issues that reference these same things, I certainly don't think I am the only react-table user out there that feels this way

sicarius97:
@tannerlinsley There are still more issues being opened about this exact problem, I would be happy to work on an improved solution, but I still feel strongly aligned to some of the sentiments and issues that this causes above^

Perhaps this functionality might still be considered? Maybe this initial implementation isn't quite the right fix, but it still really needs some sort of fix. Else I guess our companies projects will either try to find an alternative or publish and maintain our own patched custom version that fixes these issues and a couple more

tannerlinsley:
You’re more than welcome to try your hand at a PR.

nicolad:
Any update on this?

> Lastly, it just makes more sense, both for dev quality of life using the lib and for performance reasons.

💯 agree with this, it's safer to handle null/ undefined values at table state as a safe default behavior

sicarius97:
@nicolad No dice as far as I know, my original PR that would fix the crashing on null/undefined got rejected and I haven't had time to put out a new one. I can see if I can work something up if you are also running into the issue

PR #3440, opened 2021-08-25
fix(useRowSelect): do not throw if toggleRowSelected id not found in current instance

octatone:
Encountered an exception when a component callback held reference to a row that was deleted by the time it was executed. This PR adds a small change to bail early in the reducer if the `toggleRowSelected` action's row id is not found in the current instance.

### DIFF
--- a/src/plugin-hooks/tests/useRowSelect.test.js
+++ b/src/plugin-hooks/tests/useRowSelect.test.js
@@ -1,9 +1,10 @@
 import React from 'react'
 import { render, fireEvent } from '@testing-library/react'
 import { useTable } from '../../hooks/useTable'
-import { useRowSelect } from '../useRowSelect'
+import { useRowSelect, reducer } from '../useRowSelect'
 import { useExpanded } from '../useExpanded'
 import { usePagination } from '../usePagination'
+import { actions } from '../../publicUtils'
 
 const dataPiece = [
   {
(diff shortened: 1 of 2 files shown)


src/plugin-hooks/useRowSelect.js
```diff
@@ -153,6 +153,7 @@ function reducer(state, action, previousState, instance) {
 
     const handleRowById = id => {
       const row = rowsById[id]
+      if (row === undefined) return
```
theopak, line 156: from looking through some other files, i think it might be a repo convention to include the `{` `}` like this:

```suggestion
      if (row === undefined) {
        return
      }
```

src/plugin-hooks/tests/useRowSelect.test.js
```diff
@@ -382,3 +383,26 @@ test('renders a table with selectable rows, only selecting the current page', ()
 
   expect(rtl.queryAllByText('Selected').length).toBe(0)
 })
+
+describe('reducer', () => {
+  describe('toggleRowSelected', () => {
+    test('does not throw when row not found', () => {
```
theopak, line 389: hey look at that! nice unit tests 👍

theopak (commented):
wonderful! i see the same issue when using react-table and i agree that this is a clean way to fix it 

- closes #2833 
- closes #3071

faharmonyuser:
what's the status of this PR? is it going to be merged anytime soon?

cduflo:
+1 we are looking to have this fixed for our implementation as well

Issue #3061, opened 2021-02-11
`isAllRowsExpanded` incorrect state when running in strict mode


adamb70:
When using `React.StricttMode` you are unable to collapse all rows using the "expand/collapse all" button after expanding them. This can be seen in the React Table "Expanding" official example just by running the app in strict mode.

https://codesandbox.io/s/exciting-bartik-hrbeu?file=/src/index.js

1. Expand all rows
2. Try to collapse all rows

This is possibly related to #3008, however I am only seeing this behaviour in strict mode.

blovato:
I am also encountering this issue! However, I am not running my table in Strict Mode so it might be a more general bug with this expand all logic.
I've disabled this button for now, but I'd love to be able to use it as soon as possible.

dumbbell:
I confirm this behavior. Disabling strict mode works around the problem.

d-mather:
I'm also experiencing this issue (even without StrictMode). Is there any update to this being fixed?

Bromilow:
has there been any progress on this issue?

twjackysu:
I have the same problem

PR #3471, opened 2021-09-18
fix(useExpanded): state of isAllRowsExpanded on React.StrictMode

yhor1e:
#3061 mentions below.

> When using React.StricttMode you are unable to collapse all rows using the "expand/collapse all" button after expanding them. This can be seen in the React Table "Expanding" official example just by running the app in strict mode.
> https://codesandbox.io/s/exciting-bartik-hrbeu?file=/src/index.js

I change `isAllRowsExpanded` from `instance` to recalculating it by using `state` because `isAllRowsExpanded` from `instance` is toggled twice on React.StrictMode. Once `isAllRowsExpanded` from `instance` gets `true`, it toggles twice and always returns `true` whenever the "Toggle All Rows Expanded" button is clicked.  

**Test result**

```bash
 PASS   unit  src/plugin-hooks/tests/useExpanded.test.js
  ✓ renders an expandable table (91ms)

Test Suites: 1 passed, 1 total
Tests:       1 passed, 1 total
Snapshots:   0 total
Time:        3.615s
Ran all test suites matching /src\/plugin-hooks\/tests\/useExpanded\.test\.js/i.
```
**Applied example**

https://github.com/yhor1e/react-table-fix-state-of-isallrowsexpanded-on-strictmode

In the above repository, I applied this PR to the [expanding example](https://github.com/tannerlinsley/react-table/tree/8c77b4ad97353a0b1f0746be5b919868862a9dcc/examples/expanding).

Please check this PR. Thanks.

### DIFF
--- a/src/plugin-hooks/useExpanded.js
+++ b/src/plugin-hooks/useExpanded.js
@@ -69,7 +69,10 @@ function reducer(state, action, previousState, instance) {
 
   if (action.type === actions.toggleAllRowsExpanded) {
     const { value } = action
-    const { isAllRowsExpanded, rowsById } = instance
+    const { rowsById } = instance
+
+    const isAllRowsExpanded =
+      Object.keys(rowsById).length === Object.keys(state.expanded).length
 
     const expandAll = typeof value !== 'undefined' ? value : !isAllRowsExpanded
 


vlade2008:
Any updates ?

CGamesPlay:
This is still an issue in `7.7.0` and does not seem to be a breaking change.

PR #3481, opened 2021-09-27
feat: add support for shift-selecting rows

ankarhem:
Add functionality to select or deselect the range of rows
between the previously and currently selected row. Matches
functionality found in services like gmail

### DIFF
--- a/src/plugin-hooks/useRowSelect.js
+++ b/src/plugin-hooks/useRowSelect.js
@@ -1,9 +1,8 @@
 import React from 'react'
-
 import {
   actions,
-  makePropGetter,
   ensurePluginOrder,
+  makePropGetter,
   useGetLatest,
   useMountedLayoutEffect,
 } from '../publicUtils'
@@ -43,7 +42,7 @@ const defaultGetToggleRowSelectedProps = (props, { instance, row }) => {
     props,
     {
       onChange: e => {
-        row.toggleRowSelected(e.target.checked)
+        row.toggleRowSelected(e.target.checked, e.nativeEvent.shiftKey)
       },
       style: {
         cursor: 'pointer',
@@ -96,6 +95,8 @@ function reducer(state, action, previousState, instance) {
   if (action.type === actions.init) {
     return {
       selectedRowIds: {},
+      lastSelectedRowId: null,
+      lastShiftSelectedRowId: null,
       ...state,
     }
   }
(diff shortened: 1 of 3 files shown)


src/plugin-hooks/useRowSelect.js
```diff
@@ -96,6 +95,8 @@ function reducer(state, action, previousState, instance) {
   if (action.type === actions.init) {
     return {
       selectedRowIds: {},
+      lastSelectedRowId: null,
+      lastShiftSelectedRowId: null,
```
gargroh, line 99: `lastShiftSelectedRowId` can be removed..

gargroh:
Tests and document helper text can be helpful :)

src/plugin-hooks/useRowSelect.js
```diff
@@ -167,7 +168,23 @@ function reducer(state, action, previousState, instance) {
...
 
-    handleRowById(id)
+    if (shiftKey && state.lastSelectedRowId && state.lastSelectedRowId !== id) {
+      const previousIndex = rows.findIndex(
+        row => row.id === state.lastSelectedRowId
+      )
+      const currentIndex = rows.findIndex(row => row.id === id)
+      const startIndex = Math.min(previousIndex, currentIndex)
+      const endIndex = Math.max(previousIndex, currentIndex)
+      const rowIdsToToggle = Object.keys(rowsById).slice(
```
gargroh, line 178: don't depend for `rows ids sequence` on `Object.keys(rowsById)`, as it will not change after operations like sort etc. hence selecting rows via shift can produce not required effect.

gargroh: [sidenote] Also, Object.keys does not guarantee order on some envs.

ankarhem: Thanks for the feedback. I'll look over it.

src/plugin-hooks/useRowSelect.js
ankarhem: Oops. Also I wasn't sure what the convention would be here. Should they still be initialized or what about leaving it undefined until first set?

ankarhem:
> Tests and document helper text can be helpful :)

I added testing for a sorted table. I'm unsure what would need to go in the documentation though. Is this not something we'd expect to always be active and working? (i.e no toggle to enable/disable)

PR #3542, opened 2021-11-14
fix(useGridLayout): Support header groups

### LINKED ISSUE #3537: useGridLayout does not support header groups
Headers with nested columns do not render correctly with the `useGridLayout` hook because the css grid-column property is not set to make individual cells span multiple columns.

e.g. This column structure (from several examples with other layout hooks) does not render correctly.
```javascript
  const columns = React.useMemo(
    () => [
      {
        Header: 'Name',
        columns: [
          {
            Header: 'First Name',
            accessor: 'firstName',
          },
          {
            Header: 'Last Name',
            accessor: 'lastName',
          },
        ],
      },
      {
        Header: 'Info',
        columns: [
          {
            Header: 'Age',
            accessor: 'age',
            width: 50,
            maxWidth: 60,
          },
          {
            Header: 'Visits',
            accessor: 'visits',
            width: 60,
          },
          {
            Header: 'Status',
            accessor: 'status',
          },
          {
            Header: 'Profile Progress',
            accessor: 'progress',
          },
        ],
      },
    ],
    []
  )

```

CollinsSpencer:
### Header Groups Bug
Resolves #3537 

`useGridLayout` could not display header groups by default. This seemed like a bug as it is fairly common functionality that it should be able to handle.

Header groups can be supported with the grid layout by using the css grid-column property with a span of the column's `totalVisibleHeaderCount`.

### Column Resizing Bug
`useGridLayout` has a bug that comes from 'auto' or 'fr' units. When some resizing columns, the cursor doesn't stay with the resizer handle because 'auto' width columns rescale based on the column being resized. e.g. columnResizing state thinks the column width is 150, but it is actually 200 because of grid 'auto' resizing.

This column resizing bug was fixed by locking the width of all columns not being resized. Now, while resizing, this behaves exactly like the Resizing example (with `useBlockLayout`). After resizing, all columns that have not been resized will default back to their previous value. As a bonus, this supports user-provided 'width' values on columns. If 'auto' width columns is desired (instead of the default 150), that can be set with the defaultColumn useTable prop.

`useGridLayout` also had a bug where columns could be resized beyond the minWidth and maxWidth values. These values are now taken into account when resizing. However, this MR does not apply minWidth and maxWidth during initial column sizing.

### DIFF
--- a/docs/src/pages/docs/api/overview.md
+++ b/docs/src/pages/docs/api/overview.md
@@ -28,6 +28,7 @@ React Table is essentially a compatible collection of **custom React hooks**:
     - `useBlockLayout`
     - `useAbsoluteLayout`
     - `useFlexLayout`
+    - `useGridLayout`
 - 3rd Party Plugin Hooks
   - [LineUp-lite Hooks](https://lineup-lite.netlify.app)
     - Core Plugin Hooks
--- a/docs/src/pages/docs/api/useGridLayout.md
+++ b/docs/src/pages/docs/api/useGridLayout.md
@@ -10,3 +10,14 @@
 - `getTableProps`
   - **Usage Required**
   - This core prop getter is required to get the correct stying for the layout
+
+### Row Properties
+
+- `getRowProps`
+  - This core prop getter is required for rendering sub-components as a full row
+
+### Header Properties
+
+- `getHeaderProps`
+  - **Usage Required**
+  - This core prop getter is required to to enable absolute layout for headers
--- a/src/plugin-hooks/tests/useGridLayout.test.js
+++ b/src/plugin-hooks/tests/useGridLayout.test.js
@@ -31,36 +31,45 @@ const data = [
 ]
 
 function Table({ columns, data }) {
-  const {
-    getTableProps,
-    headerGroups,
-    rows,
-    prepareRow,
-  } = useTable(
+    const defaultColumn = React.useMemo(
+      () => ({
+        minWidth: 30,
+        width: 'auto',
+        maxWidth: 400,
+      }),
+      []
+    )
+
+  const { getTableProps, headerGroups, rows, prepareRow } = useTable(
     {
       columns,
       data,
+      defaultColumn,
     },
-    useGridLayout,
+    useGridLayout
   )
 
   return (
     <div {...getTableProps()} className="table">
-      {headerGroups.map(headerGroup => (
+      {headerGroups.map(headerGroup =>
         headerGroup.headers.map(column => (
-          <div key={column.id} {...column.getHeaderProps()} className="cell header">
+          <div
+            key={column.id}
+            {...column.getHeaderProps()}
+            className="cell header"
+          >
             {column.render('Header')}
           </div>
         ))
-      ))}
-      {rows.map(row => 
-        prepareRow(row) || (
+      )}
+      {rows.map(
+        row =>
+          prepareRow(row) ||
           row.cells.map(cell => (
             <div {...cell.getCellProps()} className="cell">
               {cell.render('Cell')}
             </div>
           ))
-        )
       )}
     </div>
   )
@@ -70,28 +79,42 @@ function App() {
   const columns = React.useMemo(
     () => [
       {
-        Header: 'First Name',
-        accessor: 'firstName',
+        Header: 'Name',
+        columns: [
+          {
+            Header: 'First Name',
+            accessor: 'firstName',
+            width: 'auto',
+          },
+          {
+            Header: 'Last Name',
+            accessor: 'lastName',
+            width: 350,
+          },
+        ],
       },
       {
-        Header: 'Last Name',
-        accessor: 'lastName',
-      },
-      {
-        Header: 'Age',
-        accessor: 'age',
-      },
-      {
-        Header: 'Visits',
-        accessor: 'visits',
-      },
-      {
-        Header: 'Status',
-        accessor: 'status',
-      },
-      {
-        Header: 'Profile Progress',
-        accessor: 'progress',
+        Header: 'Info',
+        columns: [
+          {
+            Header: 'Age',
+            accessor: 'age',
+            minWidth: 300,
+          },
+          {
+            Header: 'Visits',
+            accessor: 'visits',
+            maxWidth: 150,
+          },
+          {
+            Header: 'Status',
+            accessor: 'status',
+          },
+          {
+            Header: 'Profile Progress',
+            accessor: 'progress',
+          },
+        ],
       },
     ],
     []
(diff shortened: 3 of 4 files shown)


CollinsSpencer:
Column `maxWidth` could be supported with `minmax`, which has good browser support https://caniuse.com/?search=minmax.

```javascript
// getTableProps
const width = typeof column.width === 'number' ? `${column.width}px` : column.width
const maxWidth = typeof column.maxWidth === 'number' ? `${column.maxWidth}px` : column.maxWidth
return `minmax(${width},${maxWidth})`
// instead of
if (typeof column.width === 'number') return `${column.width}px`
return column.width
```

However, I'm not sure if we could support both minWidth and maxWidth at the same time. There doesn't appear to be something like `clamp` for css grid. Unfortunately you can't do nested `minmax` functions. If each of the values were numeric, then we could do a `min` inside of the `minmax`. However, that removes a lot of the value of using css grid.

CollinsSpencer:
# Adapted Examples
(since there aren't any examples with `useGridLayout`)

## Column Resizing Example

```jsx
const Styles = styled.div`
  padding: 1rem;

  .table {
    display: inline-block;
    border-spacing: 0;
    border: 1px solid black;

    .th,
    .td {
      margin: 0;
      padding: 0.5rem;
      border-bottom: 1px solid black;
      border-right: 1px solid black;

      ${'' /* One way to deal with borders in the useGridLayout table
              is to use nth-child pseudo-selectors. */}
      :nth-child(6n + 8) {
        border-right: 0;
      }
      :nth-child(6n + 8):nth-last-child(-n + 7) ~ * {
        border-bottom: 0;
      }

      .resizer {
        display: inline-block;
        background: blue;
        width: 10px;
        height: 100%;
        position: absolute;
        right: 0;
        top: 0;
        transform: translateX(50%);
        z-index: 1;
        ${'' /* prevents from scrolling while dragging on touch devices */}
        touch-action:none;

        &.isResizing {
          background: red;
        }
      }
    }
  }
`

function Table({ columns, data }) {
  const defaultColumn = React.useMemo(
    () => ({
      minWidth: 30,
      width: 'auto',
      maxWidth: 400,
    }),
    []
  )

  const {
    getTableProps,
    headerGroups,
    rows,
    prepareRow,
    state,
    resetResizing,
  } = useTable(
    {
      columns,
      data,
      defaultColumn,
    },
    useGridLayout,
    useResizeColumns
  )

  return (
    <>
      <button onClick={resetResizing}>Reset Resizing</button>
      <div>
        <div {...getTableProps()} className="table">
          {headerGroups.map(headerGroup =>
            headerGroup.headers.map(column => (
              <div {...column.getHeaderProps()} className="th">
                {column.render('Header')}
                {/* Use column.getResizerProps to hook up the events correctly */}
                <div
                  {...column.getResizerProps()}
                  className={`resizer ${column.isResizing ? 'isResizing' : ''}`}
                />
              </div>
            ))
          )}

          {rows.map((row, i) => {
            prepareRow(row)
            return row.cells.map(cell => {
              return (
                <div {...cell.getCellProps()} className="td">
                  {cell.render('Cell')}
                </div>
              )
            })
          })}
        </div>
      </div>
      <pre>
        <code>{JSON.stringify(state, null, 2)}</code>
      </pre>
    </>
  )
}
```

---

## Expanding Example

```jsx
const Styles = styled.div`
  padding: 1rem;

  table {
    border-spacing: 0;
    border: 1px solid black;

    ${'' /* One way to deal with borders in the useGridLayout table
              is to use grid-gap and background. */}
    background-color: black;
    grid-gap: 1px;
    > * {
      background-color: white;
    }

    th,
    td {
      margin: 0;
      padding: 0.5rem;
    }
  }
`

function Table({ columns: userColumns, data }) {
  const {
    getTableProps,
    headerGroups,
    rows,
    prepareRow,
    state: { expanded },
  } = useTable(
    {
      columns: userColumns,
      data,
    },
    useExpanded, // Use the useExpanded plugin hook
    useGridLayout
  )

  return (
    <>
      <table {...getTableProps()}>
        {headerGroups.map(headerGroup =>
          headerGroup.headers.map(column => (
            <th {...column.getHeaderProps()}>{column.render('Header')}</th>
          ))
        )}
        {rows.map((row, i) => {
          prepareRow(row)
          return row.cells.map(cell => {
            return <td {...cell.getCellProps()}>{cell.render('Cell')}</td>
          })
        })}
      </table>
      <br />
      <div>Showing the first 20 results of {rows.length} rows</div>
      <pre>
        <code>{JSON.stringify({ expanded: expanded }, null, 2)}</code>
      </pre>
    </>
  )
}
```

---

## Sub-Components Example

```jsx
const Styles = styled.div`
  padding: 1rem;

  .table {
    border-spacing: 0;
    border: 1px solid black;

    ${'' /* One way to deal with borders in the useGridLayout table
              is to use grid-gap and background. */}
    background-color: black;
    grid-gap: 1px;
    > * {
      background-color: white;
    }

    > div {
      margin: 0;
      padding: 0.5rem;
    }
  }
`

// A simple way to support a renderRowSubComponent is to make a render prop
// This is NOT part of the React Table API, it's merely a rendering
// option we are creating for ourselves in our table renderer
function Table({ columns: userColumns, data, renderRowSubComponent }) {
  const {
    getTableProps,
    headerGroups,
    rows,
    prepareRow,
    state: { expanded },
  } = useTable(
    {
      columns: userColumns,
      data,
    },
    useExpanded, // We can useExpanded to track the expanded state
    // for sub components too!
    useGridLayout
  )

  return (
    <>
      <pre>
        <code>{JSON.stringify({ expanded: expanded }, null, 2)}</code>
      </pre>
      <div className="table" {...getTableProps()}>
        {headerGroups.map(headerGroup =>
          headerGroup.headers.map(column => (
            <div {...column.getHeaderProps()}>
              {column.render('Header')}
            </div>
          ))
        )}
        {rows.map((row, i) => {
          prepareRow(row)
          return (
            <>
              {row.cells.map(cell => {
                return <div {...cell.getCellProps()}>{cell.render('Cell')}</div>
              })}
              {/*
                If the row is in an expanded state, render a row with a
                column that fills the entire length of the table.
              */}
              {row.isExpanded ? (
                <div {...row.getRowProps()}>
                  {/*
                    Inside it, call our renderRowSubComponent function. In reality,
                    you could pass whatever you want as props to
                    a component like this, including the entire
                    table instance. But for this example, we'll just
                    pass the row
                  */}
                  {renderRowSubComponent({ row })}
                </div>
              ) : null}
            </>
          )
        })}
      </div>
      <br />
      <div>Showing the first 20 results of {rows.length} rows</div>
    </>
  )
}
```

tannerlinsley:
Was this closed by mistake?

CollinsSpencer:
@tannerlinsley This PR is for v7 and makes `useGridLayout` production ready, with all the same features as the other layout hooks, where header groups display correctly and column resizing works.

It's unclear to me if updates will still be made to v7 going forward, but I can open a new PR to the v7 branch if you think this is worthwhile.

tannerlinsley:
Reopened

PR #3633, opened 2022-01-12
fix(useGroupBy): set depth on leaf rows when grouping

### LINKED ISSUE #3631: row.depth is 0 for leaf rows while using useGroupBy
While using `useGroupBy`, `row.depth` is always `0` for leaf rows in groupings. If multiple groupings are done, grouped rows below the top level **do** have the correct `depth` set.

https://codesandbox.io/s/elegant-poincare-h9suj?file=/src/App.js

1. Click one or more header cell groupBy toggle icon
2. Expand groupings until a set of leaf rows is visible
3. See that the cells' backgrounds are blue (`depth === 0`)

**Expected behavior (Recommended)**
`row.depth` should be parent row's depth + 1 for leaf nodes.

**Screenshots**
Blue rows are `depth === 0`, white rows are anything else.
image

**Desktop (please complete the following information):**

- OS: Windows 10
- Browser: Chrome
- Version: 96

tyrdavis:
Add the computed depth to leaf rows returned from groupUpRecursively

fixes #3631

### DIFF
--- a/src/plugin-hooks/tests/useGroupBy.test.js
+++ b/src/plugin-hooks/tests/useGroupBy.test.js
@@ -85,6 +85,7 @@ function Table({ columns, data }) {
       <thead>
         {headerGroups.map(headerGroup => (
           <tr {...headerGroup.getHeaderGroupProps()}>
+            <th>Depth</th>
             {headerGroup.headers.map(column => (
               <th {...column.getHeaderProps()}>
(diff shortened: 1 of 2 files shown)


penx:
@tannerlinsley this would be really useful for our use case (indenting sub rows) and as far as I can see, it is not a breaking change. Any chance of reopening?

PR #3773, opened 2022-03-24
Types reflect that row.original is never undefined

david-crespo:
As far as I can tell this is ok? Then you don't have to do null checks in calling code that wants to use `row.original`.


david-crespo:
Ah, ok. This doesn't work because a grouped row has no `original`. That's interesting! It would be cool to work around this somehow, but it seems like a lot of work. Curious for your thoughts, but I'll close this.

https://github.com/TanStack/react-table/blob/192c4841e87d5ff4950f3a0a612d22912905cff4/packages/react-table/src/utils/groupRowsFn.ts#L126

tannerlinsley:
We would need to create a new row type variant for grouped rows. It does introduce some complexity, but not impossible.

PR #3783, opened 2022-03-29
V6

hatim85f:
Please more the docs more clear for table expansion


tannerlinsley:
This PR needs to be redone.
- Do not include built files
- Only include changed files

PR #3798, opened 2022-04-05
feat: implement solidjs adapter for v8 alpha

Mokshit06:
This PR implements a solidjs adapter for `react-table` v8 alpha as discussed in this twitter thread https://twitter.com/tannerlinsley/status/1511336168008212481.


davedbase:
Would love to see the fork submitted to SolidHack (https://hack.solidjs.com read the submission guidelines). A small demo site would help with retaining votes for your submission!

izznatsir:
I'm sorry @Mokshit06, I don't see this PR before submitting my own PR #3801. But, my PR focus first on modifying the core to work well with Solid signal reactivity primitive. Maybe we can collaborate?

Mokshit06:
Thanks @izznatsir, I overlooked that we need to keep the options also reactive, but do we really need to modify the core to work well with solid? Would helpers like `mergeProps` from `solid-js` help with retaining reactivity for options?

Mokshit06:
I've pushed another commit that adds a `mergeOptions` function arg to the options, which was one of the things breaking reactivity, because initially it was doing `Object.assign` which would call all the getters eagerly. Now the adapter can specify how it wants to merge the options, and the solid adapter creates a proxy with `mergeProps` which lazily searches for keys from the objects passed.

I think that adding accessor API to the core might convolute the API since now only a few properties are functions and others are primitives. Rather we can just tell the user to use getters wherever they want to make the property reactive like:

```js
const instance = createTable(table, {
    get data() {
        return data()
    },
    columns: columns()
})
```

Here the data is a lazy getter and will be reactive, whereas columns would not get updated.

It also updates uses of `event.persist()` to first check it exists because its only available in react.

izznatsir:
Wow, that's cool! Definitely much cleaner approach. Let me play around with it to see if there is something missed.

packages/table-core/src/types.ts
```diff
@@ -133,7 +133,12 @@ export type Options<TGenerics extends AnyGenerics> = CoreOptions<TGenerics> &
   ExpandedOptions<TGenerics> &
   ColumnSizingOptions &
   PaginationOptions<TGenerics> &
-  RowSelectionOptions<TGenerics>
+  RowSelectionOptions<TGenerics> & {
+    mergeOptions?: (
+      defaultOptions: TableFeature,
+      options: Partial<Options<TGenerics>>
+    ) => Options<TGenerics>
+  }
```
tannerlinsley, line 141: This should go in the `CoreOptions` type

tannerlinsley (changes requested):

tannerlinsley:
I've had to comment out the `/** @jsxImportSource solid-js */` for now because it's causing some compilation issues on the build:

```
➜  react-table (alpha) ✗ yarn build
yarn run v1.22.11
$ rollup --config rollup.config.js && tsc -b

/Users/tannerlinsley/GitHub/react-table/packages/table-core/src/index.tsx → packages/table-core/build/esm...
created packages/table-core/build/esm in 3.3s

/Users/tannerlinsley/GitHub/react-table/packages/table-core/src/index.tsx → packages/table-core/build/cjs...
created packages/table-core/build/cjs in 3.1s

/Users/tannerlinsley/GitHub/react-table/packages/table-core/src/index.tsx → packages/table-core/build/umd/index.development.js...
created packages/table-core/build/umd/index.development.js in 3.9s

/Users/tannerlinsley/GitHub/react-table/packages/table-core/src/index.tsx → packages/table-core/build/umd/index.production.js...
 index.production.js ⏤  14.8 kB
created packages/table-core/build/umd/index.production.js in 6.5s

/Users/tannerlinsley/GitHub/react-table/packages/react-table/src/index.tsx → packages/react-table/build/esm...
created packages/react-table/build/esm in 317ms

/Users/tannerlinsley/GitHub/react-table/packages/react-table/src/index.tsx → packages/react-table/build/cjs...
created packages/react-table/build/cjs in 321ms

/Users/tannerlinsley/GitHub/react-table/packages/react-table/src/index.tsx → packages/react-table/build/umd/index.development.js...
created packages/react-table/build/umd/index.development.js in 294ms

/Users/tannerlinsley/GitHub/react-table/packages/react-table/src/index.tsx → packages/react-table/build/umd/index.production.js...
 index.production.js ⏤  995 B
created packages/react-table/build/umd/index.production.js in 2.2s

/Users/tannerlinsley/GitHub/react-table/packages/solid-table/src/index.tsx → packages/solid-table/build/esm...
[!] (plugin babel) SyntaxError: /Users/tannerlinsley/GitHub/react-table/packages/solid-table/src/index.tsx: importSource cannot be set when runtime is classic.
> 1 | /** @jsxImportSource solid-js */
    | ^
  2 | import {
  3 |   AnyGenerics,
  4 |   CreateTableFactoryOptions,
packages/solid-table/src/index.tsx
SyntaxError: /Users/tannerlinsley/GitHub/react-table/packages/solid-table/src/index.tsx: importSource cannot be set when runtime is classic.
> 1 | /** @jsxImportSource solid-js */
    | ^
  2 | import {
  3 |   AnyGenerics,
  4 |   CreateTableFactoryOptions,
    at File.buildCodeFrameError (/Users/tannerlinsley/GitHub/react-table/node_modules/@babel/core/lib/transformation/file/file.js:249:12)
    at NodePath.buildCodeFrameError (/Users/tannerlinsley/GitHub/react-table/node_modules/@babel/traverse/lib/path/index.js:139:21)
    at PluginPass.enter (/Users/tannerlinsley/GitHub/react-table/node_modules/@babel/plugin-transform-react-jsx/lib/create-plugin.js:143:28)
    at newFn (/Users/tannerlinsley/GitHub/react-table/node_modules/@babel/traverse/lib/visitors.js:177:21)
    at NodePath._call (/Users/tannerlinsley/GitHub/react-table/node_modules/@babel/traverse/lib/path/context.js:53:20)
    at NodePath.call (/Users/tannerlinsley/GitHub/react-table/node_modules/@babel/traverse/lib/path/context.js:40:17)
    at NodePath.visit (/Users/tannerlinsley/GitHub/react-table/node_modules/@babel/traverse/lib/path/context.js:100:31)
    at TraversalContext.visitQueue (/Users/tannerlinsley/GitHub/react-table/node_modules/@babel/traverse/lib/context.js:103:16)
    at TraversalContext.visitSingle (/Users/tannerlinsley/GitHub/react-table/node_modules/@babel/traverse/lib/context.js:77:19)
    at TraversalContext.visit (/Users/tannerlinsley/GitHub/react-table/node_modules/@babel/traverse/lib/context.js:131:19)

error Command failed with exit code 1.
info Visit https://yarnpkg.com/en/docs/cli/run for documentation about this command.
```

I'll need some help fixing those issues (which I am very unfamiliar with) before we can uncomment that line.

Mokshit06:
We would need solid's babel plugin to compile its jsx to js, since it isn't using the standard jsx transform. But I think since we're just having one jsx element inside the render function, we might as well change it to `createComponent` call, which is what it would get transformed into

Mokshit06:
Yep, babel is currently transpiling it to the following, which wouldn't work:
```js
function render(Comp, props) {
  if (!Comp) return null;

  if (typeof Comp === 'function') {
    return /*#__PURE__*/React.createElement(Comp, props);
  }

  return Comp;
}
```

Mokshit06:
Should I create a new PR for fixing this issue?

Mokshit06:
#3840 fixes this

PR #3799, opened 2022-04-05
add getIsSomeRowsExpanded instance property (v8 alpha)

KevinVandy:
A small and efficient addition to the expanded feature I use in my project.

In most cases where expanded state is not handled manually, this should be an instant constant lookup time.

### DIFF
--- a/packages/table-core/src/features/Expanding.ts
+++ b/packages/table-core/src/features/Expanding.ts
@@ -62,6 +62,7 @@ export type ExpandedInstance<TGenerics extends PartialGenerics> = {
   getToggleAllRowsExpandedProps: <TGetter extends Getter<ToggleExpandedProps>>(
     userProps?: TGetter
   ) => undefined | PropGetterValue<ToggleExpandedProps, TGetter>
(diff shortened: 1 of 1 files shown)


tannerlinsley:
I can’t remember does the row api have this already?

KevinVandy:
@tannerlinsley 

My original code I was going to pr here looked just like the `getIsAllRowsExpanded` code, kinda like this:

```ts
getIsSomeRowsExpanded: () => {
  const expanded = instance.getState().expanded
  
  // If expanded is true, save some cycles and return true
  if (expanded === true) {
    return true
  }

  // If any row is not expanded, return false
  if (
    Object.keys(instance.getRowModel().rowsById).some(
      id => instance.getIsRowExpanded(id)
    )
  ) {
    return true
  }

  // None must be expanded :shrug:
  return false
},
```

but then I thought I could skip the loop by just looking at the expanded state instead.

The use case for this may be small, as I'm just using it for an indeterminate value, similar to `getIsSomeRowsSelected`

PR #3801, opened 2022-04-06
Feat: Modify core for SolidJS

izznatsir:
Solid will only re-render the DOM if it detect `signal` values that being used via their accessor functions in the DOM are changed via their corresponding setter functions.

Options that being changed are options that intended to possibly be controlled by the users such as `data`, `column`, and `state`. By allowing them to be accessor functions, any instance method that calls accessor functions inside it will also be treated as an accessor function. Therefore, those three options can be correctly controlled by the user.

Without this change, user has no way to change `data`, `column`, or `state` directly (via signal setter function) in runtime (changes after initialization) that should have been re-render the table.

Options that not intended to be directly controllable by the user at runtime are not changed to accept accessor function.

The tests hasn't been written yet. But, you can manually test by running the examples that being prefixed with `solid-` in `solid-table` branch of my `react-table` repository fork. Here is the [repository branch](https://github.com/izznatsir/react-table/tree/solid-table). The simplest way to try it is to run that repo branch in Github Codespace.

Already working examples:
- [x] `basic`
- [x] `column-ordering`
- [x] `column-pinning`
- [x] `column-sizing`
- [x] `column-visibility`
- [x] `expanding`
- [x] `filters`
- [x] `fully-controlled`
- [ ] `grouping`
- [x] `pagination`
- [x] `row-selection`
- [x] `sorting`

Files that need to be modified for examples to work:
- [x] `core.tsx`
- [x] `features/ColumnSizing.ts`
- [x] `features/Expanding.ts`
- [x] ~~`features/Filters.ts`~~ (no change needed)
- [ ] `features/Grouping.ts`
- [ ] `features/Headers.ts`
- [x] ~~`features/Ordering.ts`~~ (no change needed)
- [x] ~~`features/Pagination.ts`~~ (no change needed)
- [x] ~~`features/Pinning.ts`~~ (no change needed)
- [x] ~~`features/RowSelection.ts`~~ (no change needed)
- [x] `features/Sorting.ts`
- [x] ~~`features/Visibility.ts`~~ (no change needed)


izznatsir:
`e.persist` call is only available in React that comes from `SyntheticEvent`. So, I checked it first before calling it.

izznatsir:
Closed in favor of #3798 and a separate PR that focus on removing or isolating React only types and logic  from the core.

PR #3829, opened 2022-04-13
fix(getCanNextPage): Use the `getPageCount` function over manual calculation on the fly

colindjk:
This update will ensure that the `getCanNextPage` function works correctly when doing server side pagination (or some other form of pagination handled outside of react-tables internals).

One thing to not is while I've manually tested this, my changes didn't trigger / get picked up by any of the tests when running `yarn dev`. I could add one but that could take up some time to figure out, is there an outline on how to quickly add a test for v8? I see some of the current tests are still from the old (v7).


tannerlinsley:
The testing story is still a bit shaky, but you could duplicate the `Visibility.test.tsx` file to a new `Pagination.test.tsx` file and go from there.

PR #3833, opened 2022-04-17
import performance module for node < v16 until platforms like Vercel support node 16

KevinVandy:
Vercel builds throw an error `ReferenceError: performance is not defined` because Vercel supports a max node version of node v14 still, where the performance module is not yet globally defined.

Stack Overflow: https://stackoverflow.com/questions/46436943/referenceerror-performance-is-not-defined-when-using-performance-now

### DIFF
--- a/packages/table-core/src/utils.tsx
+++ b/packages/table-core/src/utils.tsx
@@ -1,3 +1,4 @@
+import { performance } from 'perf_hooks' //needed until node 16 is more universal
 import { Getter, NoInfer, PropGetterValue, TableState, Updater } from './types'
 
 export type IsAny<T> = 0 extends 1 & T ? true : false


KevinVandy:
Unfortunately, I had a misconception for how that import worked. May need to revert.

```
Adding to empty esbuild project✘ [ERROR] Could not resolve "perf_hooks"

    node_modules/@tanstack/react-table/node_modules/@tanstack/table-core/build/cjs/utils.js:16:25:
      16 │ var perf_hooks = require('perf_hooks');
         ╵                          ~~~~~~~~~~~~

  The package "perf_hooks" wasn't found on the file system but is built
  into node. Are you trying to bundle for node? You can use "platform:
  'node'" to do that, which will remove this error.

✖ Adding to empty esbuild project
 ERROR  Error: Build failed with 1 error:
```

PR #3838, opened 2022-04-18
fix(examples): tidy up unused code in examples

shiftkey:
Follow-up to #3835

As part of reading the current samples I noticed some unused code that distracted from the rest of the code.

Again, happy to drop commits or refine work if things don't align with your direction for these.


tannerlinsley:
This needs to be unconflicted. Let me know if you want to reopen.

shiftkey:
@tannerlinsley all good, I'll open a fresh one when I next have some spare time

PR #3843, opened 2022-04-19
fix: remove react from core

izznatsir:
Remove `react` from `table-core` to make it truly framework agnostic in terms of event types and event handling.

### Testing
All tests pass. The examples also run without problem.


tannerlinsley:
I worry about removing some of those persist calls. We need the data that gets persisted to do a lot of internal things that wouldn't be there if we didn't persiste them.

izznatsir:
The persist calls are not removed, it simply removed from the type and only call it when it exist.

tannerlinsley:
Gotcha. Well, if you can fixup the conflicts, I'll merge. :)

izznatsir:
Conflicts resolved.

PR #3845, opened 2022-04-19
access data using nested accessor keys

caravinci:
adds a function to recursively dig values out of nested data objects

### DIFF
--- a/packages/table-core/src/core.tsx
+++ b/packages/table-core/src/core.tsx
@@ -4,6 +4,7 @@ import {
   propGetter,
   memo,
   RequiredKeys,
+  getProp,
 } from './utils'
 
 import {
@@ -331,7 +332,7 @@ export function createTableInstance<TGenerics extends AnyGenerics>(
         accessorFn = columnDef.accessorFn
       } else if (columnDef.accessorKey) {
         accessorFn = (originalRow?: TGenerics['Row']) =>
-          (originalRow as any)[columnDef.accessorKey]
+          getProp(originalRow as any, columnDef.accessorKey)
       }
 
       if (!id) {
--- a/packages/table-core/src/utils.tsx
+++ b/packages/table-core/src/utils.tsx
@@ -163,6 +163,27 @@ export function memo<TDeps extends readonly any[], TResult>(
   }
 }
 
+export function getProp(
+  dataObj: Record<string, any>,
+  accessorKeys: string | string[] = '',
+  defaultVal = ''
+): string {
+  const keys = Array.isArray(accessorKeys)
+    ? accessorKeys
+    : accessorKeys
+        .replace(/(\[(\d)\])/g, '.$2')
+        .split('.')
+        .filter(key => !['', undefined].includes(key))
+
+  const value = dataObj[keys[0]]
+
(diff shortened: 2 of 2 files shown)


packages/table-core/src/utils.tsx
```diff
@@ -163,6 +163,27 @@ export function memo<TDeps extends readonly any[], TResult>(
   }
 }
 
+export function getProp(
+  dataObj: Record<string, any>,
+  accessorKeys: string | string[] = '',
+  defaultVal = ''
+): string {
```
caravinci, line 170: meant to replicate (at a rather basic level) lodash's `get`

packages/table-core/src/utils.tsx
```diff
@@ -163,6 +163,27 @@ export function memo<TDeps extends readonly any[], TResult>(
...
+  const keys = Array.isArray(accessorKeys)
+    ? accessorKeys
+    : accessorKeys
+        .replace(/(\[(\d)\])/g, '.$2')
```
caravinci, line 174: this allows `accessor.[0].accessor' type of keys... WIP

packages/table-core/src/utils.tsx
```diff
@@ -163,6 +163,27 @@ export function memo<TDeps extends readonly any[], TResult>(
...
+        .split('.')
+        .filter(key => !['', undefined].includes(key))
+
+  const value = dataObj[keys[0]]
+
+  if (value && keys.length > 1) {
+    return getProp(value, keys.slice(1), defaultVal)
```
caravinci, line 181: recursively keeps on digging until it find the innermost property

packages/table-core/src/utils.tsx
```diff
@@ -163,6 +163,27 @@ export function memo<TDeps extends readonly any[], TResult>(
...
+  }
+
+  return value ?? defaultVal
```
caravinci, line 184: returning an empty string if nothing is found.

packages/table-core/src/core.tsx
```diff
@@ -331,7 +332,7 @@ export function createTableInstance<TGenerics extends AnyGenerics>(
         accessorFn = columnDef.accessorFn
       } else if (columnDef.accessorKey) {
         accessorFn = (originalRow?: TGenerics['Row']) =>
-          (originalRow as any)[columnDef.accessorKey]
+          getProp(originalRow as any, columnDef.accessorKey)
```
caravinci, line 335: get the nested property, rather than expecting data like `{ 'accessor.accessor': value }`

caravinci (commented):
couple of comments to explain intention, and facilitate feedback

tannerlinsley:
This would need full typesafety/autocomplete before I’ll consider merging. This means I would want to see “foo.bar.baz” auto complete.

caravinci:
Absolutely fair! that'd be great, and I'd love to help... but I'm not quite sure how to implement something like that. Closing this so others can take a shot at it.

Thanks again for this amazing library!

orenklein:
@caravinci You can take a look at this https://stackoverflow.com/a/58436959/15026946 for the needed type.
Not sure how it might affect performance though.

PR #3848, opened 2022-04-20
feat: implement vue adapter

Mokshit06:
This PR implements a vue adapter for react-table v8 alpha. It needs a bit more testing, but from the few examples I tested it with, it seemed to function properly


package.json
```diff
@@ -2,7 +2,7 @@
   "private": true,
   "repository": "https://github.com/tanstack/react-table.git",
   "scripts": {
-    "linkAll": "lerna exec 'yarn link' && lerna exec 'yarn link @tanstack/react-table' && lerna exec 'yarn link @tanstack/solid-table'",
+    "linkAll": "lerna exec 'yarn link' && lerna exec 'yarn link @tanstack/react-table' && lerna exec 'yarn link @tanstack/solid-table'  && lerna exec 'yarn link @tanstack/vue-table'",
```
tannerlinsley, line 5: These extra links shouldn’t be necessary. Only the core. Right?

packages/vue-table/package.json
```diff
@@ -0,0 +1,42 @@
+{
+  "name": "@tanstack/vue-table",
+  "author": "Tanner Linsley",
+  "version": "8.0.0-alpha.31",
+  "description": "Helpers for building lightweight, fast and extendable datagrids for Solid",
```
tannerlinsley, line 5: Change to vue

tannerlinsley (changes requested):

package.json
Mokshit06: I added them because the `@tanstack/react-table` existed before I opened the solid-table PR, so I thought that other adapters should also be added, but do we need to really do `yarn link` with workspaces at all?

tannerlinsley: You’re not supposed to have to, but for whatever reason, I found it necessary for the core.

Mokshit06: So should I remove all these packages and add `@tanstack/table-core` instead?

tannerlinsley: Yeah, let's do that

PR #3862, opened 2022-04-27
fix ssr warning due to `useLayoutEffect`

Mokshit06:
The react adapter was calling `useLayoutEffect` which showed a warning during SSR. This PR makes the `useLayoutEffect` isomorphic and switches it for `useEffect` on server


tannerlinsley:
Excellent :)

PR #3864, opened 2022-04-28
Include example projects in yarn & tsc -b builds

JakeGinnivan:
WIP PR to see if you would like this fixed.

Currently there are a number of compilation errors with the example projects, I started compiling them because we have some performance issues around the types of the react-table alpha which I wanted to investigate.

@tannerlinsley would you be interested in landing this, if so I can fix the compilation errors.

### DIFF
--- a/examples/basic/package.json
+++ b/examples/basic/package.json
@@ -1,5 +1,6 @@
 {
-  "name": "dynamic",
+  "name": "basic",
+  "version": "0.0.0",
   "private": true,
   "scripts": {
     "dev": "vite",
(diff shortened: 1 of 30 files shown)


tannerlinsley:
I’m interested to hear more of what this is doing under the hood.

JakeGinnivan:
Sure, 

Few bits, I have included all the example projects in the TypeScript composite project. This essentially means that when you run `tsc -b` the examples will be compiled at the same time.

Examples also have a few extra dependencies, so I have included the projects in the yarn workspace so libraries like faker which is required by some of the examples projects. The example projects will not compile unless those dependencies are installed.

Yarn also needs each package.json name needs to be unique and have a version, so I updated each example's package.json name from `dynamic` to the same as the folder and added a version.

You should be able to see the compilation errors in the examples by just running `yarn` and then `yarn tsc -b`

tannerlinsley:
I went ahead and added some of this locally, but with some other changes to make it all work. Thanks for the inspo!

PR #3871, opened 2022-04-30
row-selection initial docs

KevinVandy:
Left some TODOs for some row model methods that I believe Tanner would be able to explain and document much better than I could, but otherwise, filled out ~90% of initial row-selection docs


KevinVandy:
Guess this PR is slightly out of date now with the latest refactor. I can update it sometime this week.

tannerlinsley:
Better to get it in that let it get even more out of date. WIP is fine so I'm going to merge. We can touch it up later as we go.

PR #3872, opened 2022-05-01
feat: implement svelte adapter

Mokshit06:
This PR implements a svelte adapter for react-table. The API isn't finalized yet, especially about how changes to options should be tracked. This is how it currently looks:-

```svelte
<script>
import { createTableInstance } from '@tanstack/svelte-table';
import { writable } from 'svelte/store';

let options = writable({
  data,
  columns
})
let instance = createTableInstance(table, options)
</script>

<button on:click={() => {
  options.update(o => ({ ...o, data: newData() }))
}}>
  Refresh data
</button>
<table {...$instance.getTableProps()}>
  <!-- rest of the table -->
</table>
````

`options` doesn't need to be a store if it isn't getting updated. It can just be a normal js object if it is static.

The update method seems to work fine for simple use cases like updating data etc, but might be hard for nested updates like updating the table state

```js
options.update(o => ({
  ...o,
  state: {
    ...o.state,
    sorting
  }
}))
```

We could export an immer-like `produce` method to make these easier, or a helper to create proxy-based observables that can be subscribed to.


tdung (approved):

tdung (commented):
Y

Issue #3847, opened 2022-04-20
Suppor for React 18


christianhultin:
### Describe the bug

Peer dependencies complain when installing `react@18.0.0` while having `react-table@7.7.0` in the project. I know there is a new major version coming (v8), but could it be an idea to allow v7 to support React 18 still to avoid having to migrate react-table to a new major version when upgrading react?

```
npm ERR! While resolving: react-table@7.7.0
npm ERR! Found: react@18.0.0
npm ERR! node_modules/react
npm ERR!   react@"18.0.0" from the root project
npm ERR!   peer react@"^16.11.0 || ^17 || ^18" from @auth0/auth0-react@1.10.0
npm ERR!   node_modules/@auth0/auth0-react
npm ERR!     @auth0/auth0-react@"^1.10.0" from the root project
npm ERR!   26 more (@emotion/react, @emotion/styled, @mui/icons-material, ...)
npm ERR!
npm ERR! Could not resolve dependency:
npm ERR! peer react@"^16.8.3 || ^17.0.0-0" from react-table@7.7.0
npm ERR! node_modules/react-table
npm ERR!   react-table@"^7.7.0" from the root project
npm ERR!
npm ERR! Conflicting peer dependency: react@17.0.2
...
```

Thanks for a great project!

### Your minimal, reproducible example

...

### Steps to reproduce

1. Create a new project and install `react-table@7.7.0`
2. Try to install react-18 through `npm install react@18.0.0 react-dom@18.0.0 --save`

### Expected behavior

Being able to install react 18

### How often does this bug happen?

Every time

### Platform

macOs

### react-table version

v7.7.0

### TypeScript version

4.6.3

### Terms & Code of Conduct

- [X] I agree to follow this project's Code of Conduct
- [X] I understand that if my bug cannot be reliable reproduced in a debuggable environment, it will probably not be fixed and this issue may even be closed.

tannerlinsley:
I'll try and do this next publish I make to v7.

trumbitta:
> I'll try and do this next publish I make to v7.

Thank you @tannerlinsley , this will help us unblock updating our apps to React 18 while we wait for v8 🙏

PR #3873, opened 2022-05-01
chore(deps): allow React 18

chentsulin:
Allow React 18 on v7 and test against it.


stormsilver:
Do we think this is good enough to publish a patch release for? Would be quite helpful for use on new projects that use React 18.

pSnehanshu:
Has this been published yet? I am facing the same issue #3847.

chentsulin:
Not yet.

PR #3892, opened 2022-05-05
resetState options to reset to default state

KevinVandy:


KevinVandy:
Currently having problems getting the examples to run locally with `yarn link`, leaving in draft for a moment

tannerlinsley:
I see some renaming to clearX. I thought were going keep the reset naming

KevinVandy:
> I see some renaming to clearX. I thought were going keep the reset naming

Assuming you're talking about `clearSorting(columnId?: string)`, I separated that into its own method since already had its own special param and functionality. 

`resetSorting()` can only be concerned with resetting all sorting state to initial/default (consistent with all other state resets)

and `clearSorting(columnId?: string)` can be concerned with removing sorting from 1 column, though the instance method might actually be a bit redundant. Could only keep something `column.clearSort()`

KevinVandy:
Removed `instance.clearSorting()`, but kept `column.clearSorting()` since that contained unique functionality

tannerlinsley:
You have me thinking now…. Should we remove all column specific methods from the instance?

KevinVandy:
I think it made sense here since there was a redundant method on the instance, as the same thing could be accomplished with `instance.resetSorting(true)`, but I think in general its nice to have the entire api available from just the instance for discoverability. ¯\_(ツ)_/¯

KevinVandy:
I'll open PR in case you want to do something with it, but I'll also try to fix my local workspace tonight so I can do better testing

PR #3894, opened 2022-05-06
v8 feat - add resetState default state options

KevinVandy:
resetState methods now accept an optional "defaultState" boolean that, when true, will ignore initialState, and instead be set to the empty default state.

Better name than "defaultState" for the boolean?


KevinVandy:
This was redone from a new fork.

Docs will need a new pass after this and other changes, which I can handle this weekend too.

PR #3904, opened 2022-05-10
v8 examples - create basic svelte example

KevinVandy:
added a basic example for svelte

updated publish script to loop through all framework examples


examples/svelte/basic/src/App.svelte
```diff
@@ -0,0 +1,143 @@
...
+
+  const rerender = () => {
+    data = data
+  }
+
+  $: instance = createTableInstance(table, {
+    data,
+    columns,
+    getCoreRowModel: getCoreRowModelSync(),
+  })
```
Mokshit06, line 104: `instance` should be a normal variable and not a dependent one. Making it dependent would cause react-table to create a new instance each time any of the options like data and columns update which would cause it to lose its internal state. To make the options reactive you'll have to make them a svelte store

examples/svelte/basic/src/App.svelte
```diff
@@ -0,0 +1,143 @@
...
+</script>
+
+<div class="p-2">
+  <table>
+    <thead>
+      {#each $instance.getHeaderGroups() as headerGroup}
+        <tr>
+          {#each headerGroup.headers as header}
+            <th colSpan={header.colSpan}>
+              {header.isPlaceholder ? null : header.renderHeader()}
```
Mokshit06, line 114: This should instead be
```svelte
{#if !header.isPlaceholder}
  <svelte:component this={header.renderHeader()} />
{/if}
```
because `renderHeader()` might be rendering a custom component, so in the svelte adapter the `render` method returns the custom components with props passed to them and the plain values as a dummy component

examples/svelte/basic/src/App.svelte
```diff
@@ -0,0 +1,143 @@
...
+            </th>
+          {/each}
+        </tr>
+      {/each}
+    </thead>
+    <tbody>
+      {#each $instance.getRowModel().rows as row}
+        <tr>
+          {#each row.getVisibleCells() as cell}
+            <td>{cell.renderCell()}</td>
```
Mokshit06, line 124: Same here, this should be
```svelte
<td><svelte:component this={cell.renderCell()} /></td>
```

examples/svelte/basic/src/App.svelte
```diff
@@ -0,0 +1,143 @@
...
+          {/each}
+        </tr>
+      {/each}
+    </tbody>
+    <tfoot>
+      {#each $instance.getFooterGroups() as footerGroup}
+        <tr>
+          {#each footerGroup.headers as header}
+            <th colSpan={header.colSpan}>
+              {header.isPlaceholder ? null : header.renderFooter()}
```
Mokshit06, line 134: And the footer should be
```svelte
{#if !header.isPlaceholder}
  <svelte:component this={header.renderFooter} />
{/if}
```

examples/svelte/basic/src/App.svelte
KevinVandy: Thanks for your help.

though I get this error with that code

```Uncaught TypeError: Class constructor SvelteComponent cannot be invoked without 'new'```

Can't tell if its a config issue or not

Comment (Mokshit06) [reply examples/svelte/basic/src/App.svelte:114]: I think that this issue is caused due to transpiling the classes in svelte-table to functions in the bundle here:
```js
// source code
class WrapperComponent extends SvelteComponent {
   // rest of code here
}

// in bundle
function WrapperComp(options) {
  var _this;

  _this = _SvelteComponent.call(this) || this;
```
We'll need to update the rollup/babel config of svelte-table to fix it.

Mokshit06: #3907

examples/svelte/basic/src/App.svelte
```diff
@@ -0,0 +1,148 @@
...
+
+  const rerender = () => {
+    data = data
+  }
+
+  const instance = createTableInstance(table, {
+    data: $data,
+    columns: $columns,
+    getCoreRowModel: getCoreRowModel(),
+  })
```
Mokshit06, line 105: I'm curious as to was this example rerendering for you on clicking the button? This was how the API was supposed to work for updatable options:-

```js
const options = writable({
  data: defaultData,
  columns: defaultColumns,
  getCoreRowModel: getCoreRowModel()
});

const rerender = () => {
  options.update(options => ({
    ...options,
    data: defaultData
  }));
}

const instance = createTableInstance(table, options);
```

KevinVandy: yep, that makes more sense. Will change to that

KevinVandy: after upgrading to the new alpha, I get this error

```Uncaught TypeError: switch_value is not a constructor```

Transpile error again?

Comment (Mokshit06) [reply examples/svelte/basic/src/App.svelte]: `renderCell` and `renderFooter` should be `renderCell()` and `renderFooter` in the template. I think that should make it work

Comment (KevinVandy) [reply examples/svelte/basic/src/App.svelte]: silly mistake on my part, fixed, but same error. :(

Comment (Mokshit06) [reply examples/svelte/basic/src/App.svelte]: This was a bug in the adapter that didn't allow cases where a function was passed as header rather than the string/number literal. I'll open a pr to fix it

File: examples/svelte/basic/src/App.svelte [hunk 6]
```diff
@@ -104,7 +104,7 @@
     }))
   }
 
-  const instance = createTableInstance(table, options)
+  const instance = createTableInstance(table, $options)
```
Mokshit06, line 107: In this case, it should be the store (`options`) that is passed to `createTableInstance` rather than its value (`$options`), because that way the adapter can internally call `subscribe` on the store to listen to changes.

examples/svelte/basic/src/App.svelte
KevinVandy: Thanks for working through this this week.

examples/svelte/basic/src/App.svelte
KevinVandy: ah ok, wasn't clear on that. interesting

examples/svelte/basic/src/App.svelte
Mokshit06: #3917

KevinVandy: It is working identical to the react basic example now

PR #3948, opened 2022-05-23
Fix ordering example

rodolphoasb:
Currently, when a user tries to execute this example it breaks because of a missing dependency. This PR fixes this!

@tannerlinsley, thank you for your working on this! Loved the v8 API... 

Also, if you could take a look at this discussion [here](https://github.com/TanStack/react-table/discussions/3947)... Possibly an error, if not, it will help us improve the docs!


KevinVandy:
to avoid merge conflicts, handled in my branch, but gave credit in commit message to you. Closing

PR #3956, opened 2022-05-23
Add getNextSortingOrder to sorting column api

mr-bjerre:
Add getNextSortingOrder to resolve the discussion in #3944

### DIFF
--- a/packages/table-core/src/features/Sorting.ts
+++ b/packages/table-core/src/features/Sorting.ts
@@ -58,7 +58,7 @@ export type SortingColumn<TGenerics extends TableGenerics> = {
   getAutoSortingFn: () => SortingFn<TGenerics>
   getAutoSortDir: () => SortDirection
   getSortingFn: () => SortingFn<TGenerics>
-  getSortDescFirst: () => boolean
+  getNextSortingOrder: () => SortDirection | 'none'
   getCanSort: () => boolean
   getCanMultiSort: () => boolean
   getSortIndex: () => number
@@ -220,7 +220,10 @@ export const Sorting: TableFeature = {
             }
           }
 
-          const sortDescFirst = column.getSortDescFirst()
+          const sortDescFirst =
+            column.columnDef.sortDescFirst ??
+            instance.options.sortDescFirst ??
+            column.getAutoSortDir() === 'desc'
 
           // Handle toggle states that will remove the sorting
           if (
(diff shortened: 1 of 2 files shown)


mr-bjerre:
I'd say `column.getNextSortingOrder()` would be even better than the more low level method `column.getSortDescFirst()`. The only use case for `column.getSortDescFirst()` is to determine `column.getNextSortingOrder()` AFAIK.

But `toggleSorting` seems pretty complex so I'm not sure how to make `getNextSortingOrder`.

Why this in the first place? You might want to hint the next sorting order on mouse hover. E.g. see https://tailwindui.com/components/application-ui/lists/tables#component-7b5a46e74e475708d966ca31716f1771

packages/table-core/src/features/Sorting.ts
```diff
@@ -278,6 +279,24 @@ export const Sorting: TableFeature = {
...
+        const firstSortDirection = sortDescFirst ? 'desc' : 'asc'
+
+        const isSorted = column.getIsSorted()
+        if (!isSorted) {
+          return firstSortDirection
+        }
+        if (isSorted === firstSortDirection) {
+          return isSorted === 'desc' ? 'asc' : 'desc'
+        } else {
+          return 'none'
```
tannerlinsley, line 296: Let’s make this a false, that way we can check it for truthiness much easier.

tannerlinsley (changes requested):
Any way we can have the toggler logic use this new function?

mr-bjerre:
Yeah I was going in that direction first but then my head started spinning in the toggler logic 😂

I'll revisit again tomorrow (Copenhagen time here so done for the night)

packages/table-core/src/features/Sorting.ts
```diff
@@ -191,6 +192,9 @@ export const Sorting: TableFeature = {
         //   return
         // }
 
+        // this needs to be outside of instance.setSorting to be in sync with rerender
+        const nextSortingOrder = column.getNextSortingOrder()
```
mr-bjerre, line 196: @tannerlinsley I first put this line inside `instance.setSorting` but then the react sorting example broke when `sortAction` became `toggle`. It wasn't clear to me exactly why but I assume some state not being in sync on rerender.

mr-bjerre:
Updated the PR - let me know if it needs more attention

mr-bjerre:
Maybe I should change the method so we return both sortAction and the next sortDirection (if any) instead?

tannerlinsley (approved):

PR #4003, opened 2022-06-07
Feat example vue column pinning

boussadjra:


KevinVandy:
Thanks for making this. I'll merge soon, but for future, also remember to add an entry to the path references in the root `tsconfig.json` file, and add an entry to your examples in the docs/config.json file so that they show up on the TanStack Table docs. I see you are adding to the package.json workspaces correctly now.
Also I might delete the assets and css that are not being used in these examples

KevinVandy:
I might clean up the vue example directories a bit tonight, but looks good.

@tannerlinsley be aware of potential conflicts since looks like prettier was run on some of the json and config files here if you're working on that.

Will merge soon

PR #4031, opened 2022-06-16
Fix pagination bug when pageCount = -1

oeddyo:


oeddyo:
@tannerlinsley not sure where should I add a test for this fix. Could you please let me know?

PR #4051, opened 2022-06-24
examples: add createdAt date column for sorting example

abdullahtariq1171:
This PR adds a date column in sorting example.

### DIFF
--- a/examples/react/sorting/src/main.tsx
+++ b/examples/react/sorting/src/main.tsx
@@ -62,6 +62,9 @@ function App() {
               }),
             ],
           }),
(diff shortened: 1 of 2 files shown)


KevinVandy:
This is good, but can we also update the Solid, Svelte, and Vue examples? If you don't want to, I can do it later

Also, we might want to add a `cell: ({getValue}) => ...` to format the date during render to a shorter, more readable date string that does not take up so much space.

tannerlinsley:
Happy to see this improve as we go, but decided to get it in before it gets left behind.

PR #4064, opened 2022-06-28
adding material ui table example

subbarao:


tannerlinsley:
This will need to be refactored to use the latest changes, API, example patterns etc.

subbarao:
> This will need to be refactored to use the latest changes, API, example patterns etc.

can you release new version of react-table. current version doesn't include flexRender export.
https://github.com/TanStack/table/blob/v8.0.13/packages/react-table/src/index.tsx

tannerlinsley:
Excellent :)

PR #4074, opened 2022-06-29
Fix: RowSelection now respects enableMultiRowSelection

krispya:
Simple PR, if enableMultiRowSelection is false, clear out the object so only one item can be selected at a time.


tannerlinsley:
Feels right to me!

PR #4082, opened 2022-07-02
Adds React full width resizable table example

juhanakristian:
This PR adds a React example similar to the old [full-width resizable table example](https://github.com/TanStack/table/tree/main/old-examples/full-width-resizable-table)

The new example doesn't have the row selection from the old example.


examples/react/full-width-resizable-table/src/main.tsx
```diff
@@ -0,0 +1,158 @@
...
+              </tr>
+            )
+          })}
+        </tbody>
+      </table>
+    </div>
+  )
+}
+
+ReactDOM.render(
```
subbarao, line 153: can you upgrade to react 18.2.0
https://github.com/TanStack/table/blob/main/examples/react/basic/src/main.tsx#L169

examples/react/full-width-resizable-table/yarn.lock
```diff
@@ -0,0 +1,323 @@
+# THIS IS AN AUTOGENERATED FILE. DO NOT EDIT THIS FILE DIRECTLY.
+# yarn lockfile v1
+
+
+"@faker-js/faker@6.3.1":
```
subbarao, line 5: don't add yarn.lock. none of the other examples added yarn.lock.

examples/react/full-width-resizable-table/package.json
```diff
@@ -0,0 +1,22 @@
...
+  "scripts": {
+    "dev": "vite",
+    "build": "vite build",
+    "serve": "vite preview",
+    "start": "vite"
+  },
+  "dependencies": {
+    "@faker-js/faker": "6.3.1",
+    "@tanstack/react-table": "8.0.0",
+    "faker": "^6.6.6",
```
subbarao, line 14: you don't need faker dependency as you already have faker-js.

examples/react/full-width-resizable-table/package.json
```diff
@@ -0,0 +1,22 @@
...
+    "react": "17.0.2",
+    "react-dom": "17.0.2"
+  },
+  "devDependencies": {
+    "@rollup/plugin-replace": "^4.0.0",
```
subbarao, line 19: add faker type as dev dependency.
https://github.com/TanStack/table/blob/main/examples/react/column-ordering/package.json#L19

tannerlinsley:
Still working on this?

juhanakristian:
Yeah, sorry for taking this long. I’m also migrating it to useReactTable.

examples/react/full-width-resizable-table/src/main.tsx
juhanakristian: Example is now using React 18.2.0

examples/react/full-width-resizable-table/yarn.lock
juhanakristian: yarn.lock has been removed from the PR

examples/react/full-width-resizable-table/package.json
juhanakristian: faker removed from deps

examples/react/full-width-resizable-table/package.json
juhanakristian: @types/faker has been added to devDependencies

PR #4093, opened 2022-07-04
Add ability to define return type for getValue

fekete965:
It is a simple idea, I might be completely wrong here, but think it would be nice if we could define the return type of `getValue`.
It would be even better if `getValue` could find out the return type based on the `accessorKey` or `accessorFn` but I am not sure if that is possible. :/

### DIFF
--- a/packages/table-core/src/core/cell.ts
+++ b/packages/table-core/src/core/cell.ts
@@ -1,8 +1,8 @@
-import { RowData, Cell, Column, Row, Table } from '../types'
+import { RowData, Cell, Column, GetValue, Row, Table } from '../types'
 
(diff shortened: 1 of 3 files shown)


Sadnyster:
I think it should infer the type from the `TData` and the corresponding `accessorKey` like in the previous version.

fekete965:
> I think it should infer the type from the `TData` and the corresponding `accessorKey` like in the previous version.

I totally agree, I just wasn't sure if it is possible. If the previous version had that feature, I'll update this PR with a better implementation.

tannerlinsley:
Closing in favor of #4109

PR #4109, opened 2022-07-06
feat: column/cell/header value types

tannerlinsley:


KevinVandy (approved):
I think this is good. Easier than the `as unknown as myType`, though maybe devs will make TypeScript lie to themselves this way 🤷🏻‍♂️

fekete965 (approved):
So much batter than my "approach" 😅

tannerlinsley:
You can now extend these core types like so:

```tsx
// Could also be the core or any other adapter
declare module '@tanstack/react-table' {
  interface TableMeta {
    foo: string
  }

  interface ColumnMeta {
    bar: number
  }

  interface FilterMeta {
    rank: number
  }
}
```

fekete965 (approved):

Martijn-Schoenmaker-Webbio:
Is is just me or does this still not solve the getValue() types? When creating new columnDefs like this `ColumnDef<MyType>`
the `row.getValue()` is still unknown. This used to be no problem in the beta.

Or am I missing someting?

bazalp:
> Oui, il semble qu'il retourne inconnu. :/
> 
> Capture d&apos;écran 2022-07-12 à 15 15 07

I see the same problem

PR #4118, opened 2022-07-08
create simple non-grouped columns react example

KevinVandy:
Enough people have asked for a simpler example since the "basic" example is not simple enough 🤷 

Can follow up with Solid, Svelte, and Vue examples


tannerlinsley:
Sounds good. Let's name this new one `basic` and move the existing basic to be `column-groups`?

tannerlinsley:
Or would you rather keep the basic with groups?

KevinVandy:
> Sounds good. Let's name this new one `basic` and move the existing basic to be `column-groups`?

I agree. I'll update this PR to include the other adapter examples too

KevinVandy:
I had originally also updated this for Vue, but could not get any Vue sandboxes to work online, or locally.

Basic and Column Groups examples updated/created for React, Solid, and Svelte

PR #4151, opened 2022-07-14
docs: fix a few typos in `column-defs.md`

twhitbeck:


KevinVandy (commented):
I think the `props.renderValue()` was actually now the new recommended way to render cells as of 8.1. Though I don't see anywhere else in the docs where its referenced. 

Might have just been a convo I had to with @tannerlinsley that I'm misremembering though. We need clarification on `cell.getValue()` vs `cell.renderValue()`

docs/guide/column-defs.md
```diff
@@ -209,7 +209,7 @@ You can provide a custom cell formatter by passing a function to the `cell` prop
 ```tsx
 {
   accessorKey: 'firstName',
-  cell: props => <span>{props.renderValue().toUpperCase()}</span>,
+  cell: props => <span>{props.getValue().toUpperCase()}</span>,
```
tannerlinsley, line 212: renderValue is correct. It needs docs, but please leave it.

PR #4198, opened 2022-07-21
feat: React.useRef -> React.useMemo

StarHosea:
React.useRef -> React.useMemo to simplify code


tannerlinsley:
I appreciate the suggestion but I think we’ll keep it this way for now. Thanks!

Issue #4212, opened 2022-07-23
8.3.1 breaks getIsSomeRowsSelected behaviour


EvHaus:
### Describe the bug

After upgrading from 8.3.0 to 8.3.1 I've noticed that `table.getIsSomeRowsSelected()` will return `true` even if no rows are selected. Rolling back to 8.3.0 makes the issue go away. 8.3.2 also has this bug.

I suspect the problem is caused by this commit (https://github.com/TanStack/table/commit/620ba46fa74a440e19a654a08b411ab1ccf42356) but not sure why exactly.

### Your minimal, reproducible example

https://tanstack.com/table/v8/docs/examples/react/row-selection

### Steps to reproduce

Check out the row selection example page: https://tanstack.com/table/v8/docs/examples/react/row-selection Note how the checkbox in the header has the indeterminate state set to true even when nothing is selected yet.

### Expected behavior

If no rows are selected, `getIsSomeRowsSelected()` should return `false`.

### How often does this bug happen?

Every time

### Platform

macOS, Chrome

### react-table version

8.3.1

### TypeScript version

4.7.4

### Terms & Code of Conduct

- [X] I agree to follow this project's Code of Conduct
- [X] I understand that if my bug cannot be reliable reproduced in a debuggable environment, it will probably not be fixed and this issue may even be closed.

EvHaus:
Thanks for the quick fix @roborich. Can confirm 8.3.3 fixes the problem.

ValentinH:
@roborich I noticed that `getIsSomeRowsSelected` returns `false` if all rows are selected. According to the documentation, this should return `true`:
![image](https://user-images.githubusercontent.com/2678610/182397300-7d9c46c4-79f9-4acb-a5bc-3f08dbc832f3.png)

IMO, it should only return false when no row is selected.

KevinVandy:
I've noticed this too, but it makes sense if using for indeterminate logic.

ValentinH:
I just checked and if both `checked` and `indeterminate` are `true`, the checkbox displays as checked and not as indeterminate. Therefore, I don't think this should be the reason 🙂

Issue #4121, opened 2022-07-08
`getToggleAllRowsSelectedHandler` is selecting not allowed rows.


jpedroh:
### Describe the bug

Hello!

The table method `getToggleAllRowsSelectedHandler` is selecting all rows, even the ones which return false for the `getCanSelect` row check. I don't know if this is either a bug or intended behaviour though.

### Your minimal, reproducible example

https://codesandbox.io/s/romantic-silence-sh1x95

### Steps to reproduce

1. First, notice that by table definition in line 115,  only rows with age greater than 18 should be able to be selected. `enableRowSelection: (row) => row.original.age > 18,`
2.  Check that the checkboxes for ages smaller than 18 are disabled.
3. Tick select all. All the boxes will be checked, even the disabled ones.

### Expected behavior

The disabled checkboxes should not be checked.

### How often does this bug happen?

Every time

### Platform

React 18, Chrome 103, Windows 10.

### react-table version

8.2.0

### Terms & Code of Conduct

- [X] I agree to follow this project's Code of Conduct
- [X] I understand that if my bug cannot be reliable reproduced in a debuggable environment, it will probably not be fixed and this issue may even be closed.

LetItRock:
hey @jpedroh, @tannerlinsley!

@tannerlinsley is this just a matter of a missed call to `row.getCanSelect()` in the `RowSelection.ts:111`?

jpedroh:
Unfortunately, I didn't have the time to check the source code for Row Selection, but looking from outside, it indeed looks like a missing filter in the selectAll method.

jpedroh:
All right, had the time to give the code a look. I agree with @LetItRock solution. However, I believe there's room for improvement in also adding this check in the `mutateRowIsSelected` function itself - so this bug is avoided in other places as well (toggleAll does not use it). I've opened a PR so we can discuss these changes there.

jpedroh:
Not so fast on this one! This change introduced a bug in `getIsAllRowsSelected`, since the method also doesn't use a filter to only match the rows which are allowed to be selected. This breaks the toggle selection feature, because the toggle method - which was patched - calls this method.

Let's make this right. I would like to implement some tests to assert this behaviour is correct on my next PR. @tannerlinsley is it ok if I implement a test for this feature in the react package or would you rather have a test implemented in the core one? Perhaps setting up infrastructure for the core package will be a little bit trickier because of state management?

tannerlinsley:
[...]

PR #4232, opened 2022-07-27
tests: add tests for row selection feature

jpedroh:
This adds a test for the behaviour described in #4121. It also covers the bug which was fixed in #4212. It also adds a few more common use cases.

### DIFF
--- a/.github/workflows/pr.yml
+++ b/.github/workflows/pr.yml
@@ -13,7 +13,7 @@ jobs:
         with:
           node-version: ${{ matrix.node }}
       - run: |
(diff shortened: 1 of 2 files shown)


jpedroh:
@tannerlinsley maybe this suite can substitute the skipped one in [packages/react-table/__tests__/features/useRowSelect.notest.js](https://github.com/TanStack/table/blob/main/packages/react-table/__tests__/features/useRowSelect.notest.js)?

KevinVandy (approved):

PR #4233, opened 2022-07-27
add sub components example

berdyshev:


KevinVandy (approved):

KevinVandy (changes requested):
We also just need to add a workspace entry in the root package.json for this new example

KevinVandy:
guess I'll do that myself

PR #4250, opened 2022-08-01
Remove yarn.lock from previously added examples

fekete965:
I saw that in other PRs, people asked the author to remove the committed yarn.lock file.

None of the other example folders contains yarn.lock and I would like to match my previously added examples to that.


fekete965:
I will investigate the failure as soon as I have a bit more time this week!

fekete965:
#4232 seems to have a working solution regarding the workflow failure.

fekete965:
@KevinVandy thanks for the heads up. I've seen in a few other PRs that removing the lock file was "mandatory". I thought I should follow that example.
Shall we delete the branch?

Issue #4223, opened 2022-07-25
paginateExpandedRows problem ?


Its-Just-Nans:
### Describe the bug

I am using the model `getExpandedRowModel` and I wanted to display subRows and not to be paginated.

So I setted the `paginateExpandedRows` to `false`.
But that is not working :/

Also, in the code I so that comment https://github.com/TanStack/table/blob/d5d5501c5a0b76bda45dcf55f0d466fe7c7f63d2/packages/table-core/src/utils/getExpandedRowModel.ts#L17
```
          // Do not expand if rows are not included in pagination
          !paginateExpandedRows ||
```
Why we do not expand the table ? The `paginateExpandedRows` should be used for pagination only no ?

### Your minimal, reproducible example

https://codesandbox.io/s/react-table-paginateexpandedrows-bug-wv3hd2

### Steps to reproduce

1. Go to https://codesandbox.io/s/react-table-paginateexpandedrows-bug-wv3hd2
2. See that the option `paginateExpandedRows` is set to `false`
3. Try expand the table
4. Not working :(

### Expected behavior

Expand the table (and don't update pagination)

### How often does this bug happen?

Every time

### Platform

- Browser: Chrome (and I suppose others)

### react-table version

8.3.3

### Terms & Code of Conduct

- [X] I agree to follow this project's Code of Conduct
- [X] I understand that if my bug cannot be reliable reproduced in a debuggable environment, it will probably not be fixed and this issue may even be closed.

dhanushkumarsivaji:
I am also facing the same problem :(

Its-Just-Nans:
No direct fix for this, sorry

I made a PR ( #4263 ), but not sure if it's compatible with others options

dhanushkumarsivaji:
@tannerlinsley could u fix this one ?

tannerlinsley:
Working on it

Tanner Linsley

tannerlinsley:
Fixed!

PR #4263, opened 2022-08-03
try fixing paginateExpandedRows

Its-Just-Nans:
Hi

For me it's fixing #4223 

**But be careful, I didn't check if it's working with all the others options.**


tannerlinsley:
This is close, but requires more code in a few other places. We're working on a wholistic fix soon. Thanks!

PR #4267, opened 2022-08-04
examples: create column ordering DnD example with react-dnd

KevinVandy:


examples/react/column-dnd/src/main.tsx
```diff
@@ -0,0 +1,220 @@
...
+      ref={dropRef}
+      key={header.id}
+      colSpan={header.colSpan}
+      style={{ opacity: isDragging ? 0.5 : 1 }}
+    >
+      <div ref={previewRef}>
+        {header.isPlaceholder
+          ? null
+          : flexRender(header.column.columnDef.header, header.getContext())}
+        <button ref={dragRef}>👊</button>
```
tannerlinsley, line 120:
```suggestion
        <button ref={dragRef}>🟰</button>
```

Issue #4241, opened 2022-07-29
columnHelper.accessor Type issue


opti21:
### Describe the bug

When defining the columns for the table, `columnHelper.accessor` seems to cause a type issue. Although, table seems to render fine.

Type error:
`Type 'ColumnDef<Room, string>' is not assignable to type 'ColumnDef<Room, unknown>'.`

StackBlitz Type Error:
```
Type instantiation is excessively deep and possibly infinite.(2589)
(property) accessor: <"id", string>(accessor: "id", column: Omit<ColumnDef<Room, string>, "accessorKey">) => ColumnDef<Room, string>
```

### Your minimal, reproducible example

https://stackblitz.com/edit/github-sq9c9a?file=pages%2Findex.tsx

### Steps to reproduce

Line 22 on `index.tsx`

### Expected behavior

No type error

### How often does this bug happen?

Every time

### Platform

Windows 11 Pro
Chrome: Version 103.0.5060.134 (Official Build) (64-bit)

### react-table version

8.5.2

### TypeScript version

4.7.4

### Terms & Code of Conduct

- [X] I agree to follow this project's Code of Conduct
- [X] I understand that if my bug cannot be reliable reproduced in a debuggable environment, it will probably not be fixed and this issue may even be closed.

dustinwdev:
We just installed the latest version with a fresh Next.js app and I am seeing the exact same issue with columnHelper types.

tannerlinsley:
Working on a fix right now

Tanner Linsley

ebramanti:
Is this related? https://github.com/TanStack/table/discussions/4249

tannerlinsley:
We just shipped new types for the columnHelper along with better underlying column def types. Will you try them out and report back?

ebramanti:
My compiler is still hanging with this type @tannerlinsley

opti21:
Both vscode and stackblitz still show the error: https://stackblitz.com/edit/github-sq9c9a?file=package.json,pages%2Findex.tsx

tannerlinsley:
I'm definitely seeing the error in stackblitz, but can't replicate locally in the examples. Is stackblitz on the latest version of TS?

opti21:
both local and the stackblitz is on 4.7.4 and it still shows the error.

sjbuysse:
Same here, haven't been able to use the columnHelper since it's creation not too long ago.

Obiwarn:
FYI: updated to v8.5.6 labeled:

[...]

PR #4268, opened 2022-08-04
fix: columnHelper not fit with Column type

Himself65:
Closes: https://github.com/TanStack/table/issues/4241


tannerlinsley:
Fixing this a different way. Thanks!

Issue #4302, opened 2022-08-14
[v8] Type Error in ColumnDef


celsodias12:
### Describe the bug

`ColumnDef` typing is not working correctly, when a type is passed as a parameter it doesn't show me the correct typing

### Your minimal, reproducible example

https://codesandbox.io/s/react-typescript-forked-gr4xw1?file=/src/App.tsx

### Steps to reproduce

when you fill in the `accessorKey` you will see that its typing is string, and it should be `keyof Person`

### Expected behavior

should return the typing as: `accessorKey: firstName | lastName | age | visits | status | progress` and any other value that is entered should be returned typing error

### How often does this bug happen?

Every time

### Screenshots or Videos

![image](https://user-images.githubusercontent.com/47264001/184558845-8e7ccdb5-7c73-4b19-9e8d-106b13e9bce8.png)

### Platform

- OS: windows
- Browser: any browser

### react-table version

v8.5.11

### TypeScript version

v4.6.3

### Terms & Code of Conduct

- [x] I agree to follow this project's Code of Conduct
- [X] I understand that if my bug cannot be reliable reproduced in a debuggable environment, it will probably not be fixed and this issue may even be closed.

tannerlinsley:
Please use the createColumnHelper<Person>().accessor(options) API

Tanner Linsley

iblecher:
That did work for me - columnHelper returns a type for a specific field of the column and not the entire column type.
Example:
```
type Person = {
  name: string;
  age: number;
};

const columnHelper = createColumnHelper<Person>();


// Type of nameColumn = ColumnDef<Person, string>;
const nameColumn = columnHelper('name', {
  header: `Name`,
  cell: info => info.getValue(),
});

// Type of ageColumn = ColumnDef<Person, number>;
const ageColumn = columnHelper('age', {
  header: `Age`,
  cell: info => info.getValue(),
});

// This will not work:
const columns: ColumnDef<Person> = [nameColumn, ageColumn];

**Comment by utkuturunc** [comment]
i'm working around this by using

```typescript
const columns: ColumnDef<Person, any>[] = [nameColumn, ageColumn];
```

iblecher:
thanks @utkuturunc but we use `@typescript-eslint/no-explicit-any` rule which doesn't allow me to use `any`

utkuturunc:
@iblecher in that case

```typescript
const nameColumn = columnHelper('name', {
  header: `Name`,
  cell: info => info.getValue(),
[...]

PR #4303, opened 2022-08-14
fix: ColumnDef type

celsodias12:
this adds a fix on the columnDef type described at https://github.com/TanStack/table/issues/4302


KevinVandy:
Not sure this applies anymore after #4419 got merged, and this now has conflicts, so am going to close for now.

PR #4352, opened 2022-08-31
Adding the react plugin for Vite

cereallarceny:
Adding the plugin to the full-width example. Currently, this triggering a 502 gateway issue with CodeSandbox. Adding the plugin fixes the issue.


KevinVandy:
handled in #4528

Issue #4271, opened 2022-08-05
Page index gets set to -1 with manual pagination and zero pages


lexanth:
### Describe the bug

When using manual pagination, if we have zero pages of results, any attempt to set the pageIndex gives a pageIndex of -1 (which would often be invalid as an API parameter).

* `autoResetPageIndex` doesn't work with manual pagination, because it resets to page zero whenever the data changes, which it inherently will do with manual pagination
* As a workaround (or legitimate functionality, if the page count is likely to reduce while paging through), we have/had an implementation to reset the page index if it got "out of bounds". This was trying to set pageIndex to 0 or pageCount - 1 (or the greater of the two), but if pageCount = 0, react-table would "clamp" it to -1 anyway.
* Better functionality is for the page index to reset to 0 when changing the filter/sort order, but I still think that setting pageIndex of -1 is incorrect
* Another workaround is when passing the pageIndex parameter to the API to make sure it is "reasonable", but I don't think react-table should be setting an unreasonable value in the first place
* Another workaround would be to say that we should never have zero pages, we should always have at least 1, even if it's empty. But that doesn't necessarily seem right either.

https://github.com/TanStack/table/blob/main/packages/table-core/src/features/Pagination.ts#L123-L139 is the relevant bit - the way that the clamp works assumes that pageCount - 1 > 0, but it isn't necessarily.

### Your minimal, reproducible example

https://codesandbox.io/s/confident-roman-iwj4le?file=/src/main.tsx

### Steps to reproduce

1. Go next page
2. Type "no_results" into the search box
3. See that pageIndex gets set to -1, and the fake fetchData API errors because of it

### Expected behavior

react-table shouldn't set pageIndex to -1 - it should not set it to < 0

### How often does this bug happen?

Every time

### Platform

macOS, chrome 102

### react-table version

8.5.6

### TypeScript version

4.5.5

### Additional context

I think the workarounds are reasonable and probably give a better UX, but I still think that the clamp behaviour is wrong because it shouldn't be possible to get to pageIndex -1

### Terms & Code of Conduct

- [X] I agree to follow this project's Code of Conduct
- [X] I understand that if my bug cannot be reliable reproduced in a debuggable environment, it will probably not be fixed and this issue may even be closed.

carlosbaraza:
[...]

PR #4370, opened 2022-09-09
Don't set pageIndex to -1 when resetting page index with pageCount = 0

lexanth:
Fixes https://github.com/TanStack/table/issues/4271

### DIFF
--- a/packages/table-core/src/features/Pagination.ts
+++ b/packages/table-core/src/features/Pagination.ts
@@ -130,7 +130,7 @@ export const Pagination: TableFeature = {
               ? Number.MAX_SAFE_INTEGER
               : table.options.pageCount - 1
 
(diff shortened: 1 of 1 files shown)


packages/table-core/src/features/Pagination.ts
```diff
@@ -130,7 +130,7 @@ export const Pagination: TableFeature = {
               ? Number.MAX_SAFE_INTEGER
               : table.options.pageCount - 1
 
-          pageIndex = Math.min(Math.max(0, pageIndex), maxPageIndex)
```
lexanth, line 133: These only differ in the case where maxPageIndex is < 0, which can only happen when pageCount <= 0

KevinVandy (approved):

PR #4373, opened 2022-09-09
fix: filter should filter out null values

### LINKED ISSUE #4245: [v8] Filter not working if accessorFn returns null for any row
### Describe the bug

Filtering a column defines like:

```
{
        id: 'name',
        header: 'Name',
        accessorKey: 'name',
}
```

works, even if that column has null values.

Now when doing 

```
{
        id: 'name',
        header: 'Name',
        accessorFn: (row) => row.name !== 'ignore' ? row.name : null,
}
```

in this case filtering does not work anymore at all.
So even not for the rows returning a value.

I can use this workaround:

```
{
        id: 'name',
        header: 'Name',
        accessorFn: (row) => row.name !== 'ignore' ? row.name : '',
}
```

but this should not be needed I would say.

-

### Steps to reproduce

See above.

### Expected behavior

Rows with actual value to be filtered succesfully.

Every time

### Screenshots or Videos

_No response_

### Platform

macOS

### react-table version

v8.5.1, v.8.5.2

### TypeScript version

v4.7.4

### Additional context

_No response_

### Terms & Code of Conduct

simPod:
Fixes #4245

### DIFF
--- a/packages/table-core/__tests__/filterFns.test.ts
+++ b/packages/table-core/__tests__/filterFns.test.ts
@@ -1,0 +1,132 @@
+import { createRow, createTable, filterFns, getCoreRowModel } from '../src';
+
+const ColumnId = 'col';
+
+let table = createTable({
+  data: [],
+  state: {},
+  getCoreRowModel: getCoreRowModel(),
+  onStateChange: () => {
+  },
+  renderFallbackValue: () => {
+  },
+  columns: [{
+    accessorKey: 'prop',
+    id: ColumnId,
+  }],
+});
+
+const createRowWithValue = (value: unknown) => createRow(
+  table,
+  'A',
+  { prop: value },
+  1,
+  0,
+  [],
+);
+
+describe('includesString', () => {
+  test.each([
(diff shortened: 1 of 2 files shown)


simPod:
@tannerlinsley I suppose the `autoRemove` should somehow get involved so maybe my fix is in wrong direction?

Though I was thinking those tests might be valuable nonetheless.

stephanvdhorn:
We currently experience problems because some of our values are null, so could this get a higher priority?

ahammouda:
This is also super useful if you are trying to run filter functions over virtualized rows - I've copied the filter function in a project just to get me over that hurdle.  Definitely +1.

KevinVandy:
You guys can override the filter fns with your own custom ones to fit your requirements

simPod:
Ofc but that's not a solution that benefits others nor systematic one. I'd like to solve it on library level though workaround are possible.

KevinVandy:
Closing this in favor of #4441, as it got fixed in v8.5.26

Issue #4365, opened 2022-09-06
react-table-v7 documentation UI breaks


himansh-gjr:
### Describe the bug

I was going through the documentation of the react-table-v7 and the moment I scrolled down, two components from right section of the page started to collapse and stacked on each other.

<img width="1438" alt="Screenshot 2022-09-06 at 7 28 13 PM" src="https://user-images.githubusercontent.com/77732259/188654429-5d5c5777-a682-474a-9020-ece1c88c0c9e.png">

### Your minimal, reproducible example

https://react-table-v7.tanstack.com/docs/overview#what-is-a-headless-ui-library

### Steps to reproduce

1. Go to [react-table-v7-overview](https://react-table-v7.tanstack.com/docs/overview)
2. Scroll down to  `#what-is-a-headless-ui-library `
3. see an error

### Expected behavior

As a user, If I have the option to navigate to a specific section of the page from the 'ON THIS PAGE' component it should be 
clearly visible to me all the time but I am unable to navigate to the other parts of the page because while scrolling two components are stacked on each other.

### How often does this bug happen?

Every time

### Platform

OS: macOS - Browser: Chrome - Version : 104.0.5112.79

### react-table version

v7

### Terms & Code of Conduct

- [X] I agree to follow this project's Code of Conduct
- [X] I understand that if my bug cannot be reliable reproduced in a debuggable environment, it will probably not be fixed and this issue may even be closed.

framerate:
Guys, you completely broke your v7 documentation and I'm doing an emergency hotfix. It doesn't really make for a good taste in my mouth for whatever "TanStack" is trying to  be :(

arturoribessanz:
Same here. There are links to v7 docs but not working. Looks like I have to clone the repository and run my own documentation app...
Yes, I want to update to v8, but I have no time yet and v7 works well. Not yet worth the investment.

Joe-Cahill-Greystone:
bumping, it would be nice to have the v7 docs on a working site or github pages

jsc144:
It would be nice to get the v7 docs back up, it looks bad for LTS for these to go down. Perhaps we could host on github pages?

PR #4381, opened 2022-09-14
docs: Fixed an issue with overlapping Bytes Form in v7.

dazlygonsalves:
Fixed an issue in v7 docs where the "Subscribe to Bytes form would overlap the "On This Page" section on Scroll as described in
#4365


mzaharie:
The site redirects [v7](https://tanstack.com/table/v7) to [v8](https://tanstack.com/table/v8) can anybody also fix this?

PR #4393, opened 2022-09-17
[CREATE] create rendertemplate for template syntax

kyumoon:
hello
I think 'cell' needs html rendering.
currently, The FlexRender in vue-table doesn't support that feature.
so I implemented renderTemplate.
if appropriate, it would be nice to create a function like this.
thank you.

plugin: es6-string-html


KevinVandy:
I think the vue adapter does need something like this, but not sure if we want to put this in the example right now. Is the es6-string-html safe and good practice? Or is it more like dangerously__render...?

PR #4398, opened 2022-09-22
Update react-table.md

AliwerDev:


KevinVandy:
don't think this code snippet should be deleted. Probably an accidental PR?

PR #4399, opened 2022-09-22
filter out invalid sort id

missing1984:
When user provides invalid sort id either from initialState or state the table blows up.

This PR improve default getSortedRowModel to guard against this case.

Behavior change:

- Invalid sort id will be ignored and table can still render.

### DIFF
--- a/packages/table-core/src/utils/getSortedRowModel.ts
+++ b/packages/table-core/src/utils/getSortedRowModel.ts
@@ -18,8 +18,9 @@ export function getSortedRowModel<TData extends RowData>(): (
         const sortedFlatRows: Row<TData>[] = []
 
         // Filter out sortings that correspond to non existing columns
+        const columns = table._getAllFlatColumnsById();
         const availableSorting = sortingState.filter(sort =>
-          table.getColumn(sort.id).getCanSort()
+          columns[sort.id] && columns[sort.id]?.getCanSort()
         )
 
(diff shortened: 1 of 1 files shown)


jlaneve:
+1, ran into the same issue

packages/table-core/src/utils/getSortedRowModel.ts
```diff
@@ -18,8 +18,9 @@ export function getSortedRowModel<TData extends RowData>(): (
         const sortedFlatRows: Row<TData>[] = []
 
         // Filter out sortings that correspond to non existing columns
+        const columns = table._getAllFlatColumnsById();
```
KevinVandy, line 21: I think we want to avoid the getAll call here, and simply just add the optional chaining down below maybe?

missing1984: We're using `table.getColumn` which internally uses `table._getAllFlatColumnsById()` anyway. It's a cache value and accessing its value is O(1) operation so it may be just fine?

KevinVandy:
should be fixed in #4490

KevinVandy:
I think we are going to close this in favor of changing `table.getColumn` itself possibly not throwing an error, and instead just console.error()ing and returning undefined as that is the root of this problem. Still need to look into the possible side effects of doing that a bit more.

Issue #4320, opened 2022-08-18
[v8] dot in accessorKey crashes the table


oeddyo:
### Describe the bug

looks like when I try to upgrade from 8.2.3 -> 8.5.11
it crashes my app and I found out through a test case. When accessor key has dot, it crashes in 8.5.11 

Might due to recent change though I'm not 100% sure

```
commit f7be4cfb25eff51a8438346ffbb0be039bc1be2e
Author: Tanner Linsley <tannerlinsley@gmail.com>
Date:   Wed Jul 20 09:29:05 2022 -0600

    feat: column helper utility, deep accessorKey support (#4185)

    * feat: column helper utility, deep accessorKey support

    * Update main.tsx

    * fix: export column def union types

diff --git a/packages/table-core/src/core/column.ts b/packages/table-core/src/core/column.ts
--- a/packages/table-core/src/core/column.ts
+++ b/packages/table-core/src/core/column.ts
@@ -98,1 +94,1 @@
-    resolvedColumnDef.accessorKey ??
+    (accessorKey ? accessorKey.replace('.', '_') : undefined) ??
```

### Your minimal, reproducible example

https://codesandbox.io/s/quizzical-sun-4zm023?file=/src/main.tsx

### Steps to reproduce

1. create a simple table
2. give one of the column name "a.b"
3. use accessorkey a.b

### Expected behavior

it should show a.b in header
instead it crashes the table

### Platform

sandbox
mac os
v8.5.11

### react-table version

v8.5.11

### Terms & Code of Conduct

- [X] I agree to follow this project's Code of Conduct
- [X] I understand that if my bug cannot be reliable reproduced in a debuggable environment, it will probably not be fixed and this issue may even be closed.

oeddyo:
> I'm not sure it's a bug, although I agree that it can be confusing.
> 
> `.` in `accessorKey` is used to access deeply nested keys in objects.
> 
> When providing a `a.b` as the `accessorKey` `react-table` will expect `{ a: { b: "value" }` instead of `{ "a.b.": "value" }`.
> 
> Unless the maintainers want to support this use case it might be helpful to provide a more helpful error message explaining this.

From the code I think it was intentionally supporting using a dot as value since it's replacing dots with underscore. We relied on the assumption that dots in names are supported but thankfully tests caught the issues when upgrading. I believe maintaining consistency is important for users and the library. What do you think?

dennemark:
[...]

PR #4407, opened 2022-09-27
feat: error message for invalid nested keys

petermakowski:
Provide a more helpful error message when using nested keys with missing data.

Fixes: https://github.com/TanStack/table/issues/4320

### DIFF
--- a/packages/table-core/src/core/column.ts
+++ b/packages/table-core/src/core/column.ts
@@ -56,6 +56,11 @@ export function createColumn<TData extends RowData, TValue>(
 
         for (const key of accessorKey.split('.')) {
           result = result[key]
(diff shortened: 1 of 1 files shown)


KevinVandy (approved):

ricsam:
How could one allow missing data? What if a nested object is optional and I want it to resolve to undefined, is there an option for that?

KevinVandy:
> How could one allow missing data? What if a nested object is optional and I want it to resolve to undefined, is there an option for that?

That's where using a custom accessorFn is handy. You control the exact logic for how the data gets resolved.

ricsam:
Ah, that is awesome! Thank you

Issue #4411, opened 2022-09-29
Subrow sorting will not return to default. 


Victor2333:
### Describe the bug

With sub row and sorting function. Do sorting and sub rows will be sorted. But when recover to not sorting. The sub rows still are sorted.

### Your minimal, reproducible example

https://codesandbox.io/s/morning-violet-3i7rhi?file=/src/main.tsx

### Steps to reproduce

1. Click age to sort age column
2. click again.
3. click again.
4. age column is not sorted but the sub rows still are sorted.

### Expected behavior

Sub rows should return to the default value.

### How often does this bug happen?

Every time

### Platform

Chrome.

### react-table version

v8.5.13

### TypeScript version

v4.8.3

### Terms & Code of Conduct

- [X] I agree to follow this project's Code of Conduct
- [X] I understand that if my bug cannot be reliable reproduced in a debuggable environment, it will probably not be fixed and this issue may even be closed.

PR #4412, opened 2022-09-29
bug: default subrow will change after sorting

Victor2333:
Related to #4411

### DIFF
--- a/packages/table-core/src/utils/getSortedRowModel.ts
+++ b/packages/table-core/src/utils/getSortedRowModel.ts
@@ -44,7 +44,7 @@ export function getSortedRowModel<TData extends RowData>(): (
         const sortData = (rows: Row<TData>[]) => {
           // This will also perform a stable sorting using the row index
           // if needed.
-          const sortedData = rows.slice()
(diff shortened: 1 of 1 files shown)


packages/table-core/src/utils/getSortedRowModel.ts
```diff
@@ -44,7 +44,7 @@ export function getSortedRowModel<TData extends RowData>(): (
         const sortData = (rows: Row<TData>[]) => {
           // This will also perform a stable sorting using the row index
           // if needed.
-          const sortedData = rows.slice()
+          const sortedData = rows.map(r => Object.assign({}, r))
```
KevinVandy, line 47: Would 

```js
const sortedData = [...rows]
```

do the same thing without having to loop through every row?

KevinVandy: testing it locally and the spread seems like the better efficient option

KevinVandy (changes requested):

KevinVandy:
Closed in favor of #4490

PR #4419, opened 2022-10-02
perf: optimise types

terry-au:
This PR optimises TypeScript types, following [Microsoft's Writing Easy-to-Compile Code](https://github.com/microsoft/TypeScript/wiki/Performance#writing-easy-to-compile-code) section of the TypeScript wiki.

After migrating to version 8, I encountered severe performance degradation in a file with 100+ column definitions, to the point that TS type checking and autocomplete ceased to function in the editor. I believe IntelliJ uses the standard TypeScript language server, so these issues may also crop up in other editors.

The changes in this PR have resolved the issues I was having. Most of the work in this PR migrates from `type` to `interface` based type definitions, but there are a few smaller optimisations as well. The aforementioned link to the TS wiki details the benefits of using `interface` when possible.

## Results
These trace excerpts were generated with `tsc --generateTrace my_trace && npx analyze-trace my_trace/ --forceMillis 100 --skipMillis 50 --color`.

**Before:**
`Check file <path>/column-definitions.tsx (394ms)`

**After:**
`Check file <path>/column-definitions.tsx (287ms)`

### DIFF
--- a/packages/table-core/src/columnHelper.ts
+++ b/packages/table-core/src/columnHelper.ts
@@ -50,7 +50,7 @@ import { DeepKeys, DeepValue } from './utils'
 // })
 
 type Accessor<TData extends RowData> = <
-  TAccessor extends AccessorFn<TData> | DeepKeys<TData>,
+  TAccessor extends AccessorFn<TData> | string,
   TValue extends TAccessor extends AccessorFn<TData, infer TReturn>
     ? TReturn
     : TAccessor extends DeepKeys<TData>
--- a/packages/table-core/src/columnHelper.ts
+++ b/packages/table-core/src/columnHelper.ts
@@ -6,7 +6,7 @@ import {
   IdentifiedColumnDef,
   RowData,
 } from './types'
-import { DeepKeys, DeepValue } from './utils'
+import { DeepValue } from './utils'
 
 // type Person = {
 //   firstName: string
@@ -53,9 +53,7 @@ type Accessor<TData extends RowData> = <
   TAccessor extends AccessorFn<TData> | string,
   TValue extends TAccessor extends AccessorFn<TData, infer TReturn>
     ? TReturn
-    : TAccessor extends DeepKeys<TData>
-      ? DeepValue<TData, TAccessor>
-      : never
+    : DeepValue<TData, TAccessor>
   >(
   accessor: TValue extends never ? never : TAccessor,
   column: TAccessor extends AccessorFn<TData>
--- a/packages/table-core/src/columnHelper.ts
+++ b/packages/table-core/src/columnHelper.ts
@@ -49,17 +49,24 @@ import { DeepValue } from './utils'
 //   cell: info => info.getValue(),
 // })
 
+type AccessorType<TData, TAccessor> = TAccessor extends AccessorFn<
+  TData,
+  infer TReturn
+>
+  ? TReturn
+  : DeepValue<TData, TAccessor>
+
 type Accessor<TData extends RowData> = <
   TAccessor extends AccessorFn<TData> | string,
-  TValue extends TAccessor extends AccessorFn<TData, infer TReturn>
-    ? TReturn
-    : DeepValue<TData, TAccessor>
-  >(
+  TValue extends AccessorType<TData, TAccessor>
+>(
   accessor: TValue extends never ? never : TAccessor,
   column: TAccessor extends AccessorFn<TData>
     ? DisplayColumnDef<TData, TValue>
     : IdentifiedColumnDef<TData, TValue>
-) => AccessorColumnDef<TData, TValue>
+) => TValue extends never
+  ? never
+  : AccessorColumnDef<TData, AccessorType<TData, TAccessor>>
 
 export interface ColumnHelper<TData extends RowData> {
   accessor: Accessor<TData>
--- a/packages/table-core/src/columnHelper.ts
+++ b/packages/table-core/src/columnHelper.ts
@@ -1,12 +1,12 @@
 import {
+  AccessorColumnDef,
   AccessorFn,
-  ColumnDef,
   DisplayColumnDef,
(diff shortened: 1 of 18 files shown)


tannerlinsley:
Wow, incredible work! I’m currently undergoing something similar with TanStack Router. The burning question here: is any of this breaking?

packages/table-core/src/columnHelper.ts
```diff
@@ -49,22 +49,24 @@ import { DeepKeys, DeepValue, RequiredKeys } from './utils'
...
 // })
 
-export type ColumnHelper<TData extends RowData> = {
-  accessor: <
-    TAccessor extends AccessorFn<TData> | DeepKeys<TData>,
-    TValue extends TAccessor extends AccessorFn<TData, infer TReturn>
-      ? TReturn
-      : TAccessor extends DeepKeys<TData>
+type Accessor<TData extends RowData> = <
+  TAccessor extends AccessorFn<TData> | string,
```
terry-au, line 53: `DeepKeys<TData>` is expensive, and the result of calling it doesn't seem to be cached by the compiler. We can replace the first type check with `string`, then, if the accessor is a function, we can short circuit and avoid ever calling `DeepKeys`. If the accessor is not a function, we simply narrow the type from `string` to a key.

terry-au:
> Wow, incredible work! I’m currently undergoing something similar with TanStack Router. The burning question here: is any of this breaking?

From what I can tell, it shouldn't be. I just fixed a regression I picked up on in `columnHelper.ts`. I've got some several hundred column definitions in my project. I don't use all of the extensions, so I haven't been able to verify if there's anything breaking as of yet. It'd be cool if someone who has wider feature use and valid types to pull this version in and see if anything is broken.

tannerlinsley:
Has any of the runtime functionality broken with this? Or just types?

KevinVandy:
@terry-au  I've been trying to test this PR in my own project. It seems to break the current DeepKeys functionality. Is the DeepValue supposed to be able to be used in the same way?

terry-au:
> @terry-au I've been trying to test this PR in my own project. It seems to break the current DeepKeys functionality. Is the DeepValue supposed to be able to be used in the same way?

@KevinVandy How were you using `DeepKeys`? The issue with `DeepKeys`, is that it recursively walks through a type. This can cause a lot of issues for types with many properties and types that contain references to themselves. It's kind of like BFS, but it is the worst-case 100% of the time, since it _always_ calculates every single key path.

This PR drops `DeepKeys` for `DeepValue<T, TProp>`, and should run at _O(1 • d)_, where _d_ is the depth of the property path.

I just added a utility type `HasProperty`, which should error if an invalid path is supplied. That should cover one such use case for `DeepKeys` that this PR would have broken. I intentionally did not use the name `DeepKeys`, as 

```typescript
interface FooBar {
  foo: string;
  bar: {
    baz: FooBar;
  }
}

const myFn = <T extends string>(pp: HasProperty<FooBar, T>): number => {
  return 123;
}

myFn('foo'); // works
myFn('bar'); // works
myFn('bar.baz.foo'); // works
myFn('bar.invalid'); // error
```

KevinVandy:
> I just added a utility type HasProperty, which should error if an invalid path is supplied. That should cover one such use case for DeepKeys that this PR would have broken. I intentionally did not use the name DeepKeys, as

@terry-au If it works like that we are set. I'm personally running into some problems though that may or may not be my fault.

I'm trying to use it like this:

image

But am currently getting TS errors from trying to use it

image

KevinVandy:
@terry-au Would love to merge this, but would like to get some more verification about the `HasProperty` typesafety. At the very least, we might want to keep `DeepKeys<TData>` as an optional import. As far as I can tell, the `HasProperty` helper does not have the same TS auto-complete as DeepKeys, and I know a lot of devs really like that feature.

terry-au:
> @terry-au Would love to merge this, but would like to get some more verification about the `HasProperty` typesafety. At the very least, we might want to keep `DeepKeys<TData>` as an optional import. As far as I can tell, the `HasProperty` helper does not have the same TS auto-complete as DeepKeys, and I know a lot of devs really like that feature.

@KevinVandy I've updated this PR to re-add `DeepKeys` based on [this](https://stackoverflow.com/a/58436959). I also updated column helper to now calculate deep keys as part of the generic param. Hopefully this means the keys are only calculated once per column-helper instance. I haven't run into the same performance issues as I did with the current release, but haven't tested extensively yet.

KevinVandy (approved):

KevinVandy:
@tannerlinsley @terry-au I think this is good to go. No regression when tested in my project. Will merge tonight if I don't hear otherwise from Tanner.

KevinVandy:
@terry-au So even though it appeared perfectly fine in my testing locally, now that I've released, whenever I try to use the new `DeepKeys` implementation, it always results in `Type instantiation is excessively deep and possibly infinite.ts(2589)
` throughout all column definitions

terry-au:
> @terry-au So even though it appeared perfectly fine in my testing locally, now that I've released, whenever I try to use the new `DeepKeys` implementation, it always results in `Type instantiation is excessively deep and possibly infinite.ts(2589) ` throughout all column definitions

@KevinVandy can you please provide a sample type which causes `DeepKeys` to error? I can't reproduce locally when testing against types with circular references.

KevinVandy:
@terry-au I think we just need to look at these TS optimizations separately and in smaller pieces. Probably want to just get all the types converted to interfaces that we can first.

I think the Deep Keys refactor can be handled separately

supersnager:
@KevinVandy @terry-au I wanted to do the same today, but I found your work.
I want to say thank you for these PR's! Not only is it a performance improvement, but it should help with declaration merging. For example, I need to add some properties to the ColumnSort and HeaderContext, with your PR's this should be possible :) I can't wait for it to be released.

Issue #4321, opened 2022-08-18
[v8] Column resizing doesn't work in mobile browsers


Transfusion:
### Describe the bug

[The column resizing handler](https://tanstack.com/table/v8/docs/api/features/column-sizing) doesn't appear to work with mobile browsers.

### Your minimal, reproducible example

https://mpjrpm.sse.codesandbox.io/

### Steps to reproduce

1. Visit the demo https://tanstack.com/table/v8/docs/examples/react/column-sizing on a mobile device.
2. Try to resize any column.

### Expected behavior

The columns should resize just as they do in desktop mode.

### Screenshots or Videos

https://user-images.githubusercontent.com/3119646/185305392-3cea6f18-300e-48b1-ad83-89d601192bc7.mp4

### Platform

- Chrome 104 on Android
- Safari on iOS 15
- Any mobile browser which emits touch events
- Using the devtools of any desktop browser which allows you to simulate mobile mode, e.g. <https://developer.chrome.com/docs/devtools/device-mode/>

### react-table version

8.5.11

### Terms & Code of Conduct

- [X] I agree to follow this project's Code of Conduct
- [X] I understand that if my bug cannot be reliable reproduced in a debuggable environment, it will probably not be fixed and this issue may even be closed.

quynhlx:
I got same issue above. 
@lucasra1 can you update your pull request?

lucasra1:
@quynhlx There is nothing to update. If you mean the ci checks, there is nothing i can do. The github action is broken

OmarOmeiri:
There is a dependency conflict in one of the examples:
`react:18.2.0 from tanstack-table-example-virtualized-infinite-scrolling`
and
`react@"^16.6.3 || ^17.0.0" from react-virtual@2.10.4`

It got me wondering how the example PR was merged in the first place.

PR #4422, opened 2022-10-04
fix columnn resize not working on touch devices

lucasra1:
add missing event listeners for `touchmove` and `touchend`

fixes #4321

PR #4470, opened 2022-10-21
Fix column resizing doesn't work in mobile browsers

quynhlx:
Fixes https://github.com/TanStack/table/issues/4321


KevinVandy:
Duplicate of #4422

PR #4477, opened 2022-10-23
docs: add details to options and Table API

tfarhan00:
Added further explanation about options and Table API

### DIFF
--- a/docs/api/core/table.md
+++ b/docs/api/core/table.md
@@ -4,13 +4,51 @@ title: Table
 
 ## `useReactTable` / `createSolidTable` / `useVueTable` / `createSvelteTable`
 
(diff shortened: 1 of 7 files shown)


docs/api/core/table.md
```diff
@@ -97,7 +135,7 @@ declare module '@tanstack/table-core' {
 state?: Partial<
   VisibilityTableState &
   ColumnOrderTableState &
-  ColumnPinningTableState &
+  ColumnPinningTabelState &
```
module17, line 138: Looks like a typo here.

KevinVandy:
I think some of this is a good idea, but also a bit redundant. Going to close for now.

PR #4493, opened 2022-10-29
Fix sorting leafRows when there is one subRow

### LINKED ISSUE #4414: Sorting grouped rows sometimes doesn't work
### Describe the bug

Scenario: Two grouped columns and attempting to sort by a third column. 

If there is a value in the first grouped column that only has one subRow of the second grouped column, but that subRow has multiple subRows then those subsequent subRows (the leafRows of the initial row) do not get sorted when attempting to sort them.

https://codesandbox.io/s/react-table-grouping-and-sorting-bug-l8mh8t

### Steps to reproduce

1. Group the `First Name` column
2. Group the `Last Name` column
3. Expand both so all the  rows are showing
4. Click to "Sort" in the `Age` column header
5. Confirm that it does not sort the rows correctly

### Expected behavior

As a user, I expect for the rows to be sorted correctly after grouping some columns, but they are not sorted

Every time

### Screenshots or Videos

_No response_

### Platform

OS: macOS - Browser: Chrome - Version: 105.0.5195.125

### react-table version

v8.5.13

### TypeScript version

v4.4.3

### Additional context

Proposed Fix: Replace line 99 of `main.tsx` in the sandbox with `getSortedRowModel: getNewSortedRowModel()` and test out the steps above. The logic of the loop starting on line 85 of `getNewSortedRowModel.ts` in the sandbox is updated to iterate over subRows even if there is only one in case that subRow has it's own subRows.

Sandbox here: https://codesandbox.io/s/react-table-grouping-and-sorting-bug-fix-d4p776

I think we could make the logic a bit more complex to avoid diving into subRows more than necessary by also checking something like `row.leafRows.length <= 1`, but I think the current change is a bit simpler to read.

Also I am willing to open up a PR with a fix, but wasn't sure if I should just do that or open up this issue first so just let me know. Thanks!

### Terms & Code of Conduct

bdknox:
Fixes #4414

### DIFF
--- a/packages/table-core/src/utils/getSortedRowModel.ts
+++ b/packages/table-core/src/utils/getSortedRowModel.ts
@@ -90,10 +90,9 @@ export function getSortedRowModel<TData extends RowData>(): (
           // If there are sub-rows, sort them
           sortedData.forEach(row => {
             sortedFlatRows.push(row)
-            if (!row.subRows || row.subRows.length <= 1) {
-              return
+            if (row.subRows?.length) {
+              row.subRows = sortData(row.subRows);
             }
-            row.subRows = sortData(row.subRows)
           })
 
           return sortedData


KevinVandy (approved):
no semicolons ;)

PR #4495, opened 2022-10-31
[v8] Improve grouping performance on large clusters of values

jpcamara:
I maintain an application which renders a virtualized `react-table` table with up to 50k rows, 50 columns and up to 3 grouped columns. We were noticing huge performance problems with grouping and tracked it down to `[...previous, row]` in `groupBy` inside of `getGroupedRowModel`. This PR fixes that issue and adds a spec to verify the performance of that scenario.

* When grouping data that has large clusters of values, generating grouped columns becomes extremely slow. For instance: 50k rows, 3 grouped columns, each grouped column has an identical value (so each groups together 50k rows each) - takes ~30 seconds

* The performance bottleneck is `[...previous, row]` - it has to recreate the previous array over and over again, getting slower and slower each iteration. Using `.push` instead takes it from 30 seconds down to < 100ms to generate the groupings (overall table generation goes down to < 1 second).

* `push` is safe in this case because all of the data is constructed in the `groupBy` method so we are not mutating data from anywhere that would require generating new arrays using spread every iteration

### DIFF
--- a/packages/table-core/src/utils/getGroupedRowModel.ts
+++ b/packages/table-core/src/utils/getGroupedRowModel.ts
@@ -176,7 +176,7 @@ function groupBy<TData extends RowData>(rows: Row<TData>[], columnId: string) {
     if (!previous) {
       map.set(resKey, [row])
     } else {
-      map.set(resKey, [...previous, row])
+      previous.push(row)
     }
     return map
   }, groupMap)
--- a/packages/react-table/__tests__/makeTestData.ts
+++ b/packages/react-table/__tests__/makeTestData.ts
@@ -1,4 +1,4 @@
-import faker from 'faker'
+import faker from '@faker-js/faker'
 
 export type Person = {
   firstName: string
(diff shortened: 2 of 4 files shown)


packages/table-core/src/utils/getGroupedRowModel.ts
```diff
@@ -176,7 +176,7 @@ function groupBy<TData extends RowData>(rows: Row<TData>[], columnId: string) {
     if (!previous) {
       map.set(resKey, [row])
     } else {
-      map.set(resKey, [...previous, row])
```
jpcamara, line 179: In cases of large clusters of values, this operation gets more and more expensive every iteration - recreating 49k entries, recreating 49,001 entries, recreating 49,002 entries, etc

tannerlinsley:
Hot dang!!!

KevinVandy:
Just tested this out with my project with 200k rows and this makes all the difference. Thanks for this PR

jpcamara:
> Just tested this out with my project with 200k rows and this makes all the difference. Thanks for this PR

Awesome! Glad to hear it helped - also dang, 200k 👀 👌🏼 

On 50k rows I noticed our code doing cell dimension calculations in less than a second, but grouping was taking 60+ seconds on the same set of data. Turned out almost all time spent was in that small piece of code, glad it was a minor fix!

danalloway:
@jpcamara this is awesome, excellent work, thank you!!!

darkterminal:
Nice catch! Thanks for this PR :fire: @jpcamara

tyteen4a03:
Will this be backported to V7?

Issue #4440, opened 2022-10-10
[v8] vue-table: FlexRender component has invalid prop definitions


lrstanley:
### Describe the bug

`@tanstack/vue-table` currently defines it's `FlexRender` component like so:

```typescript
export const FlexRender = {
  props: ['render', 'props'],
  setup: (props: { render: any; props: any }) => {
    return () => {
      if (
        typeof props.render === 'function' ||
        typeof props.render === 'object'
      ) {
        return h(props.render, props.props)
      }

      return props.render
    }
  },
}
```

However, it's expected with Vue 3 that if not using `<script setup />` (or similar), that the component is wrapped in `defineComponent()` for proper type inference. Additionally, it looks as though without this, Volar (through VSCode), is accepting `string[]`, rather than `any`:

![](https://cdn.liam.sh/share/2022/10/Code_AqwWYxMVxP.png)

Looks like it might be picking the wrong `props` field from the defined component, as this is the generated `FlexRender` type from the `index.d.ts` file, and as you can see from the above screenshot, it thinks the `props` prop is actually the `props` field of the component that defines what props should be passed in:

```typescript
declare const FlexRender: {
    props: string[];
    setup: (props: {
        render: any;
        props: any;
    }) => () => any;
}
```

### Your minimal, reproducible example

https://codesandbox.io/s/github/tanstack/table/tree/main/examples/vue/basic?from-embed=&file=/src/App.vue

### Steps to reproduce

1. Take above CodeSandbox example (directly from tanstack table website), and open it in an editor that supports Vue component intellisense (in my case, VSCode with Volar).
2. Open `App.vue` and you should see that it is not expecting the appropriate type.

### Expected behavior

I would expect that it should pick up `any`, as the `@tanstack/vue-table` defines the `props` property on `FlexRender` without a type, and thus should be `any`.

### How often does this bug happen?

Every time

### Platform

- OS: Windows 11
- Browser: Firefox v106.0b9
- Editor: VSCode v1.71.2 (Volar v1.0.3)
- Vue: v3.2.40

### react-table version

v8.5.15

### TypeScript version

v4.8.4

### Additional context

If I mirror the `FlexRender` component into my application, and wrap it with `defineComponent`, it does seem to recognize that the `props` prop should actually be `any`:

```typescript
import { defineComponent, h } from "vue"

export const FlexRender = defineComponent({
  props: ["render", "props"],
  setup: (props: { render: any; props: any }) => {
    return () => {
[...]

PR #4508, opened 2022-11-08
Fixes issue with types in FlexRender

TgMrP:
Fixes issue with types in FlexRender
#4440


KevinVandy:
@Mokshit06 We don't have many Vue proficient table maintainers here. If you still pay attention to this repo, does this look good to merge?

Mokshit06:
Hey! This should fix the types, but since the `FlexRender` doesn't have any state and just renders based on props, this could also just be a functional component like this https://vuejs.org/guide/extras/render-function.html#functional-components

KevinVandy (approved):

PR #4516, opened 2022-11-12
[svelte] add types to renderComponent

ambiguous48:
New types `ComponentProps` and `ComponentType` were added in version [3.49](https://github.com/sveltejs/svelte/blob/master/CHANGELOG.md#3490) of svelte. We can use them to make `renderComponent` a generic function, where the `props` argument can reflect the given component's expected props.

Implementation taken from here: https://github.com/ivanhofer/sveltekit-typescript-showcase/blob/main/src/01-props/09-svelte-component/Component.svelte


KevinVandy:
@ambiguous48 @tannerlinsley It looks like the new types are causing typescript issues with the `renderComponent` method. The types need to be updated there too, or else the build will continue to fail

PR #4524, opened 2022-11-14
perf: convert types to interfaces

KevinVandy:
This PR re-implements much of @terry-au 's #4419  PR. However, it only focusses on the `type` to `interface` conversions


changyeamoon:
How much of a difference did it make?

PR #4533, opened 2022-11-15
Update README.md [add row spanning feature as todo]

Reda-ELOUAHABI:


KevinVandy:
I don't think we have any plans to add a row spanning feature at all right now.

PR #4550, opened 2022-11-22
Define separate steps in pr.yml

fekete965:
I thought it would be easier to see which step failed in the future if they are in a separate step.

### DIFF
--- a/.github/workflows/pr.yml
+++ b/.github/workflows/pr.yml
@@ -7,13 +7,24 @@ jobs:
     strategy:
       matrix:
         node: [16]
+
     steps:
-      - uses: actions/checkout@v2
-      - uses: actions/setup-node@v2
+      - name: Checkout
+        uses: actions/checkout@v2
+
+      - name: Install Node.js
+        uses: actions/setup-node@v2
         with:
           node-version: ${{ matrix.node }}
-      - run: |
-          npm install --legacy-peer-deps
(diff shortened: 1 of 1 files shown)


KevinVandy:
Not familiar with Github Actions as much as Gitlab or BitBucket, but wouldn't we have to set up a bunch of caches and artifacts between the steps to do this? I think we can already easily look at the console output to see while pipelines fail, so we might leave this alone?

Your work that you've contributed is highly appreciated though. I could be wrong and this might be better.

KevinVandy:
Actually this pipeline ran your modified pipeline just fine, probably will merge

KevinVandy (approved):

fekete965:
> Not familiar with Github Actions as much as Gitlab or BitBucket, but wouldn't we have to set up a bunch of caches and artifacts between the steps to do this? I think we can already easily look at the console output to see while pipelines fail, so we might leave this alone?
> 
> Your work that you've contributed is highly appreciated though. I could be wrong and this might be better.

Sorry for the late reply! Today's my wife's birthday 🎂 and I was out with her. :D

If you create multiple jobs, then each job will run on a different virtual machine (as far as I know).
If you do that and want to reuse certain data, you need to handle the cache somehow.

If you define steps inside a job, those will be run on the same VM one after another, so cache management is not needed. :)

PR #4552, opened 2022-11-22
feat: better esm support

DamianOsipiuk:
This PR should bring better ESM support for both older and newer bundlers.
I tried to replicate the setup that we have in https://github.com/TanStack/query.

There are few typescript issues in `packages/table-core/src/features/Filters.ts`

It would be ideal if people using different adapters could test this before integrationg to main branch.

### DIFF
--- a/babel.config.js
+++ b/babel.config.js
@@ -6,10 +6,13 @@ module.exports = {
   targets: 'defaults, not ie 11, not ie_mob 11',
   presets: [
     [
-      '@babel/env',
+      '@babel/preset-env',
       {
         loose,
         modules: false,
+        include: [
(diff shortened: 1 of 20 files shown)


tannerlinsley:
I’m fine rolling this out. If we get kickback from any broken builds we can investigate further.

DamianOsipiuk:
@KevinVandy I believe that was the old behavior, where `core` was also present in adapter builds for some reason. Which should not be the case anymore.

KevinVandy:
> @KevinVandy I believe that was the old behavior, where `core` was also present in adapter builds for some reason. Which should not be the case anymore.

Yep sorry, I needed to do a clean build to see only the new build files in `/lib`

PR #4553, opened 2022-11-23
fix: do not check for `browser` in package.json

DamianOsipiuk:
@KevinVandy Small fix to stop requiring `browser` field in `package.json` during deploy as it's breaking older webpack versions.

### DIFF
--- a/scripts/publish.ts
+++ b/scripts/publish.ts
@@ -361,7 +361,7 @@ async function run() {
       )
 
       await Promise.all(
-        (['module', 'main', 'browser', 'types'] as const).map(
+        (['module', 'main', 'types'] as const).map(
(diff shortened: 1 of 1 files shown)


KevinVandy:
Was just looking into this. Thanks

KevinVandy (approved):

KevinVandy:
I believe this should be merged as a feat: RELEASE_ALL so will do that

KevinVandy:
@DamianOsipiuk Is this actually related to the webpack "enhanced-resolve" bug? Could be temporary. Will merge this regardless I think

https://github.com/storybookjs/storybook/issues/19927#issuecomment-1324075415

PR #4558, opened 2022-11-24
Fix hooks error in default column cell renderer

MertHaddad:
React Hooks "useState" && "useEffect" are called in function "cell" that is neither a React function component nor a custom React Hook function. React component names must start with an uppercase letter. So I simply capitalized the C to fix this error.

### DIFF
--- a/examples/react/editable-data/src/main.tsx
+++ b/examples/react/editable-data/src/main.tsx
@@ -26,7 +26,7 @@ declare module '@tanstack/react-table' {
 
 // Give our default column cell renderer editing superpowers!
 const defaultColumn: Partial<ColumnDef<Person>> = {
-  cell: ({ getValue, row: { index }, column: { id }, table }) => {
+  Cell: ({ getValue, row: { index }, column: { id }, table }) => {
     const initialValue = getValue()
     // We need to keep and update the state of the cell normally
     const [value, setValue] = React.useState(initialValue)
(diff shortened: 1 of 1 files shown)


KevinVandy:
`cell` is just a function that can return JSX, not a component itself. There should be no error here. 

And this would be a huge breaking change throughout all of TanStack Table, as it is referenced as `cell` everywhere else in the api and codebase. 

You should still be able to run react hooks in other components defined in cell?

I do like the idea of cell, header, etc. being capitalized to indicate they can have component markup, but we can't just make this breaking change here.

MertHaddad:
You can still see this error in your example -line 32/40- [sandbox ](https://codesandbox.io/s/github/tanstack/table/tree/main/examples/react/editable-data?from-embed=&file=/src/main.tsx)

After I did the fix my code worked without any problems so I wanted to share my experience here incase anyone faces this issue in the future.
thanks

MafuyuRyan:
@MertHaddad Thanks Mert.

You can not see the error because have no lint check? I think that should be fixed, bad code.

mawnir:
@MertHaddad @KevinVandy 
i have a `columns` like this : `export const columns: ColumnDef<Post>[] = [{`

   ```
  {
        id: 'actions',
       cell: ({ row }) => {
            const { push } = useRouter();   <=======
            const editPost = (id: string) => {
                push(`/dashboard/posts/edit/${id}`);
            };
            return (....
```

i get this error : React Hook "useRouter" is called in function "cell" that is neither a React function component nor a custom React Hook function.
how can i fix this? can i pass the `id` value to the main component? 
thank you!

periakteon:
If there is anyone having this problem in 2024, check this out:

https://github.com/TanStack/table/discussions/4205

PR #4564, opened 2022-11-25
Update README.md: Add pnpm and yarn install scripts

forinda:
Add `pnpm` and `yarn` install script definitions


KevinVandy (approved):
sure

Issue #4479, opened 2022-10-24
Without StrictMode / In Production Mode onColumnSizingChange does not return any values during resize


bjoe87:
### Describe the bug

No values are returned when using onColumnSizing during the resize event when app is not wrapped with strictmode.
Provided sandbox is based of https://tanstack.com/table/v8/docs/examples/react/column-sizing 

Uncomment StrictMode at the bottom to get column resizing working.

### Your minimal, reproducible example

https://codesandbox.io/s/dazzling-lake-pr2n28?file=/src/main.tsx

### Steps to reproduce

1. go to https://codesandbox.io/s/dazzling-lake-pr2n28?file=/src/main.tsx
2. Try to resize any column

### Expected behavior

Column should resize. The values for onColumnSizingChange should be passed to the updater.

### Platform

- macOS Monterey: 12.6
- Chrome 106.0.5249.119

### react-table version

v8.5.15

### Terms & Code of Conduct

- [X] I agree to follow this project's Code of Conduct
- [X] I understand that if my bug cannot be reliable reproduced in a debuggable environment, it will probably not be fixed and this issue may even be closed.

talatkuyuk:
It is very strange. **The resizing feature** **has broken** in my nextjs app a week ago. I tried to figure it out. It screwed my several days.

Then I see this issue. Yes, `reactStrictMode: false` in the `next.config.js`.

Then I turned it into `true`, then resizing started to work. But I would like to keep it false, since I haven't migrated the useEffects in my app according to React18 yet.

bjoe87:
Was able to look into it today. The underlying problem is the newColumnSizing variable does not contain what gets calculated within the updater. The quick fix is to just move newColumnSizing outside of updateOffset function. (I confirmed this and it works) https://github.com/TanStack/table/blob/main/packages/table-core/src/features/ColumnSizing.ts#L229

rdecoito:
Looks like https://github.com/TanStack/table/pull/4606 reverted the above MR, but it doesn't seem as if there was any explanation why this happened. I'm having this problem in my production build, and I can replicate the problem easily with v8.7.3 (my project is using v8.7.0, so it's present on that version as well). I just tested it with a minimum reproducible example, and the issue was definitely fixed in v8.7.2 by the above MR, so I'd really love to know why the fix was reverted.

I spent a long time migrating our rather large codebase to v8 only to discover this bug once I pushed it up to our dev environment, so that's been rather frustrating.


[...]

PR #4573, opened 2022-11-29
Fix: onColumnSizingChange does not return correct value when react.strictmode is not used

bjoe87:
When onColumnSizingChange is used without react.strictmode / production the function does not return the correct updated values. This breaks the resize feature as discussed in https://github.com/TanStack/table/issues/4479

### DIFF
--- a/packages/table-core/src/features/ColumnSizing.ts
+++ b/packages/table-core/src/features/ColumnSizing.ts
@@ -218,7 +218,7 @@ export const ColumnSizing: TableFeature = {
             ? Math.round(e.touches[0]!.clientX)
             : (e as MouseEvent).clientX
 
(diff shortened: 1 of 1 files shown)


packages/table-core/src/features/ColumnSizing.ts
```diff
@@ -218,6 +218,8 @@ export const ColumnSizing: TableFeature = {
             ? Math.round(e.touches[0]!.clientX)
             : (e as MouseEvent).clientX
 
+          let newColumnSizing: ColumnSizingState = {}
```
igoryusha, line 221: Variable is not being reassigned, its better to use `const` instead.

KevinVandy (approved):

PR #4588, opened 2022-12-06
Render the actual component for svelte when returned from cell function

itssumitrai:
Currently the flexRender function checks if the result of `cell` function is a Svelte component, and if it is, it just returns the actual Component instead of rendering the component with the props. This PR fixes this issue

### DIFF
--- a/packages/svelte-table/src/index.ts
+++ b/packages/svelte-table/src/index.ts
@@ -55,7 +55,7 @@ export function flexRender(component: any, props: any) {
     const result = component(props)
 
     if (isSvelteComponent(result)) {
-      return result
+      return renderComponent(result, props)
     }
 
     return wrapInPlaceholder(result)


KevinVandy:
Would this be a breaking change?

vd-aloker:
I'd like to see this merged. 

It's rather a bugfix than a breaking change. 


Here's a usecase of mine:
```
const columns = [
	display({
		id: 'mail',
		cell: () => EmailColumn,
		header: () => L.colMail()
	})
];
//...
<td>
	<svelte:component this={flexRender(cell.column.columnDef.cell, cell.getContext())} />
</td>
```
Without the fix, the EmailColumn component won't receive any props. With the fix applied, the EmailColumn component will receive all props from getContext().

The only current workaround is to use the component itself as the value of cell:

```
display({
	id: 'mail',
	cell: EmailColumn,
	header: 'E-Mail'
}),
```

But this will a) cause a Typescript error and b) disable the possibility to switch components based on values passed to the cell function.

bfovez:
This does seem to me like a major problem. I am currently investigating for a headless Svelte table after been relatively disappointed by [Svelte Headless Table](https://svelte-headless-table.bryanmylee.com/) (instabilities, ), but that issue is a deal breaker.
I'm surprised that nobody from the team reacted since last december. The fix seems to be pretty straightforward.

KevinVandy (approved):

KevinVandy:
Taking @vd-aloker 's word for it

KevinVandy:
This PR seems to be causing some issues in #4778 . Do the examples just need to be updated? @itssumitrai  @vd-aloker @bfovez 

Otherwise this PR may need to be reverted.

Eric-Arz:
This completely broke TypeScript in Svelte. Please revert this.

thenbe:
I was trying to understand why this PR is needed.

This is how I rendered components before this PR and how I *still* continue to do so. However, this PR caused a type [issue](https://github.com/TanStack/table/issues/4778) that requires a [workaround](https://github.com/TanStack/table/issues/4778#issuecomment-1517680029). I'm just wondering, what did we gain in return? What am I missing?

```typescript
import { renderComponent } from '@tanstack/svelte-table';

const columns = [
	display({
		id: 'mail',
		cell: () => renderComponent(EmailColumn), // without props
		cell: () => renderComponent(EmailColumn, { myprop: 'foo' }), // with props (type-safe)
	}),
];
```

KevinVandy:
I've just been waiting for someone to make a PR to fix it correctly. If it is as simple as reverting this, we can.

PR #4590, opened 2022-12-07
fix: throw with error message for `getColumn`

Dev-CasperTheGhost:
This PR includes:

- Throw with error message in prod/dev when a column is not found. This makes it easier to debug.


KevinVandy:
In discussions in other PRs, we were considering changing this to return `undefined` instead of throwing an error

KevinVandy:
Closing this in favor of https://github.com/TanStack/table/pull/4678

Issue #4593, opened 2022-12-08
[Row Grouping bug] - Rows grouped by a field with object like value makes only one row group - @tanstack/react-table


lucabrini:
### Describe the bug

Grouping enabled for a column, whose cells' value is an object type, makes only one big row group, even if each cell's value is different from the other's one.

### Your minimal, reproducible example

https://codesandbox.io/s/row-grouping-bug-sn3fys

### Steps to reproduce

Just open the codesanbox above

### Expected behavior

Should groups together only rows with the same object.  

### How often does this bug happen?

Every time

### Platform

- OS : Void Linux, i3 WM, Picom
- Browser : Firefox & Brave Browser

### react-table version

v8.7.0 & v8.5.13

### TypeScript version

4.8.4

### Terms & Code of Conduct

- [X] I agree to follow this project's Code of Conduct
- [X] I understand that if my bug cannot be reliable reproduced in a debuggable environment, it will probably not be fixed and this issue may even be closed.

KevinVandy:
I think our recommendation is to not provide objects as a value to columns that need sorting, filtering, or grouping. You can instead provide an `accessorFn` that parses a complex row object into a primitive value that these features can use.

PR #4600, opened 2022-12-10
fixed bug: Rows grouped by a field with object like value makes only …

briinki:
Fixed bug #4593


KevinVandy:
I'm not sure if JSON.stringify is always reliable here? The best solution is to use a proper `accessorFn`. Will need further consideration if we should support this.

briinki:
Alright, thanks for your time! We can close it here :)

PR #4620, opened 2022-12-23
fix: return a null value for undefined nested keys

SeanStove:
Since PR #4407 undefined nested keys throw an error message.
This might be helpful, but does not solve the issue when you have to work with data which looks like

```json
[
  { "firstName": "John", "address": { "street": "Main street 1" } },
  { "firstName": "Jane", "address": null }
]
```

```accessor``` ```address.street``` in this case will throw the error in dev.
(or will break the application later in production in case the data here is not as complete)

So the suggestion from @dennemark in issue #4499 makes sense to me.
This PR will return a ```null``` value, if the nested value does not exist.
Optionally, there could also be a ```console.warn``` statement with what used to be the error message.

Fixes: #4499

### DIFF
--- a/packages/table-core/src/core/column.ts
+++ b/packages/table-core/src/core/column.ts
@@ -52,7 +52,7 @@ export function createColumn<TData extends RowData, TValue>(
         let result = originalRow as Record<string, any>
 
         for (const key of accessorKey.split('.')) {
-          if (!(key in result)) {
+          if (result == null || !(key in result)) {
             if (process.env.NODE_ENV !== 'production') {
               console.warn(
                 `"${key}" in deeply nested key "${accessorKey}" returned undefined.`
--- a/packages/table-core/src/core/column.ts
+++ b/packages/table-core/src/core/column.ts
@@ -53,6 +53,11 @@ export function createColumn<TData extends RowData, TValue>(
 
         for (const key of accessorKey.split('.')) {
           if (!(key in result)) {
+            if (process.env.NODE_ENV !== 'production') {
+              console.warn(
+                `"${key}" in deeply nested key "${accessorKey}" returned undefined.`
+              )
+            }
             return null
           }
 
--- a/packages/table-core/src/core/column.ts
+++ b/packages/table-core/src/core/column.ts
@@ -52,12 +52,11 @@ export function createColumn<TData extends RowData, TValue>(
         let result = originalRow as Record<string, any>
 
         for (const key of accessorKey.split('.')) {
-          result = result[key]
-          if (process.env.NODE_ENV !== 'production' && result === undefined) {
-            throw new Error(
-              `"${key}" in deeply nested key "${accessorKey}" returned undefined.`
-            )
+          if (!(key in result)) {
+            return null
           }
+
+          result = result[key]
         }
 
         return result

### CODE REVIEW

**Review by KevinVandy — COMMENTED** [review]
I think we do need to do this, but let's also keep a console.warn with that error message in dev mode

**Comment by SeanStove** [comment]
> I think we do need to do this, but let's also keep a console.warn with that error message in dev mode

@KevinVandy added the warning message.

**Comment by kachun333** [comment]
is it possible for this change to be merged, please?

**Comment by SeanStove** [comment]
I noticed the ```key in result``` will fail if the object is null, so I've added an additional check for that.

So it should now handle these cases:

``` { "firstName": "Jane", "address": { "city": "New York"  } }```

``` { "firstName": "James", "address": null }```

``` { "firstName": "Jimmy" }```

when trying to access ```address.street```

File: packages/table-core/src/core/column.ts [hunk 1]
```diff
@@ -52,12 +52,16 @@ export function createColumn<TData extends RowData, TValue>(
         let result = originalRow as Record<string, any>
 
         for (const key of accessorKey.split('.')) {
-          result = result[key]
-          if (process.env.NODE_ENV !== 'production' && result === undefined) {
-            throw new Error(
-              `"${key}" in deeply nested key "${accessorKey}" returned undefined.`
-            )
+          if (result == null || !(key in result)) {
```
KevinVandy, line 55: should this be "===" triple equals equality check? and actually, wouldn't we want to check for undefined? Why is the result assignment moved to the last line?

packages/table-core/src/core/column.ts
```diff
@@ -52,12 +52,16 @@ export function createColumn<TData extends RowData, TValue>(
...
+            if (process.env.NODE_ENV !== 'production') {
+              console.warn(
+                `"${key}" in deeply nested key "${accessorKey}" returned undefined.`
+              )
+            }
+            return null
           }
+
+          result = result[key]
```
KevinVandy, line 64: shouldn't this still be above the if statement?

KevinVandy (changes requested):

packages/table-core/src/core/column.ts
SeanStove: I used "==" to handle both undefined and null, because you cannot use the ```in``` operator on either of them. Unless you want to handle undefined and null differently?

packages/table-core/src/core/column.ts
SeanStove: The result assignment is moved to the last line because you cannot read properties of null/undefined (which is covered by the if statement above it).

packages/table-core/src/core/column.ts
KevinVandy: @SeanStove I'm working on a PR where I'm simply making this change. Wonder if it works in your testing

```ts
accessorFn = (originalRow: TData) => {
        let result = originalRow as Record<string, any>

        for (const key of accessorKey.split('.')) {
          result = result?.[key]
          if (process.env.NODE_ENV !== 'production' && result === undefined) {
            console.warn(
              `"${key}" in deeply nested key "${accessorKey}" returned undefined.`
            )
          }
        }

        return result
      }
```

KevinVandy:
This code was good, but cleaned up a little in this another PR. Closing this in favor of #4678

packages/table-core/src/core/column.ts
SeanStove: @KevinVandy Tested and it covers the 3 cases I described above. Thanks for this.

KevinVandy: Shipped this last week if you hadn't seen that yet

PR #4654, opened 2023-01-12
Update Example: react/virtualized-rows

dgildeh:
I did this update out of frustration trying to find examples of doing this with the V3 version of react-virtual and also figuring out how to virtualize the table using the Window scrollbar (useWindowVirtualizer).

The example has been updated to switch between two versions of the virtualized table, one a fixed height container as in the original example, and the other using the Window Scrollbar so users can scroll down the page:

* Updated to use latest version of @tanstack/react-virtual: 3.0.0-beta.33
* Migrated existing example to fixedHeightTable.tsx
* Added windowHeightTable.tsx with example of how to render page length virtualized table using useWindowVirtualizer
* A few other fixes that were causing TypeScript errors in my IDE


examples/react/virtualized-rows/src/fixedHeightTable.tsx
```diff
@@ -0,0 +1,145 @@
...
+  const virtualRows = rowVirtualizer.getVirtualItems()
+  const totalSize = rowVirtualizer.getTotalSize()
+
+  // This is where the magic happens, essentially create a large row with height of total rows
+  // before and after the first/last displayed row so the scroll bar works correctly
+  const paddingTop = virtualRows.length > 0 ? virtualRows?.[0]?.start || 0 : 0
+  const paddingBottom =
+    virtualRows.length > 0
+      ? totalSize - (virtualRows?.[virtualRows.length - 1]?.end || 0)
+      : 0
```
piecyk, line 66: We can keep those values together

```suggestion
  const [paddingTop, paddingBottom] =
    virtualRows.length > 0
      ? [
          Math.max(0, virtualRows[0].start),
          Math.max(0, rowVirtualizer.getTotalSize() - virtualRows[virtualRows.length - 1].end),
        ]
      : [0, 0]
```

examples/react/virtualized-rows/src/fixedHeightTable.tsx
```diff
@@ -0,0 +1,145 @@
...
+        <tbody>
+          {paddingTop > 0 && (
+            <tr>
+              <td style={{ height: `${paddingTop}px` }} />
+            </tr>
+          )}
+          {virtualRows.map(virtualRow => {
+            const row = rows[virtualRow.index] as Row<Person>
+            return (
+              <tr key={row.id}>
```
piecyk, line 122:
```suggestion
              <tr key={virtualRow.key}>
```

examples/react/virtualized-rows/src/fixedHeightTable.tsx
```diff
@@ -0,0 +1,145 @@
...
+  })
+
+  const tableContainerRef = React.useRef<HTMLDivElement>(null)
+
+  const { rows } = table.getRowModel()
+  const rowVirtualizer = useVirtualizer({
+    count: rows.length,
+    getScrollElement: () => tableContainerRef.current,
+    estimateSize: () => 54,
+    overscan: 10,
```
piecyk, line 54: Let's we need to pass correct keys to virtualizer it's important when rows change position
```suggestion
    overscan: 10,
    getItemKey: React.useCallback(index => rows[index].id, [rows])
```

examples/react/virtualized-rows/src/fixedHeightTable.tsx
```diff
@@ -0,0 +1,145 @@
...
+  })
+
+  const virtualRows = rowVirtualizer.getVirtualItems()
+  const totalSize = rowVirtualizer.getTotalSize()
```
piecyk, line 58: Moved to inline calculation of paddingBottom

examples/react/virtualized-rows/src/fixedHeightTable.tsx
```diff
@@ -0,0 +1,145 @@
...
+  SortingState,
+  useReactTable,
+} from '@tanstack/react-table'
+import { useVirtualizer } from '@tanstack/react-virtual'
+
+import { Person } from './makeData'
+
+type FixedHeightTableProps = {
+  // The data to render
+  data: any
```
piecyk, line 17: Why any?

examples/react/virtualized-rows/src/windowHeightTable.tsx
```diff
@@ -0,0 +1,147 @@
...
+  React.useLayoutEffect(() => {
+    tableOffsetRef.current = tableContainerRef.current?.offsetTop ?? 0
+  }, [])
+
+  const { rows } = table.getRowModel()
+  const rowVirtualizer = useWindowVirtualizer({
+    count: rows.length,
+    scrollMargin: tableOffsetRef.current,
+    estimateSize: () => 54,
+    overscan: 10,
```
piecyk, line 54: Add getItemKey

examples/react/virtualized-rows/src/windowHeightTable.tsx
```diff
@@ -0,0 +1,147 @@
...
+      : 0
+
+  return (
+    <div
+      ref={tableContainerRef}
+      className="container"
+      style={{
+        // Need to add 52px to totalSize to account for header height + container border top and bottom
+        height: rowVirtualizer.getTotalSize() + 52,
+      }}
```
piecyk, line 75: This should not be needed, if header is sticky then it size should be added to scrollMargin

examples/react/virtualized-rows/src/windowHeightTable.tsx
```diff
@@ -0,0 +1,147 @@
...
+          {paddingTop > 0 && (
+            <tr>
+              <td style={{ height: `${paddingTop}px` }} />
+            </tr>
+          )}
+          {virtualRows.map(virtualRow => {
+            const row = rows[virtualRow.index] as Row<Person>
+            return (
+              <tr
+                key={row.id}
```
piecyk, line 123:
```suggestion
                key={virtualRow.key}
```

examples/react/virtualized-rows/src/windowHeightTable.tsx
```diff
@@ -0,0 +1,147 @@
...
+  const virtualRows = rowVirtualizer.getVirtualItems()
+
+  // We need to add a padding row at top that grows as we scroll down to ensure
+  // the table scrolls down with the page
+  const paddingTop =
+    virtualRows.length > 0
+      ? virtualRows?.[0]?.start
+        ? virtualRows?.[0]?.start - tableOffsetRef.current
+        : 0 || 0
+      : 0
```
piecyk, line 66: We can use almost same logic as in, just here we need to remove also scrollMargin as we use it 

```suggestion
 const [paddingTop, paddingBottom] =
    virtualRows.length > 0
      ? [
          Math.max(0, virtualRows[0].start - rowVirtualizer.options.scrollMargin),
          Math.max(0, rowVirtualizer.getTotalSize() - virtualRows[virtualRows.length - 1].end),
        ]
      : [0, 0]
```

piecyk (changes requested):
Overall looks good, added few comments, please update also file names as those exporting components via convention should be pascal case.

examples/react/virtualized-rows/src/fixedHeightTable.tsx
dgildeh: Updated but had to add some extra logic to remove undefined errors in TS

examples/react/virtualized-rows/src/fixedHeightTable.tsx
dgildeh: Done

examples/react/virtualized-rows/src/fixedHeightTable.tsx
dgildeh: Added, but had to update callback to get rid of type errors

examples/react/virtualized-rows/src/fixedHeightTable.tsx
dgildeh: I removed totalSize as your combined paddingTop/paddingBottom calculation doesn't use it

examples/react/virtualized-rows/src/windowHeightTable.tsx
dgildeh: Fixed

examples/react/virtualized-rows/src/windowHeightTable.tsx
dgildeh: I removed it - in my original version I wasn't using paddingBottom, now that that's added the container scales to the correct height as needed.

examples/react/virtualized-rows/src/windowHeightTable.tsx
dgildeh: Added

dgildeh:
Fixed all comments and updated filenames to PascalCase.

KevinVandy:
It would be cool to still get fixed height and window examples into the docs. Closing for now because of merge conflicts. Can work with you if you are interested, albeit, a year later.
I started a whole [virtualization guide](https://tanstack.com/table/v8/docs/guide/virtualization) in the table docs that can definitely use these concepts

PR #4665, opened 2023-01-18
fix: do not crash when sorting state has non existing columns

targos:
We have very dynamic columns in our project (they can be added, sorted, then removed based on the data manipulated by the user).
It would be unnecessarily complicated and error prone to have to reset the sorting everywhere we make changes that may affect the existing columns.

### DIFF
--- a/packages/table-core/src/utils/getSortedRowModel.ts
+++ b/packages/table-core/src/utils/getSortedRowModel.ts
@@ -19,6 +19,7 @@ export function getSortedRowModel<TData extends RowData>(): (
 
         // Filter out sortings that correspond to non existing columns
         const availableSorting = sortingState.filter(sort =>
+          table._getAllFlatColumnsById()[sort.id] &&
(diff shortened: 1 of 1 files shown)


packages/table-core/src/utils/getSortedRowModel.ts
```diff
@@ -19,6 +19,7 @@ export function getSortedRowModel<TData extends RowData>(): (
 
         // Filter out sortings that correspond to non existing columns
         const availableSorting = sortingState.filter(sort =>
+          table._getAllFlatColumnsById()[sort.id] &&
```
KevinVandy, line 22: Think we can just add some optional chaining here

KevinVandy (changes requested):

KevinVandy:
Closing this in favor of #4678

targos:
Thanks, I like what was done in #4678

PR #4698, opened 2023-02-03
fix sorting with undefined values

### LINKED ISSUE #4289: Sorting state behavior is changed if column data includes undefined values
### Describe the bug

In some cases, when an undefined cell value is present in a column, the sorting toggle for that column doesn't follow the `asc-desc-undefined` pattern; instead, it may go `asc-undefined`, `asc-desc`, `asc-undefined-desc-undefined`, and a multitude of other variations.

I'm not sure _exactly_ what's causing the bug to appear, but I did manage to get a somewhat simple repro case. I think this issue occurs when using manual sorting (whether or not passing in the `manualSorting: true` flag), and when the first row has an undefined value for a column. Interestingly, the problem disappears if you set `sortDescFirst: false` on the column with the undefined value.

https://codesandbox.io/s/fragrant-framework-2noe5r?file=/src/main.tsx

### Steps to reproduce

1. Click on the `lastName` header, note that sort goes into `desc` mode
2. Click on the `lastName` header again, note that sort goes into undefined sort mode
2. Click on the `lastName` header once more, note that sort goes back into `desc` mode and skips `asc`

### Expected behavior

As a user, I expect that clicking on a sortable header would toggle between the `asc-desc-undefined` states (or reverse, if `sortDescFirst`).

Every time

### Screenshots or Videos

Aug-08-2022 16-46-39

### Platform

- OS: Mac
- Browser: Chrome

### react-table version

v8.5.11

### TypeScript version

v4.7.4

### Additional context

_No response_

### Terms & Code of Conduct

SeanStove:
Issue #4289 describes a sorting issue with undefined values.

Right now in the ```getAutoSortDir``` function, the value of the first row is used to determine the (first) sort direction. If it's a string, sorting is ```asc```, otherwise ```desc```.

The problem with this is:
If the value of the first row happens to be ```undefined```, than the ```getAutoSortDir``` will return ```desc``` because of the ```typeof value === 'string'``` check.

So the items are sorted descending (putting the undefined ones at the bottom) after the first sort click event.

When you click sorting again, the ```getNextSortingOrder``` function will get called and ```getAutoSortDir``` function will be called again, but now the first row is no longer an undefined value, but a string, so it will return ```asc```.

Making the following statement evaluate to ```true```

https://github.com/TanStack/table/blob/423d46ecce8d1f3f02c78e5d833b54112c21136c/packages/table-core/src/features/Sorting.ts#L298-L304

and thus returning ```false``` when called in the ```setSorting``` function, making it remove the sorting state

https://github.com/TanStack/table/blob/423d46ecce8d1f3f02c78e5d833b54112c21136c/packages/table-core/src/features/Sorting.ts#L235-L238

So what I've done now is iterated over all rows to verify if the type of the column is a string.
I'm not so sure about this implementation **as this will for sure impact performance** (if the column is for example a number, than all rows will be iterated before hitting the final return statement...).
A performance optimization could be to also check for additional types (date, number...) so the function can return early.

I noticed getAutoSortingFn uses .slice(10), not sure where this number comes from?
The same could be used in getAutoSortDir, but the issue will of course still occur if you have more than 10 rows with undefined values.

https://github.com/TanStack/table/blob/423d46ecce8d1f3f02c78e5d833b54112c21136c/packages/table-core/src/features/Sorting.ts#L139

This technique of using the value of the first row is for example also used to determine the filter function, so this might cause issues here as well

https://github.com/TanStack/table/blob/423d46ecce8d1f3f02c78e5d833b54112c21136c/packages/table-core/src/features/Filters.ts#L207

Fixes #4289

### DIFF
--- a/packages/table-core/src/features/Sorting.ts
+++ b/packages/table-core/src/features/Sorting.ts
@@ -162,13 +162,15 @@ export const Sorting: TableFeature = {
 
         return sortingFns.basic
       },
-      getAutoSortDir: () => {
-        const firstRow = table.getFilteredRowModel().flatRows[0]
+      getAutoSortDir: () => {        
+        const rows = table.getFilteredRowModel().flatRows
 
-        const value = firstRow?.getValue(column.id)
+        for (const row of rows) {
+          const value = row?.getValue(column.id)
 
-        if (typeof value === 'string') {
-          return 'asc'
+          if (typeof value === 'string') {
+            return 'asc'
+          }
         }
 
         return 'desc'


KevinVandy:
Maybe with a max of 10 rows being checked, we could do this. Also consider you can the `sortDescFirst` and `sortUndefined` column options manually yourself if your data has null or undefined values.

SeanStove:
Closing in favor of ```sortDescFirst``` and ```sortUndefined```

KevinVandy:
> Closing in favor of ```sortDescFirst``` and ```sortUndefined```

Did those solve all of the use cases you have?

SeanStove:
> > Closing in favor of `sortDescFirst` and `sortUndefined`
> 
> Did those solve all of the use cases you have?

It solves mine, but won't solve the case metioned in the issue.

PR #4699, opened 2023-02-03
fix: Column hidden function maxDepth calculation error

cfj1996:


KevinVandy:
Hey, could you give an example for what this fixes?

cfj1996:
> 嘿，你能举个例子说明这个修复了什么吗？
https://codesandbox.io/p/sandbox/intelligent-nova-gobffy
When the Info->More Info->visits column is hidden, getHeaderGroups().length should be equal to 2, not always 3. getHeaderGroups().length doesn't change with columns hidden

cfj1996:
> Hey, could you give an example for what this fixes?

https://codesandbox.io/p/sandbox/intelligent-nova-gobffy
When the Info->More Info->visits column is hidden, getHeaderGroups().length should be equal to 2, not always 3. getHeaderGroups().length doesn't change with columns hidden

PR #4705, opened 2023-02-11
feat: vue pagination controlled example

radusuciu:
Thanks for supporting Vue! I wanted to use this in a project but didn't find much material for Vue past the basic examples, so I thought I'd take a crack at contributing an example to start.

Things I'd like to draw your attention to:
1. I'm not sure if `setPagination` is correctly implemented
2. The "Go to page" input's value is tracked with a separate `ref`. This is slightly different from the react implementation
3. I'm using https://jsonplaceholder.typicode.com/posts for the API call. I can simulate data using `faker` if that is preferred

You can also run this example on CodeSandbox: https://codesandbox.io/p/sandbox/musing-https-uvewmd


KevinVandy:
modified this PR into #4775

PR #4706, opened 2023-02-12
Fix documentation for ColumnMeta redeclration

Oigen43:
Example from the doc doesn't work, meta still has any type. Here is a fix of redeclartion.


radusuciu:
I also am a bit confused about how to add properties either to `ColumnMeta` or to the `ColumnDef` in Vue 3 -- for example for adding a custom `headerClassName`, but this doesn't look right to me.. why are we declaring a module called `@tanstack/table-table`?

KevinVandy:
pretty sure this change is wrong? No such package as `@tanstack/table-table`

PR #4728, opened 2023-02-24
chore(svelte-table): cleanup package.json

benmccann:
Svelte only supports ESM, so there's no reason to list CJS or UMD files in the `package.json`


KevinVandy:
@lachlancollins closing this as we seem to have made the cleanup necessary over the last week. However, wondering the reason for type: commonJS we have now?

benmccann:
I'm not sure if you're working on the `main` branch or another branch, but none of these changes have been made on `main` so perhaps this PR should be reopened.

KevinVandy:
I'm cleaning up stale PRs. Your change makes sense to me, but @lachlancollins has done some major repo build maintenance over the past week which seems to have the opposite approach. I hope he can weigh in here.

benmccann:
I just reviewed the recent PRs and don't think they're opposite, but rather are unrelated. You can read about the reason for explicitly specifying `"type": "commonjs"` here: https://publint.dev/rules#use_type. Adding an explicit `type` field is good practice, but shouldn't stop us from changing it to one more compatible with the Svelte ecosystem.

We can wait for Lachlan to chime in and confirm his reason for setting it was what I mentioned, but I do think it'd be good to reopen this so that we don't lose track of it.

lachlancollins:
@benmccann is absolutely correct, this package should be switched to ESM. It might be the case that we do this when we switch to using `@tanstack/config` over Rollup (v9?) to avoid breaking changes.

KevinVandy:
closing, as the svelte package is already going under a massive refactor in other PRs

PR #4753, opened 2023-03-13
Grouping value accessors

tombuntus:
This PR is WIP.

It adds optional columnDef options `groupingValueAccessorKey` and `groupingValueAccessorFn` to enable custom values to be used for row grouping.

It also adds a default behaviour to group rows when the row-grouping column is a column group.


tombuntus:
closed in favour of #4833

PR #4760, opened 2023-03-17
Fix issue toggling row selection for a grouped row

### LINKED ISSUE #4759: Cannot toggle selection of grouped row to deselect its children
### Describe the bug

I have a table with grouping and selection enabled. I would like to achieve the following behavior:
- Non-grouped rows are selectable; selection is toggled via a checkbox on the row
- Grouped rows have selection checkboxes, but they don't have their own selection state
    - the checkbox is checked if all sub rows are checked
    - the checkbox is not checked is all subrows are unchecked
    - the checkbox is indeterminate if some sub rows are checked
    - interactively toggling the checkbox manually selects/deselects all sub rows

To get this behavior, I've configured my table as follows:
- `enableRowSelection`: function that returns `true` for non-grouped rows and `false` for grouped rows
- `enableMultiRowSelection`: true
- `enableSubRowSelection`: true
- state of the checkbox for grouped rows is updated programmatically using `row.getIsAllSubRowsSelected()` and `row.getIsSomeSelected()`
- interaction with checkbox for rows (both grouped and non-grouped) calls `row.toggleSelected(newCheckboxState)`

Problem:
Interactively trying to uncheck a grouped row's checkbox does nothing when it should deselect all the rows beneath it. The problem is that the grouped row will always return `false` from `row.getIsSelected()`. Calling `toggleSelection(false)` on this grouped row, which should deselect all the sub rows, returns early because it detects that the row is already not selected so there is nothing to be done to deselect it.

https://codesandbox.io/p/sandbox/crimson-river-2ysx73

### Steps to reproduce

In the code sandbox:
1. Expand a group
2. Click the selection checkbox for the expanded group to mark it as selected
3. See that all the subrows were selected (correct)
4. Click the selection checkbox for the expanded group again to mark it as not selected
5. Notice that nothing happened. The group and subrows remained selected when the desired behavior was for the group and subrows to all become not selected.

### Expected behavior

I expected to create a grouped row whose selection is based only on the selection of its subrows. Clicking the checkbox associated with that grouped row toggles the selection for all of its subrows.

Every time

### Screenshots or Videos

_No response_

### Platform

Not platform specific

### react-table version

8.7.9

### TypeScript version

_No response_

### Additional context

As a note, I also went down an implementation path where I had `enableRowSelection` set to `true` for both grouped rows and non-grouped rows. However, this also did not lead to the behavior I wanted. The primary problem I encountered was that I didn't want a group's selection state to be part of TanStack's row selection state because the group's selection state is always determined programmatically by the state of its subrows.

### Terms & Code of Conduct

mollykreis:
Resolves #4759 

Toggling a grouped row can have up to two effects:
1. Update the selected state of the grouped row if that grouped row supports selection
2. Update the selected state of subrows if `getCanSelectSubRows` is `true`

`toggleSelected()` returned early if the selected state of the row wasn't going to change without any consideration of whether the selected state of subrows also needed to be updated. With this change, `toggleSelected()` now considers both if the rows selection state will change and if subrow selection state might change before returning early.

I've added some selection tests to `table-core` -- both a few basic tests and some tests that cover the issue I am fixing.

### DIFF
--- a/packages/table-core/src/features/RowSelection.ts
+++ b/packages/table-core/src/features/RowSelection.ts
@@ -353,7 +353,8 @@ export const RowSelection: TableFeature = {
         table.setRowSelection(old => {
           value = typeof value !== 'undefined' ? value : !isSelected
 
-          if (isSelected === value) {
+          const willUpdateSubRows = row.subRows?.length && row.getCanSelectSubRows()
+          if (!willUpdateSubRows && isSelected === value) {
             return old
           }
 
--- a/packages/table-core/__tests__/getSelectedRowModel.test.ts
+++ b/packages/table-core/__tests__/getSelectedRowModel.test.ts
@@ -1,0 +1,122 @@
+import { ColumnDef, getCoreRowModel, getGroupedRowModel, RowSelectionState, Table, Updater } from '../src'
+import { createColumnHelper } from '../src/columnHelper'
+import { createTable } from '../src/core/table'
+import { makeData, Person } from './makeTestData'
+
+type personKeys = keyof Person
+type PersonColumn = ColumnDef<Person, string | number | Person[] | undefined>
+
+function generateColumns(people: Person[]): PersonColumn[] {
+  const columnHelper = createColumnHelper<Person>()
+  const person = people[0]
+  return Object.keys(person).map(key => {
+    const typedKey = key as personKeys
+    return columnHelper.accessor(typedKey, { id: typedKey })
+  })
+}
+
+describe('row selection', () => {
+  let table: Table<Person>
+  let data: Person[]
+
+  beforeEach(() => {
+    data = makeData(10)
+    const columns = generateColumns(data)
+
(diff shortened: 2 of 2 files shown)


KevinVandy:
Unsure if this change is needed as discussed here: https://github.com/TanStack/table/issues/4759#issuecomment-1483985451

packages/table-core/src/features/RowSelection.ts
```diff
@@ -353,7 +353,8 @@ export const RowSelection: TableFeature = {
         table.setRowSelection(old => {
           value = typeof value !== 'undefined' ? value : !isSelected
 
-          if (isSelected === value) {
+          const willUpdateSubRows = row.subRows?.length && row.getCanSelectSubRows()
```
2jordan3, line 356: **row.subRows?.length** can be undefined, how about using !! to explicitly convert it to a boolean?

PR #4766, opened 2023-03-20
Introducing nx for caching

ZackDeRose:
<div align="left">
      <a href="https://www.youtube.com/watch?v=ewA42CaCF8s" target="blank">
         
      </a>
</div>

## What it does

+ Mainly a like-for-like for the following package scripts: `build`, `test:ci`, `typecheck`
  + These three commands will still be called as they have been (e.g.: `npm run build`) and have the same effect. They have been optimized to add intelligent caching so we never need to run the same command on the same code twice!
  + (`build` was also optimized to run it's `rollup` and `tsc` commands in parallel, whereas before they were in sequence)
+ The `pr` workflow file was updated - it will still run the same checks as it had before, but it will now run all checks in parallel (as opposed to in sequence)
+ Adding `nx` means that we can use nx for task running which gives us some nice things - for example we can use the command: `nx run-many --target=build` to run all build tasks in the workspace, which includes all examples.
  + this could be a nice addition to add to the pr.yml file, so we can assert none of the examples broke by the PR's changes
  + intelligent caching has been added here as well (see the video for a demo!)

### DIFF
--- a/.github/workflows/pr.yml
+++ b/.github/workflows/pr.yml
@@ -20,11 +20,5 @@ jobs:
       - name: Install Dependencies
         run: npm install --legacy-peer-deps
 
-      - name: Build the Project
-        run: npm run build
-
-      - name: Running Typechecks
-        run: npm run typecheck
-
-      - name: Running Tests
-        run: npm run test
+      - name: Run Checks
+        run: nx run-many --targets=test:ci,build-js,typecheck --projects=root
--- a/.github/workflows/pr.yml
+++ b/.github/workflows/pr.yml
@@ -21,4 +21,4 @@ jobs:
         run: npm install --legacy-peer-deps
 
       - name: Run Checks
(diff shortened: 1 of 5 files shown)


ZackDeRose:
Missed this in the video, but looks like we need a dependency on from `build-js` (the `rollup` script), so it runs before we can run our tests.

I added that to the `project.json` file:

```json
{
  "name": "root",
  "targets": {
    // ...
    "test:ci": {
      "command": "vitest",
      "inputs": ["packageSources", "{workspaceRoot}/packages/*/__tests__/**/*"],
      "dependsOn": ["build-js"] // <== just added this line
    }
  }
}

```

With the timings (and `typecheck` being the long stick), the result is this "pipeline" in the `pr` script:

Screen Shot 2023-03-19 at 10 11 09 PM

PR #4803, opened 2023-04-08
fix: handle null/undef values when flexRendering function-type component in svelte-table

madisonscott:
This handles the case when a column is defined like:
```
{
  id: 'columnId',
  header: 'Header',
  accessorKey: 'property',
},
```
but item[accessorKey] is null or undefined.


dit7ya:
If this PR is not going to be accepted, can someone please document what is the workaround for handling `null` values in Svelte?

madisonscott:
friendly bump @KevinVandy :)

KevinVandy (approved):

PR #4805, opened 2023-04-11
state, onStateChange and renderFallbackValue are optional

marceloverdijk:
https://github.com/TanStack/table/discussions/4804


KevinVandy (changes requested):
This seems to introduce a lot of type errors down below

PR #4808, opened 2023-04-11
Adding nx cloud

ZackDeRose:
## What it does:

Adds Nx cloud to allow us to skip unnecessary tasks in our pr workflow file

+ Adds the nx cloud package and runner
+ Updated the public key (see nx.json) to a read-only key (best practice for OSS project)
+ Added a CI token to be used in the pr.yml workflow file
  + @KevinVandy DM me your email and I can add you to the github cloud org so you can add the secret ci token to the github secrets for this project.
  + (once added, we can also add [the nx cloud github app](https://github.com/apps/nx-cloud) to this project so we can get auto-updates on PRs to see task status)

### DIFF
--- a/.github/workflows/pr.yml
+++ b/.github/workflows/pr.yml
@@ -1,5 +1,9 @@
 name: pr
 on: [pull_request]
+env:
+  NX_DAEMON: false
+  NX_VERBOSE_LOGGING: true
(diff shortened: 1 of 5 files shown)


KevinVandy:
being implemented in a different pr

Issue #3748, opened 2022-03-07
react-table support to readonly data


Pavinati:
### Discussed in https://github.com/TanStack/react-table/discussions/3648

<div type='discussions-op-text'>

<sup>Originally posted by **Pavinati** January 18, 2022</sup>
Seems like the table data prop is forced to be mutable. is it required to be mutable or is there a way around this?

```typescript
type SomeRecord = Readonly<....>;

const columns: Column<SomeRecord>[] = [...];

type Props {
  className?: string;
  data: readonly SomeRecord[];
}

const MyTable = ({ className, data }: Props) => {
  return <Table className={className} columns={columns} data={data} />;
};
```
results in
```
The type 'readonly Readonly<SomeRecord>[]' is 'readonly' and cannot be assigned to the mutable type 'Readonly<SomeRecord>[]'.  TS4104
```</div>

PR #4829, opened 2023-04-26
narrow options type to not ask for mutability

schicks:
This changes the `CoreOptions` type to expect a `readonly TData[]`. The result is that readonly data can be passed into table options. Because it is a type only change, no runtime functionality should be affected.

This has been brought up as an [issue](https://github.com/TanStack/table/issues/3748) and as a [question](https://github.com/TanStack/table/discussions/3648#discussioncomment-2083028) in the past. As long as there are no library functions that _return_ `CoreOptions`, this should strictly allow more usages of the library.

### DIFF
--- a/examples/react/basic/src/main.tsx
+++ b/examples/react/basic/src/main.tsx
@@ -19,7 +19,7 @@ type Person = {
   progress: number
 }
 
-const defaultData: Person[] = [
+const defaultData: readonly Person[] = [
   {
     firstName: 'tanner',
     lastName: 'linsley',
--- a/examples/react/basic/src/main.tsx
+++ b/examples/react/basic/src/main.tsx
@@ -79,7 +79,7 @@ const columns = [
 ]
 
 function App() {
-  const [data, setData] = React.useState(() => [...defaultData])
+  const [data, setData] = React.useState<readonly Person[]>(() => [...defaultData])
   const rerender = React.useReducer(() => ({}), {})[1]
 
   const table = useReactTable({
--- a/packages/table-core/src/core/table.ts
+++ b/packages/table-core/src/core/table.ts
@@ -63,7 +63,7 @@ const features = [
 export interface CoreTableState {}
 
 export interface CoreOptions<TData extends RowData> {
-  data: TData[]
+  data: readonly TData[]
   state: Partial<TableState>
   onStateChange: (updater: Updater<TableState>) => void
   debugAll?: boolean
--- a/packages/table-core/src/utils/getCoreRowModel.ts
+++ b/packages/table-core/src/utils/getCoreRowModel.ts
@@ -22,7 +22,7 @@ export function getCoreRowModel<TData extends RowData>(): (
         }
 
         const accessRows = (
-          originalRows: TData[],
+          originalRows: readonly TData[],
           depth = 0,
           parentRow?: Row<TData>
         ): Row<TData>[] => {


comp615:
Closes #3748, #3648

schicks:
Hi @KevinVandy , could I get some context on why this was closed? it seems like a nonbreaking change to me that allows for the library to be used with more types without casting.

KevinVandy:
I think it can be common to have mutable data, especially when using/making editing features. I think we would introduce a need for even more type casting with this change, and it could be a breaking change for some.

schicks:
I don't think that's true. Because this adds the readonly restriction to an _argument_, it is safe to pass in mutable data without casting; the restriction just declares that the library is not going to mutate the data internally. See [this playground](https://www.typescriptlang.org/play?#code/MYewdgzgLgBAZgVzMKBLcAVAFgQyhnAawFMIAlYnAE3ABsBPAQTCoqgQCdIBZYqavDhgBeGAB4AIoIA0APgAUVQQC4YHSjTAMYU-gG0AugEoRsmEv4A6WsTABzKFgBQT0JFgBbBPwBGNxhwcOPQA8nDYqPYQqgDeOKrQHJF2AL6GIjB6TjA5MHGqAOQQIB58WMkFKdLZufkwBcS0EMSVTgYuTgD0nTCosKgQMBA4cMQwUCAwwDi0tDBCiMho4PNgYCD8UMRU84P8hMlqGnT0MADufVgwXr425oIAhE6LKOhg2HgEJOTHWkwsbE4PD4An48huOD8xACQVC4XKUSMQA) for proof that this doesn't prevent people from passing in mutable data without casting.

schicks:
@KevinVandy just wanted to check in on this one last time. I really don't think there is any situation in which this would require a client of the library to cast or break their existing usage, but if there is one I'm happy to work on changing this so that it doesn't occur. Can you point me in the right direction?

homburg:
As @schicks says @KevinVandy (sorry if I'm piling on).

This should not require any new casts for the users of the library as mutable data can be assigned to a readonly type.

Ie.
`Array<number>` and `ReadonlyArray<number>` is assignable to `ReadonlyArray<number>`

but

`ReadonlyArray<number>` is not assignable to `Array<number>`

See simple example here

https://www.typescriptlang.org/play?ssl=14&ssc=1&pln=15&pc=1#code/GYVwdgxgLglg9mABCAzgUwCoEMBGAbNAITWDgCc0AeAESyiwD4AKOAB1gRQC5EBvRACZ0sPAIJkyWAJ41hDRAF8AlHwUAoUJA5JUmXAVHAoaMrPrM227n0HCeAJTRYBCPFPGSZtc4pW91amoQnFCIALYg9Pho3liIALyIANoAjAA0iABMGQDMALoA3GoA9MWI5YgAegD8gcFgKKEUzq5SsQnJ6Vm5eYhYKIj1jUWlFVW1gaPEpBRqutjR0+RoTLxC9DwRUQSxynPoCwRLFKvrIojNLmBuuyqjGFKsaIgmZOQAhJNlhsZk+3rRH4mU52cKRfQxYR7eYQoFkEEbC5OK43KFKIA

```
function useTableBefore<Data>(options: { data: Array<Data> }) {}
function useTableAfter<Data>(options: { data: ReadonlyArray<Data> }) {}


const mutableData = [1, 2, 3];
//    ^?

const readonlyData = [1, 2, 3] as const;
//    ^?


// Before
useTableBefore({data: mutableData})
useTableBefore({data: readonlyData}) // Type error!

// After
useTableAfter({data: mutableData})
useTableAfter({data: readonlyData})
```

kdawgwilk:
Agreed with the above comments that this is not a breaking change but supports better interactions with other libraries that use more strict types such as `readonly`.  I am seeing bad practices where consumers of this library are now making copies of their data array to satisfy the compiler when they shouldn't 😬

deammer:
Please reconsider this feature 🙏

rsslldnphy:
Just run into this myself. My ORM returns read-only types, which is great for all of my application code except for react table, which requires me to either type-cast to a mutable type or make a copy of the data.

PR #4831, opened 2023-04-27
Export type `FeatureOptions`

fw6:
Export type `FeatureOptions` for external features to extend interface


KevinVandy:

PR #4832, opened 2023-04-28
feat: Support external features

fw6:
Provide new option `features: TableFeature[]` to extend table core features

### DIFF
--- a/packages/table-core/__tests__/customFeature.ts
+++ b/packages/table-core/__tests__/customFeature.ts
@@ -1,0 +1,36 @@
+/**
+ * @module extensions/features/range-selection
+ * @reference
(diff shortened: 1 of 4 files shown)


KevinVandy:
We are actively looking into adding this, doing more research if anything else needs to be considered here.

KevinVandy:
The next version of TanStack table will have way better support for this more than likely following a similar pattern, but a lot more needs to be done, especially for the TS support

KevinVandy:

KevinVandy:
@fw6  This is finally implemented in v8.14.0 https://tanstack.com/table/latest/docs/guide/custom-features

PR #4833, opened 2023-04-29
Feat get grouping value

tombuntus:
Adds an option to column definitions to allow the user to specify a custom value to be used for row grouping.


KevinVandy:
Thinking about the name `getGroupingValue`, because it slightly collides with the `groupingValue` value on a row, but not really, since this is a columnDef option.

KevinVandy (approved):

PR #4836, opened 2023-05-02
Update getFilteredRowModel.ts

ravshansbox:
Avoid mapping `null`/`undefined` `columnFilters`


KevinVandy:
Why would column filters ever be null?

ravshansbox:
1) We are already doing this in line 31
2) I'm not yet sure how, but I had this variable null in a real project already

ravshansbox:
image
image

ravshansbox:
So, it's `undefined`, not `null`

Issue #4809, opened 2023-04-12
enableGrouping not working on table level


yotamberk:
### Describe the bug

Defining `enableGrouping: false` on table options doesn't do anything.

I want to default all columns to disable grouping and only enabling grouping when `enableGrouping: true` is defined in column.

### Your minimal, reproducible example

https://codesandbox.io/s/busy-kirch-f9prci?file=/src/main.tsx

### Steps to reproduce

I defined the table options with  `enableGrouping: false` and the column I want to allow grouping with  `enableGrouping: true`

### Expected behavior

Only the the enabled column has grouping enabled.

### How often does this bug happen?

None

### Platform

all

### react-table version

8.8.2

### Terms & Code of Conduct

- [X] I agree to follow this project's Code of Conduct
- [X] I understand that if my bug cannot be reliable reproduced in a debuggable environment, it will probably not be fixed and this issue may even be closed.

KevinVandy:
this was fixed back in pr #4843

PR #4843, opened 2023-05-08
Fix: getCanGroup always resolving to true

mozzerbob:
Small update to getCanGroup. It would always resolve to true and ignore the rest of the statement.

Fixing issue #4809.


step2341:
@mozzerbob i can see the fix above, im using "@tanstack/react-table": "^8.10.3", and still having this issue. has this fix been released?

mozzerbob:
> @mozzerbob i can see the fix above, im using "@tanstack/react-table": "^8.10.3", and still having this issue. has this fix been released? @KevinVandy i noticed you have this option in your materialTable, how did you get it to work correctly?

It never got merged, its waiting for a maintainer approval.

tannera-cargill:
Is it possible to get this merged? This function always returns true, even when enableGrouping is false.

KevinVandy (approved):

PR #4854, opened 2023-05-11
docs: Modify basic sorting algorithm description

shongs27:
Basic sorting algorithm description is incorrect

`a>b` and `b<a` are the same case. 
if `a>b` condition is not right, you shoud judge `a<b` condition


tychenjiajun:
Still incorrect, see https://github.com/TanStack/table/pull/5134

PR #4882, opened 2023-05-26
fix: assign parentId to subRows when creating aggregatedRows

### LINKED ISSUE #4881: parentId or getParentRow is undefined for grouped data (flat data + grouping column)
### Describe the bug

parentId and getParentRow returns undefined for leafRows that are grouped by a given column

https://codesandbox.io/s/row-grouping-forked-pg5nco

### Steps to reproduce

1. Go to https://codesandbox.io/s/row-grouping-forked-pg5nco
2. Expand/Collapse the "single" status
3. Notice the parentId is `undefined` and the `getParentRow` in the console is undefined too

### Expected behavior

I expected the parentId to reference the "virtual" aggregatedGroupedRow tanstack table creates internally.

Every time

### Screenshots or Videos

Screenshot 2023-05-26 at 11 19 05 AM

### Platform

- OS: all
- Browser: all

### react-table version

v8.9.1

### TypeScript version

N/A

### Additional context

I think because tanstack table is creating a "virtual/fake" row for the grouping, that row id should be assigned to its subRows so we don't lose its parenting references

### Terms & Code of Conduct

csantos1113:
Fixes #4881

### DIFF
--- a/packages/table-core/src/utils/getGroupedRowModel.ts
+++ b/packages/table-core/src/utils/getGroupedRowModel.ts
@@ -118,6 +118,10 @@ export function getGroupedRowModel<TData extends RowData>(): (
               })
 
               subRows.forEach(subRow => {
+                if (!subRow.parentId) {
+                  Object.assign(subRow, { parentId: id });
+                }
+
                 groupedFlatRows.push(subRow)
                 groupedRowsById[subRow.id] = subRow
                 // if (subRow.getIsGrouped?.()) {


packages/table-core/src/utils/getGroupedRowModel.ts
```diff
@@ -118,6 +118,10 @@ export function getGroupedRowModel<TData extends RowData>(): (
               })
 
               subRows.forEach(subRow => {
+                if (!subRow.parentId) {
+                  Object.assign(subRow, { parentId: id });
```
KevinVandy, line 122: Unfortunately, this also introduces a bug where the table will crash after ungrouping by a column. Needs to be looked into more, but this will be part of the solution.

KevinVandy (changes requested):

packages/table-core/src/utils/getGroupedRowModel.ts
csantos1113: Good catch Kevin. I was so focused on the solution for my current use-case that I forgot about ungrouping 😢

KevinVandy: Looking into it, we have to find a way to remove the parentId as we ungroup.

csantos1113: I'm trying to navigate this code, trying to follow the same approach for how the `.depth` value is assigned/unassigned.. But I see that property being modified in here:

https://github.com/TanStack/table/blob/46b160dfd37917e6cc1f8e53c091345574a0fae4/packages/table-core/src/utils/getGroupedRowModel.ts#L38

So, I have couple questions:

1. What is the difference between https://github.com/TanStack/table/blob/46b160dfd37917e6cc1f8e53c091345574a0fae4/packages/table-core/src/utils/getGroupedRowModel.ts#L38-L41 and the section I touched on this PR? mostly, they seem the same, but not sure why they happen in two diff places
2. Would that the better place to assign/unassign the `parentId` 🤔?

csantos1113: 😞 well, `depth` is also wrong

When we ungroup the column, they all stay `depth: 1`;

image

 but they should go back to `depth: 0` which is the default value when ungrouped 🤕

image

csantos1113: For what is worth:

assigning/un-assigning `depth` in v7 works just fine, here an example: https://codesandbox.io/s/amazing-water-pcepcv?file=/src/App.js

KevinVandy:
This is still an important bug fix. Will be looking into how to fix the ungrouping crash

csantos1113:
> This is still an important bug fix. Will be looking into how to fix the ungrouping crash

I'm sorry I let this PR stale. Work has been insane last months, that I haven't had the time to work on any of this

packages/table-core/src/utils/getGroupedRowModel.ts
```diff
@@ -118,6 +118,10 @@ export function getGroupedRowModel<TData extends RowData>(): (
               })
 
               subRows.forEach(subRow => {
+                if (!subRow.parentId) {
```
chatpong-v, line 121: Thank for update

chatpong-v (commented):
To update

chatpong-v (commented):
To update

PR #4895, opened 2023-06-03
feature: RowPicking

rasgo-cc:
This adds a `RowPicking` feature. It has essentially the same logic as `RowSelection` but keeps the picked rows in a separate state. 
Decided to create this PR after opening this discussion: https://github.com/TanStack/table/discussions/4892
It includes **code**, **docs** and an **example**.

- When is this useful?

Imagine you have an app which imports a CSV file. A table displays all the data that is about to be imported. You can select a row so the app can show you the issues related to it somewhere else in the UI. You can unpick rows (assuming they'd be initially all picked) to skip them from being imported. So selecting and picking rows serve different purposes.

I think having the option to pass custom features (https://github.com/TanStack/table/pull/4832) would be great but figured this could also make sense to have as a core feature. PS: I really need this in my app :)


KevinVandy:
Hey, thanks for the effort, but we are very conscious about bundle size and not introducing a lot of repeated code. After looking at this, it seems that it is almost an exact copy of the row selection feature, so devs could just use the row selection feature to achieve almost the same thing. Or just use some self managed state.

And yes, we (and @tannerlinsley) should probably get around to perfecting, merging, and releasing #4832 so that we can support 3rd party features and wrappers to TanStack Table even better than the composition already possible.

I myself have gone through implementing a Row Pinning Feature that also falls in the same category of probably not being quite useful enough for the bundle size increase it will bring along. But we should be able to install it properly as a plugin in the future.

rasgo-cc:
Thanks for the feedback @KevinVandy . I really need this for now so I guess my best option will be to use my forked repo. I went with the "self managed state" approach first, as you mention, but quickly gave up on it. The `@tanstack/table`'s API is really **super** convenient (coming from v7, and loving it) and I think it's just better DX to keep all features close to its core, so I get both `getIsPicked` and `getIsSelected` through the same API. Again, even though the code is essentially the same, a row being picked and/or selected mean different things in my app. Bundle size is not that important in my case as it's a desktop/electron app. 

Having said that, looking forward to #4832 ! That would totally suit my needs as well.

Thanks again for you time.

KevinVandy:
@rasgo-cc FYI, check out v8.14.0, where you can take the code from this PR and still use it in your codebase. You can now pass in your own custom `_features`
https://tanstack.com/table/latest/docs/guide/custom-features

rasgo-cc:
This is awesome. 

> @rasgo-cc FYI, check out v8.14.0, where you can take the code from this PR and still use it in your codebase. You can now pass in your own custom `_features` https://tanstack.com/table/latest/docs/guide/custom-features

Awesome! Thank you @KevinVandy. Will give it a try / refactor code soon I hope.

rasgo-cc:
@KevinVandy just tried it and it works great, very smooth integration. Thank you!

PR #4909, opened 2023-06-09
fix: properly mark type of arrays that are not modified by the librar…

pcorpet:
…y as readonly

Fixes #3648


KevinVandy:
I have not really seen the need to mark all of these types as readonly, but if conflicts are resolved, we can reconsider.

pcorpet:
I'm willing to resolve the conflicts, I've pushed my branch again after rebasing to #5452

PR #4955, opened 2023-07-10
Add name to `FlexRender` component

phijma-leukeleu:
Inside the Vue devtools, I think this is currently an `Anonymous Component`. This would give it the correct name, right?


KevinVandy:
If this is still important, can we open a new PR without conflicts? Sorry for the delay/inconvience.

PR #4958, opened 2023-07-11
fix getLeafHeaders

bug-brain:
I have noticed that `table.getLeafHeaders()` contained headers that had more headers in `subHeaders` and thereby are not leaves. Looking at the code I found this part where the header is added to the array in any case.

The method is used for the column sizing feature but `yarn dev` gave me no errors (could not figure out `yarn test:dev`). I hope this helps anyway or at least points you in the right direction to fix this small issue.

### DIFF
--- a/examples/react/basic/package.json
+++ b/examples/react/basic/package.json
@@ -9,7 +9,7 @@
     "start": "vite"
   },
   "dependencies": {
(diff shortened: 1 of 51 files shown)


eichelberger-c:
Hi I am also finding that getLeafHeaders returns headers that have subheaders. Is there any update on this bug?

PR #4959, opened 2023-07-11
Create custom example : Pagination- with page numbering,Delete action for every column and custom class to data cells of particular column

ObaidAshiq:
Showing page numbering to navigate through different pages at the bottom (Main focus)
Added custom column action (deleting of the row)
Used Dynamic classes to style data cells - not suited(checks every data cell in table)
Screenshot 2023-07-11 192649

Note : Add your own modules.css file


ObaidAshiq:
I have updated this example i would be closing it and provide the updated example later on

PR #4978, opened 2023-07-16
examples: vue row selection

ivolkoff:
This example shows how to set up row selection with the vue adapter.


KevinVandy:

Issue #4483, opened 2022-10-26
useResizeColumns doesn't work with React Portal in new window


JustJarethB:
### Describe the bug

When using react-table in a newly created window via ReactDOM.createPortal, column resizing does not work. From investigating, I believe this is because `useResizeColumns` references `document` to add event listeners, which refers to the original window running the code. This means the listener is on the wrong window's document. a local reference such as 
```
draggableElement.closest('html')?.parentNode as Document;
```
should work, but may not be the most 'correct'

### Your minimal, reproducible example

https://codesandbox.io/s/recursing-varahamihira-359y3m?file=/src/App.js:742-752

### Steps to reproduce

create new window
use window as Portal container
place resizable table inside portal

### Expected behavior

I expect column resizing to work in a new window but the table does not respond to the mouse event.

### How often does this bug happen?

Every time

### Screenshots or Videos

https://user-images.githubusercontent.com/31932204/198026153-cffbc447-71be-4c36-a8c1-ac4d3ee6b0c9.mov

### Platform

macOS (12.2.1) - Chrome & Safari

### react-table version

7.7.0 & 7.8.0

### TypeScript version

4.8.3

### Terms & Code of Conduct

- [X] I agree to follow this project's Code of Conduct
- [X] I understand that if my bug cannot be reliable reproduced in a debuggable environment, it will probably not be fixed and this issue may even be closed.

JustJarethB:
Interestingly AG Grid does work for this, but we have other requirements that mean we can not use AG. From digging around I think they are using [ResizeObserver](https://developer.mozilla.org/en-US/docs/Web/API/ResizeObserver) which seems very sensible. Is this something that could be considered for TanStack? I don't know how that would work with the library being headless

KishoreS-pamp:
This is because tanstack tables need to have context of where exactly the resize is happening.
you have to do something like this
` const handleMouseDown = useCallback((e) => {
    // Pass the event's ownerDocument so TanStack attaches its mousemove/mouseup
    // listeners to the correct document — works for both main and detached windows.
    getResizeHandler(e.target?.ownerDocument)(e);
  }, [getResizeHandler]);`

in this way we can say whether it is happening in the main or portal. This worked for me!

PR #5003, opened 2023-07-28
Fix column resize listener document context

JustJarethB:
Suggested fix for https://github.com/TanStack/table/issues/4483

Allows implementers to provide an alternative document for event listeners to be attached to in relation to the column resize extension. This fixes the use case that a new window has been opened but the table js code is running in the previous window.

### DIFF
--- a/packages/table-core/src/features/ColumnSizing.ts
+++ b/packages/table-core/src/features/ColumnSizing.ts
@@ -187,7 +187,7 @@ export interface ColumnSizingHeader {
    * @link [API Docs](https://tanstack.com/table/v8/docs/api/features/column-sizing#getresizehandler)
    * @link [Guide](https://tanstack.com/table/v8/docs/guide/column-sizing)
    */
-  getResizeHandler: () => (event: unknown) => void
+  getResizeHandler: (context?: Document) => (event: unknown) => void
   /**
    * Returns the current size of the header.
    * @link [API Docs](https://tanstack.com/table/v8/docs/api/features/column-sizing#getsize)
@@ -319,7 +319,7 @@ export const ColumnSizing: TableFeature = {
 
       return 0
     }
-    header.getResizeHandler = () => {
+    header.getResizeHandler = (contextDocument = document) => {
       const column = table.getColumn(header.column.id)
       const canResize = column?.getCanResize()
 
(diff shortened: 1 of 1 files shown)


KevinVandy:
If this PR is remade with no conflicts, we can merge it.

JustJarethB:
Rebased to main and cherry-picked relevant code. Attempting to re-open to pick up the latest change but will create a new PR if not

JustJarethB:
@KevinVandy thanks for the comment. As per my above comment, I pulled fresh and reset my branch to mirror. The current diff now seems correct. Please let me know if anything is missing or incorrect.

Cheers

KevinVandy (approved):

antoniobeltran:
@JustJarethB Thanks for proposing a fix. I'm wondering how you passed in the document on your sandbox example from bug reported. I've been messing around trying to get this to work for me.

`{headerGroup.headers.map((header) => {
									return <TableHeader
				
										key={header.id}
										{...getTableHeaderProps(header)}
										onMouseDown={header.getResizeHandler(popOutContainer)}
										onTouchStart={header.getResizeHandler(popOutContainer)}
									/>
								})}`

JustJarethB:
@antoniobeltran happy to help! It's tricky to be certain without the context of your codebase, but I'll try to give a generally helpful answer. 
Just by reading the semantic variable names, I assume `popOutContainer` is a reference to an `HTMLElement` in your new window (e.g. div, span...). You should pass the `Document` reference here rather than an `HTMLElement`.

A minimum (TS) example using your above snippet:

```
const TableHeadComponentRenderedInANewWindow = () => {
	const ref = useRef<HTMLDivElement>(null); //Can be any HTMLElement
	const popoutDocument = ref.current?.ownerDocument;
	return (
		<div ref={ref}>
			{headerGroup.headers.map((header) => {
				return <TableHeader key={header.id} {...getTableHeaderProps(header)}
					onMouseDown={header.getResizeHandler(popoutDocument)}
					onTouchStart={header.getResizeHandler(popoutDocument)} />
			})}
		</div>
	)
}
```

your type of `popoutDocument` in this example should be `Document | undefined` which matches the expected parameter value of `getResizeHandler`. In the scenario where `ref.current` is null, the default behaviour with `document` will be used, but this should not be the case when the component is rendering.

**For your specific case, it could be as simple as replacing `popOutContainer` with `popOutContainer.ownerDocument`**

Hope this helps :)

antoniobeltran:
Thanks @JustJarethB . It looks about to be exactly what we tried, but we didn't swap popOutContainer with popOutContainer.ownerDocument. We will try that out. Thank you

Issue #4994, opened 2023-07-26
Incorrect pagination behavior with autoResetPageIndex set to true when deleting rows


robertjvardy:
### Describe the bug

Currently working on a table with pagination and a display column that removes the current row.
- The data for the table is stored in react state and passed into the data prop.
- The display column performs a removal by filtering the current row out of the state.
Accordingly, autoResetPageIndex needs to be set to false to prevent the pageIndex state from resetting on every row delete.

Everything works well until all of the rows are deleted on the last page.
The pageIndex state doesnt update, even though the getPageCount response does.
I would assume that even with autoResetPageIndex set to false, the pageIndex state should update when data is deleted and the page count decreases.

### Your minimal, reproducible example

will provide on request

### Steps to reproduce

Steps To preproduce

1) Create a table that:
- uses pagination
- has a display column to remove the current row
- has autoResetPageIndex set to false
2) navigate to the last page
3) remove all of the rows by clicking the delete buttons one by one

### Expected behavior

Expected:
When the last row is removed the table state should decrease the pageIndex state to match the page count

Actual:
pageIndex is unchanged and there is no data shown since the pageIndex is out of range relative to the new number of pages

### How often does this bug happen?

None

### Platform

issue happens on any OS with any browser

### react-table version

8.5.15

### Terms & Code of Conduct

- [X] I agree to follow this project's Code of Conduct
- [X] I understand that if my bug cannot be reliable reproduced in a debuggable environment, it will probably not be fixed and this issue may even be closed.

p-viral:
I hope this gets fixed soon. Not sure if others are using this or not, but this is a real blocker for me.

PR #5031, opened 2023-08-19
Fix an issue where pageIndex becomes invalid when data is changed (removed) and autoResetPageIndex is false.

hamidvirani:
This fixes issue #4994 

The pageIndex now gets set to correct value, if invalid, when autoResetPageIndex is false and data changes

### DIFF
--- a/packages/table-core/src/features/RowPagination.ts
+++ b/packages/table-core/src/features/RowPagination.ts
@@ -224,14 +224,24 @@ export const RowPagination: TableFeature = {
       }
 
       if (
-        table.options.autoResetAll ??
-        table.options.autoResetPageIndex ??
+        table.getState().pagination &&
         !table.options.manualPagination
       ) {
         if (queued) return
         queued = true
         table._queue(() => {
-          table.resetPageIndex()
+          if (
+            table.options.autoResetAll ??
+            table.options.autoResetPageIndex
+          ) {
+            table.resetPageIndex()
+          } else {
+            const currentPageIndex = table.getState().pagination.pageIndex
+            const lastPageIndex = table.getPageCount() - 1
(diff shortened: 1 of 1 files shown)


KevinVandy:
I like this! I have to get around it with an useEffect in my tanstack table code bases like this:

https://github.com/KevinVandy/mantine-react-table/blob/433b24f761e10e4268039763479e8b415286334f/packages/mantine-react-table/src/hooks/useMRT_Effects.ts#L50

Will look into merging this fix

hamidvirani:
Hi,

Do you think this PR can be merged?

p-viral:
Any news on when this will be merged? 

I am hoping to get this working to release app to production. This is a blocker if pagination does not reset when the last row is deleted in the last page and the table does not reset to the previous page.

Please help.

KevinVandy:
> Any news on when this will be merged?
> 
> I am hoping to get this working to release app to production. This is a blocker if pagination does not reset when the last row is deleted in the last page and the table does not reset to the previous page.
> 
> Please help.

You can always just do this with your own code. Any fix here would just be a quality of life update.

hamidvirani:
@KevinVandy I have resolved the conflicts if we want to merge this PR

Naofal-Helal:
> I am hoping to get this working to release app to production. This is a blocker if pagination does not reset when the last row is deleted in the last page and the table does not reset to the previous page.

If anyone else is facing this issue and this is not yet merged, here is the workaround for React:

```typescript
  useEffect(() => {
    const currentPageIndex = table.getState().pagination.pageIndex;
    const lastPageIndex = table.getPageCount() - 1;
    if (currentPageIndex > lastPageIndex) {
      table.setPageIndex(lastPageIndex);
    }
  }, [data.length]);
```

PR #5037, opened 2023-08-22
fix: fix type of state in tableoptions

alisajadih:


sledgehammer999:
`TableState` has all its members marked as optional. So the usefulness of `Partial<TableState>` is questionable.

alisajadih:
No it isn't. `useReactTable` has `state` property with `Partial<TableState>` type. it also has `onStateChange` prop with `(updater: Updater<TableState>) => void` type. so if i want to pass `state` and `setState` with type `TableState` or `Partial<TableState>` to `useReactTable`, typescript will be unhappy. because type of `state` prop and `onStateChangeProp` are incompatible.

KevinVandy:
I think this is a misunderstanding of the table.setState API. It does require full state, not partial.

PR #5048, opened 2023-09-01
Increase core performance by 30%-50% 

retnag:
I replaced the Object.assign with direct mutation in create[Row,Cell,Column,Header,Table]. No features were changed, I did minimal touch, tests passed.

Object.assign is very performance Heavy, when compared with direct assignment, and can easily be replaced with direct mutations in the creational functions. Since these get called *a lot* when the data changes, especially with large amounts of data, the performance impact is quite huge. According to my measurements in my application, the performance of calculating or re-calculating the RowModels (this affects all of them) increased between 30%-50%, I measured the performance using the built-in debug features with tables of rowSize 100, 1.000, 10.000, 20.000, 50.000. The improvement becomes noticeable after about a 1,000 rows.

ref on performance of Object assign, vs mutation: https://www.pedaldrivenprogramming.com/2020/08/be-careful-with-immutability-in-javascript/

Short explanation of the changes: I changed the call signature of the create methods, they all return void now, and mutably modify the base object that was passed in, instead of creating and returning a new one, which gets Object.assigned onto the base Object later. I know this might have been intentionally designed this way because we all love pure functions, but as far as I can tell, functionality stays the same, and I would argue that 30-50% of performance gains is worth it here.

By the way, I love the modular design that's going on here, the code is really great to work with <3.

### DIFF
--- a/packages/table-core/src/core/cell.ts
+++ b/packages/table-core/src/core/cell.ts
@@ -52,14 +52,11 @@ export function createCell<TData extends RowData, TValue>(
   }
 
   table._features.forEach(feature => {
-    Object.assign(
-      cell,
-      feature.createCell?.(
-        cell as Cell<TData, TValue>,
-        column,
-        row as Row<TData>,
-        table
-      )
+    feature.createCell?.(
+      cell as Cell<TData, TValue>,
+      column,
+      row as Row<TData>,
+      table
     )
   }, {})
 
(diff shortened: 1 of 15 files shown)


KevinVandy:
@retnag This is great work. It might take a couple days to verify everything before shipping though. Good job.

KevinVandy (approved):

PR #5077, opened 2023-09-18
fix: allow empty string as column id

artchen-db:
Empty string can be a valid property name in row data objects, it should be allowed as column id. The existing check of `!id` still makes sense for other falsy values that are invalid.

### DIFF
--- a/packages/table-core/src/core/column.ts
+++ b/packages/table-core/src/core/column.ts
@@ -68,7 +68,7 @@ export function createColumn<TData extends RowData, TValue>(
     }
   }
 
-  if (!id) {
+  if (!id && id !== '') {
     if (process.env.NODE_ENV !== 'production') {
       throw new Error(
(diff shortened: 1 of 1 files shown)


KevinVandy:
Why should an empty string be allowed for a column id?

artchen-db:
For examples, I could be displaying an CSV file with a column of empty header, or I could be displaying query result of `SELECT "";`. These result in an empty string column id and row data like:

```javascript
{
  "": "some value",
  "anotherCol": "anotherVal"
}
```

KevinVandy:
AccessorKeys and IDs can be different. I suggest keeping the accessorKey as "", but providing a valid id column option with a value.

artchen-db:
@KevinVandy We did so as a workaround for now. Just curious what's the concern of allowing empty string as column id?

PR #5107, opened 2023-10-03
exemples: add `exemple` with `custom jsx column` and `pagination `and…

Iyadchafroud:
add exemple with **custom JSX component  columns** , pagination, filters and virtualized rows


KevinVandy:
I think we might take this PR branch and use it to update the 2 existing virtualized examples if that's alright. Been meaning to do that for a while. I'll be pushing to it over the weekend.

Iyadchafroud:
That is all right , but i dolt understand what is mean for for a while

Iyadchafroud:
can you add this exemple

PR #5123, opened 2023-10-10
Add test to showcase bug in `alphanumeric` sort.

matsimitsu:
Add test to showcase bug in `alphanumeric` sort.

The new implementation attempts to parse the string as a float, but does not check if the _entire_ string was parsed as a float. 

This causes issues trying to sort any alphanumeric strings that contain non-numbered characters (e.g. date-formatted strings).

Full description here: https://github.com/TanStack/table/issues/5124

One could check that the entire string was parsed as a float to fix this new behavior, e.g.:

```
  const aFloat = parseFloat(aStr);
  const bFloat = parseFloat(bStr);
  if (
    !isNaN(aFloat) && aFloat.toString() === aStr &&
    !isNaN(bFloat) && bFloat.toString() === bStr
  ) {
    return compareBasic(aFloat, bFloat)
  }
```

But at some point the amount of checks to trigger the short-cut could very well be more expensive than the gains this short-cut delivers.

### DIFF
--- a/packages/table-core/__tests__/Sorting.test.ts
+++ b/packages/table-core/__tests__/Sorting.test.ts
@@ -6,41 +6,43 @@ import {
 } from '../src'
 
 describe('Sorting', () => {
-    describe('alphanumeric sort booleans ascending', () => {
-      it('should return rows in correct ascending order', () => {
-        const data = [
-          { value: false },
-          { value: true },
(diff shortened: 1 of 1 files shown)


KevinVandy:
closing as the bug was reverted, and this has merge conflicts. We could still add some tests here though

Issue #4876, opened 2023-05-25
[vue table] row values cache doesn't play well with vue proxy-based reactivity


jonatan-zint-tfs:
### Describe the bug

While creating  table component with vue using @tanstack/table I realized that the tanstack table doesn't play well with reactivity using

```ts
useVueTable({
    [...]
    get data() {
        return props.data
    },
})
```

If you change the contents of props.data (like push a new element) it's perfectly fine for the vue reactivity model. Even the data() function does get reexcecuted - so I'm guessing that the memo() function in the utils of tanstack table decides that props.data did not actually change and it won't get propagated to the row model. See the codesandbox attached, which will make the issue clear I think.

### Your minimal, reproducible example

https://codesandbox.io/p/sandbox/busy-browser-61p7of?welcome=true

### Steps to reproduce

1. Use a reactive array in `get data()` when setting up the table with `useVueTable`
2. Push a new element to that array
3. Realize that it gets properly updated if i.e. just rendered on the template
4. Realize that no new row is displayed in the table (table.getRowModel()) returns old data

### Expected behavior

I'd like to harmonize vue reactivity with the tanstack reactivity at this point.

I _can_ workaround this issue by copying the data before using it in data() - but that'd create an unnecessary extra copy step:

```ts
const data = computed(() => {
  return [...props.data];
});

const table = useVueTable<RowData>({
  get data() {
    return data.value;
  },
});
```

### How often does this bug happen?

Every time

### Platform

OS: Windows11 with WSL2 on Ubuntu 22.04
Browser: FF 113

### react-table version

8.7.9

### TypeScript version

5.0.4

### Terms & Code of Conduct

- [X] I agree to follow this project's Code of Conduct
- [X] I understand that if my bug cannot be reliable reproduced in a debuggable environment, it will probably not be fixed and this issue may even be closed.

Tanimodori:
This library won't reprocess `data` unless it changes reference as compared via `Object.is` [(docs)](https://github.com/TanStack/table/blob/main/docs/api/core/table.md#data). Despite it claims to be an "Framework Agnostic" lib, this limitation is originated in `react` `useState` (e.g. `data`/`setData`) which has to change reference in order to trigger a full-update. Similarly, changing existing data object properties in parent won't trigger update. It is a pity that this library cannot handle partial update like how `vue` does.

radusuciu:
[...]

PR #5139, opened 2023-10-23
feat: add a "disableValueCache" option

jonatan-zint-tfs:
When using tanstack table with proxy objects as row data, copying cell values to the cache is unwanted behavior as they are no longer connected to the row proxy object. In these scenarios i'd like to turn off the value cache.

ref #4876

To illustrate the problem further You can read this comment and see the attached sandbox examples: https://github.com/TanStack/table/issues/4876#issuecomment-1719506142

### DIFF
--- a/docs/api/core/table.md
+++ b/docs/api/core/table.md
@@ -166,7 +166,17 @@ Set this option to true to output column debugging information to the console.
 debugRows?: boolean
 ```
 
-Set this option to true to output row debugging information to the console.
+### `disableValueCache`
+
+```tsx
+disableValueCache?: boolean
+```
+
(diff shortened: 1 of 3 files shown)


KevinVandy:
@OlaAlsaker Curious on your opinion on this PR. I was originally saving this type of change for v9

OlaAlsaker:
I don't this change is necessary anymore after we introduced support for reactivity in `v8.20.0`. @jonatan-zint-tfs , could you maybe test out the latest update and see if it solves your problems? 😄

jonatan-zint-tfs:
Sure I was kinda reshifted to another project, but I'll check with my collegues!

PR #5190, opened 2023-12-06
update: filters.md

zainsci:
Its confusing on quick glances.


KevinVandy:
The reason it is worded that way is to clarify what the default values are.

zainsci:
Thanks for the clarification :sweat_smile:

Issue #4331, opened 2022-08-22
RTL column resizing doesn't work


giladv:
### Describe the bug

while on RTL mode, column resizing is working backwards.

### Your minimal, reproducible example

https://codesandbox.io/s/mystifying-glade-ng5e9f

### Steps to reproduce

try to resize a column

### Expected behavior

should resize columns as expected but from right to left

### How often does this bug happen?

Every time

### Platform

all

### react-table version

latest

### Terms & Code of Conduct

- [X] I agree to follow this project's Code of Conduct
- [X] I understand that if my bug cannot be reliable reproduced in a debuggable environment, it will probably not be fixed and this issue may even be closed.

ierfaaan:
I'm getting the same issue :((

mohsenomidi:
Dear @tannerlinsley 
Thanks for your great data table library 
Same issue here also in version 8, the RTL support is very important for many languages, is there any plan to fix resizing direction

samuelarbibe:
Hey guys!   

You can check out my fork:  
https://github.com/samuelarbibe/table  
That fixes column resizing according to an added `rtl` option.  
  
To make column resizing work in an `rtl` context, you need to change 2 things:  
1. add `rtl: true` option to you `useTable` declaration  
```diff
  const table = useReactTable({
    data,
    columns,
+   rtl: true,
    columnResizeMode,
    getCoreRowModel: getCoreRowModel(),
    debugTable: true,
    debugHeaders: true,
    debugColumns: true
  });
```
2. change the `css` for you resizer
```diff
.resizer {
  position: absolute;
  top: 0;
+ left: 0;
- right: 0;
  height: 100%;
  width: 5px;
  background: rgba(0, 0, 0, 0.5);
  cursor: col-resize;
  user-select: none;
  touch-action: none;
}
```

This is a very basic WIP, just to get things to work.  
I can't promise I will form it into a complete `PR`, as this was done for personal purposes.

AlonMiz:
hey, is there any progress on that? 🙏

samuelarbibe:
Hey!
I have resolved to making my own forked library and using it, because I didn’t find a nice enough solution worthy of a PR.

samuelarbibe:
Hey! just submitted a PR to add support for RTL column resizing. check it out: #5192

KevinVandy:
^twas shipped

PR #5192, opened 2023-12-08
Feature: Add columnResizeDirection to support RTL column resizing

samuelarbibe:
This PR adds support for RTL column resizing by adding a `columnResizeDirection` to column resize options, as discussed in #4331

### DIFF
--- a/docs/api/features/column-sizing.md
+++ b/docs/api/features/column-sizing.md
@@ -152,6 +152,14 @@ columnResizeMode?: 'onChange' | 'onEnd'
 
 Determines when the columnSizing state is updated. `onChange` updates the state when the user is dragging the resize handle. `onEnd` updates the state when the user releases the resize handle.
 
+### `columnResizeDirection`
+
+```tsx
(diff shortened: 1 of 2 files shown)


docs/api/features/column-sizing.md
```diff
@@ -152,6 +152,14 @@ columnResizeMode?: 'onChange' | 'onEnd'
 
 Determines when the columnSizing state is updated. `onChange` updates the state when the user is dragging the resize handle. `onEnd` updates the state when the user releases the resize handle.
 
+### `columnResizeDirection`
```
AlonMiz, line 155: isnt there a global setting to rtl, that we can depend on instead of allowing rtl specifically for column resizing?

samuelarbibe: In another PR, I have introduced a global option rtl that could be used in other places.

But, this is the only feature that requires an explicit rtl setting. Other things can be sorted using CSS properties.

Therefore, I thought it would be more appropriate to make this a feature-specific option.

KevinVandy (approved):

PR #5221, opened 2023-12-20
fix: fixed crash when toggling selection of non-visible row

KevinVandy:
Currently, if the mutation of a selected row is tried on a row that is not in the current pagination or filtered results, the table will not be able to find the row to toggle and will crash. This can happen more frequently with the new row pinning features. This fix scans all client-side rows to perform the mutation.


akraines:
This change caused the regression reported here:
https://github.com/TanStack/table/issues/5246

PR #5233, opened 2023-12-27
chore: Migrate from TypeScript 4.9 to 5.2

lachlancollins:
An incremental step to bringing the workspace config more in line with TanStack Query. TS version is pinned so it doesn't auto update (TS doesn't use semantic versioning).

### DIFF
--- a/examples/solid/basic/package.json
+++ b/examples/solid/basic/package.json
@@ -10,7 +10,7 @@
   },
   "license": "MIT",
   "devDependencies": {
-    "typescript": "5.2.2",
+    "typescript": "5.3.3",
     "vite": "^3.2.3",
     "vite-plugin-solid": "^2.2.6"
   },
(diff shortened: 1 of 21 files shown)


KevinVandy:
Does the project still build after this? Last time I tried this a few months ago, there was some roll-up incompatibility

lachlancollins:
> Does the project still build after this? Last time I tried this a few months ago, there was some roll-up incompatibility

I tested `yarn run pr` locally (and it also seemed to work on CI). Do you know what was breaking?

KevinVandy:
@lachlancollins checked out your PR and `yarn build` is still fine. Don't remember what I ran into a few months ago.

I ran `npx npm-check-updates -u -ws --root typescript` to update to version 5.3.3 though and had the same success, so might as well upgrade to the latest

PR #5234, opened 2023-12-27
chore: Migrate to pnpm and modularise Nx tasks

lachlancollins:
Changes:
- `pnpm import` to clone existing locked versions
- Setup pnpm workspaces
- Remove `yarn.lock` and lerna
- Split tsc and vitest into per-package scripts rather than root-level scripts (can be parallelised by Nx)
- Fix broken react dependency resolution
- Replace `rm -rf` (unix-specific) with `rimraf`

CI testing time:
- Before: 42s
- After: 31s


lachlancollins:
@KevinVandy I've made some great progress here! I reckon I'll be able to submit the tsup PR after this.

KevinVandy (approved):
I've tried to find anything broken with this, but don't see anything. Thanks for the hard work. We'll see if publishing still works

packages/vue-table/package.json
```diff
@@ -42,8 +42,12 @@
     "build/umd/*",
     "src"
   ],
+  "scripts": {
+    "clean": "rimraf ./build",
+    "test:types": "tsc"
+  },
   "dependencies": {
-    "@tanstack/table-core": "8.11.2"
+    "@tanstack/table-core": "workspace:*"
```
KevinVandy, line 50: Doesn't this need to tied to a version?

KevinVandy: This gets installed in a user's node_modules

lachlancollins: Pnpm will handle replacing workspace fields with versions during publishing - you can see more about it here: https://pnpm.io/workspaces#publishing-workspace-packages. And whatever files are in node_modules would be symlinks back to the local folder :)

PR #5238, opened 2023-12-28
chore: Modularise rollup config

lachlancollins:
Changes:
- Updated rollup deps
- Turned root rollup config into `getRollupConfig.mjs`
- Added rollup config to each package
- Added separate test:types and build:types commands
- Fixed vue-tsc by downgrading TS to 5.2
- Use faster Nx commands
- Remove tsconfig references/paths


scripts/tsconfig.json
```diff
@@ -1,7 +1,7 @@
 {
   "compilerOptions": {
-    "target": "ES5",
-    "module": "commonjs",
```
KevinVandy, line 4: I'm not sure we can change this without a major version?

lachlancollins: This is for the scripts directory! Nothing here actually gets built or packaged, it just stops the language server from complaining.

KevinVandy (approved):

PR #5249, opened 2024-01-01
style: Update prettier config

lachlancollins:
Uses the prettier settings common to query, forms, router, etc.

```
semi: false,
singleQuote: true,
trailingComma: 'all',
```


KevinVandy:
I'd like to do this, but as a result I think we'd have to officially close all of the other stale PRs. And I'd have to start my v9 branch over, though that's not too hard for me.
Most of the remaining open table PRs already have merge conflicts, so maybe it is an ok time to do this.

lachlancollins:
> I'd like to do this, but as a result I think we'd have to officially close all of the other stale PRs. And I'd have to start my v9 branch over, though that's not too hard for me. Most of the remaining open table PRs already have merge conflicts, so maybe it is an ok time to do this.

It's not too hard to apply this change (just change one file and let the command format the rest), so I'd say do it whenever is going to be least inconvenient!

KevinVandy:
I think by the end of the week, maybe I am able to make a v9 branch, and we put this config there. This v9 stuff is still kinda iffy if it's going to happen or work, but I've got approval from Tanner to explore some ideas.

PR #5251, opened 2024-01-03
docs: add solid filters example

Hephaestian:
Added a working example in Solid.JS for both global and column filters. Format and structure follows the React example reasonably closely - however I did omit a few things (eg. sorting, pagination) as I felt they were separate Table features and I did not want to complicate the code for this example.

### DIFF
--- a/docs/config.json
+++ b/docs/config.json
@@ -327,6 +327,10 @@
           "label": "Column Visibility"
         },
         {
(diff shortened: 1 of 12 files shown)


Hephaestian:
Had to clean up some leftover comments and want to keep the commit history clean... will re-raise a separate PR.

PR #5268, opened 2024-01-10
fix: column.getCanSort() returns true when no sorting option is passed to options

MRTamalampudi:
`column.getCanSort()` returns true when no sorting option is passed to options
#### CASE 1:
when there is no mention of `enableSorting` property explicitly in table options or columunDef, `column.getCanSort()` is returning 
`true`

#### CASE 2:
when `enableSorting : true` is passed to table options , column.getCanSort() is returning `true` for all the columns unless `enableSorting : false`  is passed to columnDef's

#### Fix
if there is no mention of `enableSorting` property it should return `false` by default when `column.getCanSort()` is called.

### DIFF
--- a/packages/table-core/src/features/Sorting.ts
+++ b/packages/table-core/src/features/Sorting.ts
@@ -474,8 +474,8 @@ export const Sorting: TableFeature = {
 
     column.getCanSort = () => {
       return (
-        (column.columnDef.enableSorting ?? true) &&
-        (table.options.enableSorting ?? true) &&
+        (table.options.enableSorting ?? false) &&
+        (column.columnDef.enableSorting ?? false) &&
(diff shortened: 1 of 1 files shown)


KevinVandy (commented):
So the thing is that both of those cases are the intended functionality, and changes this would be a huge breaking change for everyone. The idea for how it works is that all features are enabled by default, and will work by default if you supply the correct row models. Then you can opt out of sorting on either a column or table-wide basis.

MRTamalampudi:
Hey KevinVandy, got your point, but raised this PR only based on docs provided for `column.getCanSort()` and  if we go by method name. If a column is a `Accessor Column` sorting is enabled by default, understood. Thanks @KevinVandy.

PR #5274, opened 2024-01-14
fix(table-core): Do not select subrows if enableSubRowSelection is set to false

### LINKED ISSUE #5116: getToggleAllRowsSelectedHandler doesn't take into account enableSubRowSelection property
### Describe the bug

Even if the table is provided with enableSubRowSelection: false, calling a handler from `getToggleAllRowsSelectedHandler` will still select all rows. 

https://codesandbox.io/p/sandbox/focused-ben-wfks52?embed=1&file=%2Fsrc%2Fmain.tsx%3A148%2C5

### Steps to reproduce

1. Select all columns from the header
2. Expandable columns are selected even if the enableSubRowSelection = false

### Expected behavior

Expandable rows should not be selected if enableSubRowSelection = false

Every time

### Screenshots or Videos

_No response_

### Platform

Chrome, Mac

### react-table version

8.10.3

### TypeScript version

5.1.3

### Additional context

_No response_

### Terms & Code of Conduct

jpedroh:
Fixes #5116.

There are two cases here that were not being covered by the current code:
- If `enableSubRowSelection` is a boolean and is set to true, then only the first level of rows should be selected. We can  easily achieve this by iterating over `rows` instead of `flatRows`.
- The more complicated case is when `enableSubRowSelection` is a function. In this scenario we must run the check recursively for all subrows. The easiest way I've found to solve this was to just iterate over rows using the `mutateRowIsSelected` helper function.

In order to not losing the performance improvement that was already implemented, I've kept the old flow and just in case enableRowSeleciton is a function, we will reach the slower path (which might not be very common anyway)

### DIFF
--- a/packages/table-core/__tests__/RowSelection.test.ts
+++ b/packages/table-core/__tests__/RowSelection.test.ts
@@ -355,6 +355,45 @@ describe('RowSelection', () => {
       expect(allRows).toEqual([true, false, true, false])
     })
 
+    test('it should not select subrows if enableSubRowSelection is a function and it evaluates to false', () => {
+      const data = makeData(2, 2, 1)
+      const columns = generateColumns(data)
+
+      const state: Partial<TableState> = {
+        rowSelection: {},
+      }
+
+      const table = createTable<Person>({
+        onStateChange() {},
+        renderFallbackValue: '',
+        data,
+        columns,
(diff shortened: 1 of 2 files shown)


KevinVandy:
closing in favor of #6495

PR #5331, opened 2024-02-06
chore: update migrating.md

idmx:
Added new info for migrate from v7 to v8


KevinVandy:
sorry, most of this seems inaccurate except for the column visibility stuff

idmx:
Approx If the information that worked for me is not accurate, I hope you will finally add more accurate information in the near future. It's not easy to migrate without it.

PR #5336, opened 2024-02-07
fix(docs): remove unnecessary object destructuring

amirhhashemi:
Unless it's a new trend to destructure an object to pass multiple props to a component, I think we should avoid it :)


KevinVandy:
I somewhat agree with you, but this is actually what Tanner likes doing and recommends. There was a whole tweet thread about it a couple years ago.

amirhhashemi:
> I somewhat agree with you, but this is actually what Tanner likes doing and recommends. There was a whole tweet thread about it a couple years ago.

Ok. I didn't know about that. So I'm gonna close it.

PR #5341, opened 2024-02-12
docs: updated deprecated prop

GalindoSVQ:
Update a TablePagination property from Material UI

https://mui.com/material-ui/api/table-pagination/#table-pagination-prop-SelectProps


KevinVandy:
Funny, MUI just announced today that they will be getting rid of slotProps by the end of the year, but makes sense for now lol

KevinVandy (approved):

PR #5344, opened 2024-02-13
feat: new column pinning, sizing, ordering APIs for easier sticky pinning

KevinVandy:
**perf**: getMemoOptions method to reduce memo boilerplate
**feat**: new ColumnSizing getAfter API (similar to getStart)
**perf**: added memoization to column.getStart method
**feat**: new Ordering column.getIndex API
**feat**: new Ordering column.getIsFirstColumn API
**feat**: new Ordering column.getIsLastColumn API
**perf**: improved table._getPinnedRows memo performance
**docs**: new sticky column pinning example
**docs**: new column pinning guide

image


KevinVandy:
Also looks like I lowered the bundle size by 0.6kb here

PR #5373, opened 2024-02-26
fix(table-core): Avoid process not defined in environments not setting a process variable

timrittgers:
When running the library in an environment that is not setting the process variable, the library throws a not defined error and breaks the render. This change checks to see if the variable exists before attempting to read it to avoid this.

### DIFF
--- a/packages/table-core/src/core/column.ts
+++ b/packages/table-core/src/core/column.ts
@@ -96,7 +96,7 @@ export function createColumn<TData extends RowData, TValue>(
 
         for (const key of accessorKey.split('.')) {
           result = result?.[key]
-          if (process.env.NODE_ENV !== 'production' && result === undefined) {
+          if (typeof process !== "undefined" && process.env.NODE_ENV !== 'production' && result === undefined) {
             console.warn(
               `"${key}" in deeply nested key "${accessorKey}" returned undefined.`
             )
@@ -112,7 +112,7 @@ export function createColumn<TData extends RowData, TValue>(
   }
 
   if (!id) {
-    if (process.env.NODE_ENV !== 'production') {
+    if (typeof process !== "undefined" && process.env.NODE_ENV !== 'production') {
       throw new Error(
         resolvedColumnDef.accessorFn
           ? `Columns require an id when using an accessorFn`
--- a/packages/table-core/src/core/table.ts
+++ b/packages/table-core/src/core/table.ts
@@ -389,7 +389,7 @@ export function createTable<TData extends RowData>(
       if (!row) {
         row = table.getCoreRowModel().rowsById[id]
         if (!row) {
-          if (process.env.NODE_ENV !== 'production') {
+          if (typeof process !== "undefined" && process.env.NODE_ENV !== 'production') {
             throw new Error(`getRow could not find row with ID: ${id}`)
           }
           throw new Error()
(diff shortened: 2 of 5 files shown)


KevinVandy:
What environments do that?

timrittgers:
> What environments do that?

I have a situation where we are bundling Tanstack into a library component, then distributing the component as a standalone package that can be loaded directly in an html page. In this scenario, there is no process variable defined and the tanstack library fails.

KevinVandy:
Are you sure that you're not accidentally importing from `src` instead of the proper build output?

Also, maybe try importing from the `umd` output, which does not have any process.env in its output

timrittgers:
> Are you sure that you're not accidentally importing from `src` instead of the proper build output?
> 
> Also, maybe try importing from the `umd` output, which does not have any process.env in its output

I'm importing the Tanstack library into my project's component like this:
```
import {
  Cell,
  ColumnDef,
  ColumnSort,
  createColumnHelper,
  getCoreRowModel,
  getFilteredRowModel,
  getPaginationRowModel,
  getSortedRowModel,
  Header,
  PaginationState,
  RowModel,
  Table,
  TableOptions,
  TableState,
} from '@tanstack/table-core';
```

There wouldn't be a way to just import the pieces I need using the umd, would there?

KevinVandy:
> I'm also curious what the harm would be in detecting the process variable in the library before calling it?

I have just never seen any other NPM library ever have to do this, so I think there should be a better way.

lschierer:
I hit this trying to use tanstack table in [Greenwood](https://greenwoodjs.dev/)

thescientist13:
Hey, stumbled across this with @lschierer and did a little digging and have some observations to share if it helps 👋 

> What environments do that?

browser / client environments, no bundle development workflows, etc.

> Also, maybe try importing from the umd output, which does not have any process.env in its output

Tools following NodeJS exports spec looking for ESM files would look to the `import` conditions, assuming "pure" ESM.  In the absence of an `exports` field, they can optionally look for the `module` field. 

In [_package.json_](https://unpkg.com/browse/@tanstack/table-core@8.20.5/package.json) both the `module` field and the `import` sub condition in the exports map field both point to built files that still include `process`.
- `module` - https://unpkg.com/browse/@tanstack/table-core@8.20.5/build/lib/index.esm.js
- `import` - https://unpkg.com/browse/@tanstack/table-core@8.20.5/build/lib/index.mjs
  
(Curious, is there any reason `module` and `import` would point to different file outputs?  I think they could / should probably be the same for ESM based content)

> I have just never seen any other NPM library ever have to do this, so I think there should be a better way.

It's worth noting that the reason most NPM libraries don't guard against this is because most people "erase" it away through their build tools.

For instance, you can see that happening in this [project](https://github.com/search?q=repo%3ATanStack%2Ftable%20process&type=code) in its examples and own build step
- https://github.com/TanStack/table/blob/827b09814c659a5d196e63c7b827858db243a9cd/examples/lit/basic/vite.config.js#L11
- https://github.com/TanStack/table/blob/827b09814c659a5d196e63c7b827858db243a9cd/scripts/getRollupConfig.js#L18
- https://github.com/TanStack/table/blob/827b09814c659a5d196e63c7b827858db243a9cd/examples/react/query-router-search-params/vite.config.js#L13
- https://github.com/TanStack/table/blob/827b09814c659a5d196e63c7b827858db243a9cd/examples/react/kitchen-sink/vite.config.js#L12
- etc

----

An alternative to guarding would be that for client side / browser / UI libraries don't include any non web standard (e.g. NodeJS APIs)


Not the biggest deal, but hopefully it's helpful information either way.  👍

bburns:
I'm trying a no-bundle workflow, using jsenv to generate an importmap pointing to packages in node_modules, and ran into this error. 

For now, a fix is to put this in your code before importing tanstack -

    window.process = { env: {} }

KevinVandy:
we're considering adding something to address this in v9, but you can also just define window.process to get around this.

PR #5377, opened 2024-02-27
feat: pagination row count apis, new pagination docs

KevinVandy:
feat: Added `rowCount` table option as an alternative to `pageCount`
feat: Added `getRowCount` table instance API
feat: Added `firstPage` table instance API
feat: Added `lastPage` table instance API
feat:  Deprecated `setPageCount` API
docs: Updated pagination API Docs
docs: Wrote new Pagination Guide
docs: Updated the Pagination and Pagination Controlled examples


usmanibrahim74:
is there a way to update page count after table instance? any help?

PR #5395, opened 2024-03-06
docs: rephrase options to reduce confusion

newtriks:


KevinVandy:
it's phrased this way for a reason. If undefined, these values will default to true.

newtriks:
Okay, however, the problem remains in regards to readability, it maybe worth adding your further clarification to the docs.

Why not remove the unnecessary need to handle undefined as you do using nullish coalescing and simply treat it as an actual boolean by initialising it with true instead.

Issue #5398, opened 2024-03-07
Toggle select all rows doesn't respect `canSelect()` condition 


simcha90:
### TanStack Table version

8.13.2

### Framework/Library version

17.0.2

### Describe the bug and the steps to reproduce it

When you click unselect all rows it unselect even disabled rows (that was set for `enableRowSelection`).
It looks  that happens in `mutateRowIsSelected` when value is `false` it always remove selection:
```
if (value) {
    if (!row.getCanMultiSelect()) {
      Object.keys(selectedRowIds).forEach(key => delete selectedRowIds[key])
    }
    if (row.getCanSelect()) {
      selectedRowIds[id] = true
    }
  } else {
    delete selectedRowIds[id]
  }
```

### Your Minimal, Reproducible Example - (Sandbox Highly Recommended)

https://codesandbox.io/p/devbox/loving-voice-7sqqp7?file=%2Fsrc%2Fmain.tsx%3A109%2C13

### Do you intend to try to help solve this bug with your own PR?

None

### Terms & Code of Conduct

- [X] I agree to follow this project's Code of Conduct
- [X] I understand that if my bug cannot be reliable reproduced in a debuggable environment, it will probably not be fixed and this issue may even be closed.

vishwas0459:
I am also facing the same issue. 
@simcha90  - You may want to update the `rowSelection` state type to remove the eslint error showing on your codesandbox link. 
`const [rowSelection, setRowSelection] = React.useState<RowSelectionState>({0: true})`

mylestlee:
The first issue was fixed [here](https://github.com/TanStack/table/pull/5061/commits/b76bb8d7a0e85fd33ebff7c4fe2d1729c7ceca1d) however it seems to have been reintroduced in one of the [next commits](https://github.com/TanStack/table/pull/5061/commits/42f7595d4411eeac97d9e96f6850d9002c256dcf). I'm also facing this issue.

For some reason `table.getFilteredRowModel().flatRows.filter(row => row.getCanSelect()).length` was removed which fixed the first issue.

PR #5401, opened 2024-03-10
fix(selection): update logic to fix selection of disabled rows

simcha90:
Fix issue:

https://github.com/TanStack/table/issues/5398

We updated logic for selection of disabled rows (at least for pagination).

Here next possible flows:

1. One row disabled and selected then click on Footer checkbox:
- if all other rows selected - Footer checkbox: SELECTED
- if all other rows unselected - Footer checkbox: INDETERMINATE

2. One row disabled and unselected then click on Footer checkbox:
- if all other rows selected - Footer checkbox: INDETERMINATE
- if all other rows unselected - Footer checkbox: UNSELECTED

3. One row disabled and selected and One row disabled and unselected then click on Footer checkbox:
- if all other rows selected - Footer checkbox: INDETERMINATE
- if all other rows unselected - Footer checkbox: INDETERMINATE

Status of all enabled checkboxes always will be toggled, but sign inside Footer checkbox will show real state of all checkboxes

P.S. Also `getToggleAllRowsSelectedHandler` respect disabled checkboxes

### DIFF
--- a/packages/table-core/src/features/RowSelection.ts
+++ b/packages/table-core/src/features/RowSelection.ts
@@ -447,7 +447,7 @@ export const RowSelection: TableFeature = {
               d => d.getIsSelected() || d.getIsSomeSelected()
             ) &&
               paginationFlatRows.some(
-                d => !d.getIsSelected() && !d.getIsSomeSelected()
+                d => !d.getIsSelected() || !d.getIsSomeSelected()
               )
           : paginationFlatRows
               .filter(row => row.getCanSelect())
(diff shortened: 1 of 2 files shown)


packages/table-core/src/features/RowSelection.ts
```diff
@@ -433,13 +441,25 @@ export const RowSelection: TableFeature = {
...
+          ? paginationFlatRows.some(
+              d => d.getIsSelected() || d.getIsSomeSelected()
+            ) &&
+              paginationFlatRows.some(
+                d => !d.getIsSelected() || !d.getIsSomeSelected()
+              )
+          : paginationFlatRows
+              .filter(row => row.getCanSelect())
+              .some(d => d.getIsSelected() || d.getIsSomeSelected())
+      }
```
2jordan3, line 458: I think the readability is still lacking. 
How about separating this part into an independent function or clearly splitting it with if-else statements?

KevinVandy:
closing in favor of #6495

Issue #5213, opened 2023-12-15
Svelte 5 support: ERROR: No matching export in "a" for import "b"


niemyjski:
### Describe the bug

Svelte 5 preview ships with the new create app cli as an option. When using the previous the app cannot compile due to the errors below:

```
  VITE v5.0.9  ready in 866 ms

  ➜  Local:   http://localhost:5173/next
  ➜  Network: use --host to expose
  ➜  press h + enter to show help
X [ERROR] No matching export in "node_modules/svelte/src/internal/index.js" for import "SvelteComponent"

    node_modules/@tanstack/svelte-table/build/lib/index.mjs:13:9:
      13 │ import { SvelteComponent, init, safe_not_equal, text, claim_text, insert_hydration, set_data, noop, detach, create_ssr_component, escape, validate_component, create_component, claim_com...
         ╵          ~~~~~~~~~~~~~~~

X [ERROR] No matching export in "node_modules/svelte/src/internal/index.js" for import "init"

    node_modules/@tanstack/svelte-table/build/lib/index.mjs:13:26:
      13 │ import { SvelteComponent, init, safe_not_equal, text, claim_text, insert_hydration, set_data, noop, detach, create_ssr_component, escape, validate_component, create_component, claim_com...
         ╵                           ~~~~

X [ERROR] No matching export in "node_modules/svelte/src/internal/index.js" for import "safe_not_equal"

    node_modules/@tanstack/svelte-table/build/lib/index.mjs:13:32:
      13 │ import { SvelteComponent, init, safe_not_equal, text, claim_text, insert_hydration, set_data, noop, detach, create_ssr_component, escape, validate_component, create_component, claim_com...
         ╵                                 ~~~~~~~~~~~~~~

X [ERROR] No matching export in "node_modules/svelte/src/internal/index.js" for import "claim_text"

    node_modules/@tanstack/svelte-table/build/lib/index.mjs:13:54:
      13 │ import { SvelteComponent, init, safe_not_equal, text, claim_text, insert_hydration, set_data, noop, detach, create_ssr_component, escape, validate_component, create_component, claim_com...
         ╵                                                       ~~~~~~~~~~

X [ERROR] No matching export in "node_modules/svelte/src/internal/index.js" for import "insert_hydration"

    node_modules/@tanstack/svelte-table/build/lib/index.mjs:13:66:
[...]

PR #5403, opened 2024-03-11
feat: svelte 5 adapter

wlockiv:
Overhauls the Svelte adapter so that it's compatible with Svelte 5.

Some overview of the work that was done:

- All dependencies on `svelte/internal` are removed. 
- `createSvelteTable` now returns a signal in lieu of a store. According to @dummdidumm, Stores may eventually be deprecated.
- The `flexRender` API is now replaced with a more idiomatic `FlexRender` component.
- `FlexRender` employs the use of Svelte 5's new Snippet feature - it made this adapter way simpler.
- Typescript typing is more robust than it was before
- All user-facing code that's not being carried over from the core library is documented.

Feedback welcome.

solves #5213


niemyjski (commented):
Looks good

wlockiv:
The `createSvelteTable` function needs a bit more work. It's not very efficient at the moment. I've marked as draft while I work on this.

packages/svelte-table/src/flex-render.svelte
```diff
@@ -0,0 +1,91 @@
...
+  import type { HeaderOrCellContext } from './types'
+
+  type Props = {
+    /** The cell or header field of the current cell's column definition. */
+    content: ColumnDefTemplate<TContext> | undefined
+    /** The result of the `getContext()` function of the header or cell */
+    context: TContext
+  }
+
+  let { content, context } = $props<Props>()
```
dummdidumm, line 75: This changes in the preview, now needs to be like this
```suggestion
  let { content, context }: Props = $props()
```

wlockiv: thank you 🙇🏻‍♂️

packages/svelte-table/package.json
```diff
@@ -20,37 +20,40 @@
...
+  "main": "build/index.js",
+  "types": "build/index.d.ts",
   "exports": {
     ".": {
-      "types": "./build/lib/index.d.ts",
-      "import": "./build/lib/index.mjs",
-      "default": "./build/lib/index.js"
+      "types": "./build/index.d.ts",
+      "import": "./build/index.js",
+      "svelte": "./build/index.js"
```
dummdidumm, line 31: Svelte should probably come first
```suggestion
      "svelte": "./build/index.js",
      "import": "./build/index.js"
```

packages/svelte-table/src/flex-render.svelte
```diff
@@ -0,0 +1,109 @@
...
+    {@const result = content(context)}
+    {#if result instanceof RenderComponentConfig}
+      <svelte:component this={result.component} {...result.props} />
+    {:else}
+      {result}
+    {/if}
+  {/if}
+{/snippet}
+
+{@render componentCell()}
```
dummdidumm, line 109: Why is this declaring and then rendering a snippet? Looks like the logic inside the snippet could just be inlined

packages/svelte-table/package.json
```diff
@@ -20,37 +20,40 @@
...
     "src"
   ],
   "scripts": {
     "clean": "rimraf ./build",
-    "test:types": "tsc --noEmit",
-    "build": "pnpm build:rollup && pnpm build:types",
-    "build:rollup": "rollup --config rollup.config.mjs",
-    "build:types": "tsc --emitDeclarationOnly"
+    "test:types": "svelte-check --tsconfig ./tsconfig.json",
+    "build": "svelte-package --input ./src --output ./build"
```
dummdidumm, line 45: since this is no longer using rollup - is the rollup config/build used anywhere else? If not, the related code can be removed.

dummdidumm (commented):
🎉

packages/svelte-table/src/flex-render.svelte
wlockiv: i thought wrapping this in a snippet would handle the server/client gymnastics that the svelte 4 version of this package needed to do. That's not the case for svelte 5?

packages/svelte-table/package.json
wlockiv: good call - removed the usages i found.

packages/svelte-table/src/flex-render.svelte
dummdidumm: The old version needed this for other reasons - to be honest I'm pretty sure that the old version could've used a similar approach to this one without resorting to hacking internals.

KevinVandy:
This is looking good.

KevinVandy:
@wlockiv @niemyjski @dummdidumm I've been consulting with some of the other TanStack maintainers, and we think it should actually be possible to keep backwards functionality with Svelte 3/4. If it's possible to have this adapter work with Svelte 3-5, we need to try that.

This is pretty tricky, and I'll do my best to catch up on the Svelte snippet stuff. Was there a different way that the original svelte adapter could have been using flexRender without Svelte internal?

dummdidumm:
Are you saying that there should be no breaking change at all, or that there can be breaking changes, but it should also work with Svelte 3/4? It should be possible to have it work 3-5 with a breaking change (the `FlexRender` component) by keeping the store API (`$table.get...` instead of `table.get...`), but it will not be possible to achieve that keeping the same API. That's because the current API passes the everything to `<svelte:component this={..} />`, which means it needs to augment a component on the fly, which is not possible to keep compatible across 3/4 and 5.

KevinVandy:
> Are you saying that there should be no breaking change at all, or that there can be breaking changes, but it should also work with Svelte 3/4?

There can be breaking changes with flexRender. From what I see, that seems unavoidable.

But we should try our best to have the adapter work with Svelte 3-5. From our understanding, Svelte 5 is supposed to mostly be backwards compatible. I think the expectation is that TanStack Svelte Table should work in each Svelte version.

dummdidumm:
In this case it should be possible to support 3-5 in one, if the store style API is kept. What is the idea behind supporting 3-5 if there's a breaking change anyway? The thought that adoption for 5 is gonna take a while so it's better to have one small breaking change and then another one in like half a year when switching towards an API that leverages Svelte 5 capabilities?

KevinVandy:
@dummdidumm How much more efficient is the Svelte 5 way of doing things in terms of performance? Is it large?

sledgehammer999:
Just pointing out that Svelte 5 isn't released yet, so this PR shouldn't be merged before that.

On another note, would it be reasonable to maintain 2 adapters instead? One for Svelte 3/4 and one for Svelte 5.

niemyjski:
I really don’t see why we can’t have multiple releases or packages. yes, we should push a package asap as I’m and others are blocked from even upgrading to svelte 5 unless we drop tan stack… we can’t give feedback easily.Sent from my iPhoneOn Mar 21, 2024, at 12:14 PM, sledgehammer999 ***@***.***> wrote:﻿
Just pointing out that Svelte 5 isn't released yet, so this PR shouldn't be merged before that.
On another note, would it be reasonable to maintain 2 adapters instead? One for Svelte 3/4 and one for Svelte 5.

KevinVandy:
> Just pointing out that Svelte 5 isn't released yet, so this PR shouldn't be merged before that.

Yeah, we will not just remove Svelte 3/4 support. Especially while Svelte 5 is a beta.

> On another note, would it be reasonable to maintain 2 adapters instead? One for Svelte 3/4 and one for Svelte 5.

I was originally against this because of how messy it seemed. But it might be the best way forward.

I can see table having "svelte-runes", or "vue-vapor", etc. adapters.

An issue that we would run into, however, is how to install multiple versions of Svelte in the TanStack Table replo. Probably possible with pnpm, just need to look into it.

@lachlancollins was telling me if should absolutely be possible to support Svelte 3-5 in one adapter though. I just don't know how to get that to work with flexRender, unless we just deprecate that in the current adapter.

niemyjski:
I'd just keep it simple, create a new package with preview suffix targeting 5.0, can move those bits to the old package once released and bump major. If users need to stay on the old package, don't upgrade to a new major.

wlockiv:
> @dummdidumm How much more efficient is the Svelte 5 way of doing things in terms of performance? Is it large?

Imo, it's going to be pretty negligible. At least for state management. The state management layer between Svelte (3/4 or this 5 adapter) is pretty thin. There may even be some ways that observables (from Sv 3/5) integrate w/ @tanstack/table a little more intuitively than signals.

> Just pointing out that Svelte 5 isn't released yet, so this PR shouldn't be merged before that.
> 
> On another note, would it be reasonable to maintain 2 adapters instead? One for Svelte 3/4 and one for Svelte 5.

I know this is more of a question for the _maintainers_, but I am afraid I won't have the margins to help out with an adapter that's available across the major versions (3-5) anytime soon. Assuming someone else doesn't pick that up, 2 adapters would probably mean Svelte 5 compatibility comes sooner for folks who are testing it out.

I'm also not opposed to publishing it myself as an "unofficial community adapter" of sorts. 

The new, easier programmatic rendering made available in Svelte 5 makes the adapter less of an "adapter" and more of an "implementation" if that makes sense. 

> yes, we should push a package asap as I’m and others are blocked from even upgrading to svelte 5 unless we drop tan stack…

I totally understand your frustration, and it's why I started working on this PR. I'm excited to see this available to other svelters. But if this is a blocker for your project, then it's possible to use this adapter before this PR is merged. 

This new adapter is actually really really tiny, straightforward, and, at least for the time being, you could easily drop its code into your existing project. I basically built the adapter in a SvelteKit project, then copy-pasted it into this codebase. I don't see any reason you wouldn't be able to do the same, assuming there aren't any breaking releases from Svelte 5 since the version I developed on.

KevinVandy:
@niemyjski 

> I really don’t see why we can’t have multiple releases or packages. yes, we should push a package asap as I’m and others are blocked from even upgrading to svelte 5 unless we drop tan stack…

TanStack Table should NOT be a blocker for your team at all. You can just install Table core and make your own Svelte adapter. Copy it from this PR even. This is always an option.

dummdidumm:
I don't think two adapters will help much. Supporting 3-5 should be possible if it's ok to have a breaking change to use the FlexRender component introduced in this PR. If so, then the store API can be reintroduced. Once Svelte 5 is stable another major could be released to switch over from store API to runes API (basically this PR). That way the newest version of the adapter works for Svelte 5 and takes advantage of its new features, while the previous version will work for everyone using Svelte 3/4, too.

wlockiv:
@dummdidumm i noticed that release 85 of svelte 5 has a breaking update to `effect.pre`, which we use here. Does that update affect us?

dummdidumm:
No

niemyjski:
Looks like there are conflicts in this pr.

examples/svelte/column-pinning/src/App.svelte
```diff
@@ -279,17 +236,15 @@
...
             {#each headerGroup.headers as header}
               <th colSpan={header.colSpan}>
                 <div class="whitespace-nowrap">
                   {#if !header.isPlaceholder}
-                    <svelte:component
-                      this={flexRender(
-                        header.column.columnDef.header,
-                        header.getContext()
-                      )}
+                    <FlexRender
```
niemyjski, line 245: When I do this in my project with this adapter source imported I get

Type 'ColumnDefTemplate<HeaderContext<TData, unknown>> | undefined' is not assignable to type 'undefined'.
  Type 'string' is not assignable to type 'undefined'.ts(2322)

wlockiv: If you could provide a minimal reproduction, that would be very helpful.

niemyjski:
I can't provide any feedback as pr was closed but was also seeing: 

not sure if any is banned in this project but I'm getting errors source importing and trying out.

>  Argument of type 'HeaderContext<TData, TValue> | CellContext<TData, TValue>' is not assignable to parameter of type 'HeaderContext<TData, TValue> & CellContext<TData, TValue>'.
  Type 'HeaderContext<TData, TValue>' is not assignable to type 'HeaderContext<TData, TValue> & CellContext<TData, TValue>'.
    Type 'HeaderContext<TData, TValue>' is missing the following properties from type 'CellContext<TData, TValue>': cell, getValue, renderValue, row

Feels like the typing here should be inferred from the context, I'm trying to simplify. I'm in the process of trying it out by converting: https://github.com/exceptionless/Exceptionless/blob/main/src/Exceptionless.Web/ClientApp/src/lib/components/data-table/data-table-body.svelte#L44

wlockiv:
> I can't provide any feedback as pr was closed but was also seeing:
> 
> not sure if any is banned in this project but I'm getting errors source importing and trying out.
> 
> > Argument of type 'HeaderContext<TData, TValue> | CellContext<TData, TValue>' is not assignable to parameter of type 'HeaderContext<TData, TValue> & CellContext<TData, TValue>'.
> > Type 'HeaderContext<TData, TValue>' is not assignable to type 'HeaderContext<TData, TValue> & CellContext<TData, TValue>'.
> > Type 'HeaderContext<TData, TValue>' is missing the following properties from type 'CellContext<TData, TValue>': cell, getValue, renderValue, row
> 
> Feels like the typing here should be inferred from the context, I'm trying to simplify. I'm in the process of trying it out by converting: https://github.com/exceptionless/Exceptionless/blob/main/src/Exceptionless.Web/ClientApp/src/lib/components/data-table/data-table-body.svelte#L44

See [this diff](https://github.com/TanStack/table/pull/5403/files#diff-196732b88dc7b2d5f8f4f81fa24346653ad79174e834605db3ca5061e5fe317cR82-R91).

This narrows the type of content based on context. Not saying it's perfect, but without a minimal repro of the issue it's difficult for me to understand and address your specific case.

wlockiv:
Hey @KevinVandy - may we re-open the discussion on this in anticipation of v5's official release? We discussed some options on releasing the adapter earlier: 

- Releasing a new, separate Svelte5 adapter package so Svelte4 compatibility is still maintained via the original (existing) adapter package, or
- Replacing the existing Svelte4 adapter with this one alongside the next major version (v9) of `table`, or
- Not releasing this adapter, but instead adding documentation to inform users who to use TanStack Table in Svelte 5, or
- Releasing this adapter separately as a "community adapter" until introducing this breaking update to `svelte-table` is less detrimental.

niemyjski:
My vote is just to replace it, use the current version if you need 4.x

KevinVandy:
We can go ahead and merge this into the alpha branch soon, but the documentation for svelte state and adapter needs to be updated with this.

KevinVandy (approved):

wlockiv:
I'll try and work on it this weekend.

stefa168:
Hi, I know that v9 is still an alpha, however I am already using it with Svelte 5. Would it be possible to push a new release tag to have a more up to date version of the alpha available for npm?

KevinVandy:
The next large alpha release will be in a couple weeks probably

stefa168:
Gotcha, thanks for the reply.I was asking because the examples (for svelte) in the latest alpha branch commit are not compatible with alpha 9, so I don't know exactly what I should use.

ddxv:
Hello, just looking to try out TanStack with Svelte 5. I went to the TanStack installation page which say this:

> The @tanstack/svelte-table package works with Svelte 3 and Svelte 4.
> 
> NOTE: There is not a built-in Svelte 5 adapter yet, but you can still use TanStack Table with Svelte 5 by installing the
> @tanstack/table-core package and using a custom adapter from the community. See this [PR](https://github.com/TanStack/table/pull/5403) for inspiration.
https://tanstack.com/table/latest/docs/installation

This points to this PR. Now it appears the PR is merged into alpha? So I guess possibly the documentation might need a slight change, now that the PR is merged, how would the Basic Example work for importing this?  

```
script lang="ts">
	import { writable } from 'svelte/store';
	import { flexRender, getCoreRowModel, createSvelteTable} from '@tanstack/svelte-table';
	import type { ColumnDef, TableOptions } from '@tanstack/svelte-table';
	...
```

I saw that `flexRender` -> `FlexRender` but `createSvelteTable` wasn't working.  Additionally, I tried importing from `@tanstack/table-core` but this had issues do to type script? 

Thanks!

walker-tx:
@ddxv Things are rapidly evolving with the `table-core` package such that it's difficult to know exactly what will work at any moment. Alpha is not ready for production, and since so many things are in flux, I have to recommend against migrating to it for non-experimental work. 

Seriously - _please_ don't use alpha in production. Even if it works for you today, at some point it won't. We can't possibly support something so dynamic at the moment.

The documentation still holds true though. You **_should_** be able to copy the implementation shown here to make your own "adapter" using `table-core` v8 (the current production version). This PR was made for v8, not v9 (alpha). Any subsequent changes to it are for v9.

ddxv:
@walker-tx thanks for being straight forward.

I'll be looking forward for when Svelte 5 is out and officially supported. Thanks!

walker-tx:
After some of the questions on Discord and here, I've put together a reference project so folks can see how they can implement their own Svelte 5 adapter for `@tanstack/table-core` v8:

https://github.com/walker-tx/svelte5-tanstack-table-examples

Hopefully, this gives folks a clearer path to value.

niemyjski:
Can we get a new alpha

niemyjski:
@tannerlinsley would it be possible to get a timeline and or more love for svelte 5 (new alpha or something), we have the svelte 5 core team now contributing to tanstack query. This has been a huge pain point for the svelte community.

roostad:
> @tannerlinsley would it be possible to get a timeline and or more love for svelte 5 (new alpha or something), we have the svelte 5 core team now contributing to tanstack query. This has been a huge pain point for the svelte community.

This. 100%.

ddxv:
Ah, I had struggled with the adapter previously, but was able to get it working by using the components from huntabyte/shadcn-svelte (even though the project I was working on did not use Shadcn).

Not sure why it worked well for me this time when I struggled in the past but wrote up some notes:
https://jamesoclaire.com/2025/04/09/easiest-way-to-get-tanstack-table-v8-working-with-svelte-5/

dummdidumm:
I created a drop in replacement in the meantime https://github.com/dummdidumm/tanstack-table-8-svelte-5

nlbdotdev:
@tannerlinsley Any chance we can get official support moved along? What needs to be done?

niemyjski:
I'd love this as well (and query to get moved along too)

stolinski:
Thanks @dummdidumm hit this when following the docs for the first ime.

PR #5414, opened 2024-03-16
feat: add autoResetSorting

tychenjiajun:
## Summary by CodeRabbit

* **New Features**
  * Added an autoResetSorting option at both table and column levels to automatically reset sorting when the sorting state changes, improving consistency with other auto-reset behaviors.

* **Documentation**
  * Updated sorting feature docs to include the new autoResetSorting option for tables and columns.
  * Clarified surrounding text and adjusted code snippets and formatting for readability.


KevinVandy:
What's the use-case for this? Usually changing the sorting state without user-input is bad UX, right?

tychenjiajun:
> What's the use-case for this? Usually changing the sorting state without user-input is bad UX, right?

Personally I agree that changing the sorting state without user-input is bad UX. But there's still some designers prefer automatically resetting the sorting state when data changes.

tychenjiajun:
Also, see https://github.com/TanStack/table/discussions/5217

KevinVandy:
ok, open to merging this, but need to test more.

PR #5420, opened 2024-03-18
Feat: qwik-table adapter

anxhirr:
Hi TanStack team,

I've created a qwik-table adapter designed to integrate the table core with the Qwik framework. I have tested(not fully) and created examples for column sorting, pagination, and row selection.

I thought it was a good idea to share this as a first version and get some thoughts on whether this looks promising or not.

I welcome any feedback and look forward to contributing/helping make the adapter as perfect as the core itself is xD.

This discussion is related: https://github.com/TanStack/table/discussions/5281


KevinVandy:
Looks good, will probably test this out tomorrow!

packages/qwik-table/package.json
```diff
@@ -0,0 +1,60 @@
...
+    "build": "pnpm build:rollup && pnpm build:types",
+    "build:rollup": "rollup --config rollup.config.mjs",
+    "build:types": "tsc --emitDeclarationOnly"
+  },
+  "dependencies": {
+    "@tanstack/table-core": "workspace:*"
+  },
+  "peerDependencies": {
+    "react": ">=16.8",
+    "react-dom": ">=16.8"
```
KevinVandy, line 58: Why is React a peerDep?

anxhirr: Hi, it shouldn't be there. It's just due to copy-paste from the existing adapter. I will fix it on this new commit together with the filter example I just recently added.

KevinVandy: @anxhirr I already fixed

anxhirr: > @anxhirr I already fixed

haha I saw. I had conflict xD. Solved

anxhirr:
> @anxhirr I worked a bit on this PR. I think it might be in a better place now. All tests passing. Build output looks fine.
> 
> What's the dummy state in the adapter for though?

On the first version, I was getting some serialization error if I did not provide it. 
Now, That error seems to be gone so we can simply remove it

KevinVandy:
@anxhirr I'm fixing the filters example. Also, you should be able to use `pnpm test` and expect tests to pass after my next commit.

anxhirr:
> @anxhirr I'm fixing the filters example. Also, you should be able to use `pnpm test` and expect tests to pass after my next commit.

Oh, well thanks. Yeah, my brain is not braining, I completely forgot we have tests.

KevinVandy:
@anxhirr Take a look at my adapter changes. I made it mostly mirror the react-table adapter. I don't know much about Qwik stores vs signals vs whatever, or why `NoSerialize` is needed here, but it seems to be working better, especially with the initial State.

React on Left. Qwik on Right:

image

anxhirr:
@KevinVandy I made some other small changes, feel free to look.

I switched ```store``` to ```signal``` since Qwik store is supposed to be used when storing multiple states/signals inside the same object, when in this case we don't.

Also, ```noSerialize```  is used because since table core is a third part lib qwik can not serialize($) its functions.

KevinVandy:
@anxhirr This is looking solid, (I mean qwik), pardon the pun. If we're confident that these are best practices, we just need to do some documentation setup. There's some small amount of work to do in the "docs" folder in this repo that I'll do quick.

I also have an open PR for TanStack.com that I'll get one of the other maintainers to merge for me soon. https://github.com/TanStack/tanstack.com/pull/196

KevinVandy:
@anxhirr This can be a future PR after this one, but would you be interested in forking the "React Table State" docs for what is needed for Qwik? https://tanstack.com/table/latest/docs/framework/react/guide/table-state

KevinVandy:
@anxhirr What is this error in each of the examples?

main.tsx:202  QWIK ERROR One of the children of <th> is not an accepted value. JSX children must be either: string, boolean, number, <element>, Array, undefined/null, or a Promise/Signal. Instead, it's a function named "header".
 Error: One of the children of <th> is not an accepted value. JSX children must be either: string, boolean, number, <element>, Array, undefined/null, or a Promise/Signal. Instead, it's a function named "header".

anxhirr:
> @anxhirr What is this error in each of the examples?
> 
> main.tsx:202 QWIK ERROR One of the children of is not an accepted value. JSX children must be either: string, boolean, number, , Array, undefined/null, or a Promise/Signal. Instead, it's a function named "header". Error: One of the children of is not an accepted value. JSX children must be either: string, boolean, number, , Array, undefined/null, or a Promise/Signal. Instead, it's a function named "header".

Just checked, and fixed, very strange tho why it did not appear on my side before I did the commit

anxhirr:
> @anxhirr This can be a future PR after this one, but would you be interested in forking the "React Table State" docs for what is needed for Qwik? https://tanstack.com/table/latest/docs/framework/react/guide/table-state
> 
> Probably as simple as replacing `React.useState` with `Qwik.useSignal` in a few places, I imagine.

@KevinVandy For sure I can help with docs. This weekend I might have plenty of free time

KevinVandy:
@anxhirr I'm actually not sure if state should be in a `useSignal` or `useStore` still. The state is a bunch of nested objects.

I'm guessing the only difference is that `useStore` would re-render even if you do something like `columnPinning.left.push(column.id)`, while the current `useSignal` will not. I guess `useSignal` is safer (from a react dev's perspective).

This can, of course, change later if needed.

anxhirr:
> @anxhirr I'm actually not sure if state should be in a `useSignal` or `useStore` still. The state is a bunch of nested objects.
> 
> I'm guessing the only difference is that `useStore` would re-render even if you do something like `columnPinning.left.push(column.id)`, while the current `useSignal` will not. I guess `useSignal` is safer (from a react dev's perspective).
> 
> This can, of course, change later if needed.

Hi, Yeah I'm not a Qwik expert either but changing it in the future would not be a breaking change I think.

I see this was merged. Do I need to change/add any docs or you handled it all by yourself?

KevinVandy:
> I see this was merged. Do I need to change/add any docs or you handled it all by yourself?

Any holes in the docs, especially more examples, are welcome to PRs always.

thejanasatan:
Hello! trying this out in a hobby project. Looks like I'm running into a crash in dev mode. Looks like it's related to the use of useSignal / useStore inside this hook. I'm new to Qwik myself so not sure if that's accurate - just going by the error message.

Looks like Context might be an alternative that could work?
https://qwik.dev/docs/components/context/

```
10:31:13 am [vite] Internal server error: Code(20) https://github.com/BuilderIO/qwik/blob/main/packages/qwik/src/core/error/error.ts#L28
  File: /work/src/routes/(app)/orders/current/index.tsx:17:13
  15 |  const Table = component$(() => {
  16 |    
  17 |    const t = useQwikTable({
     |               ^
  18 |      data: [],
  19 |      columns: [],
```

KevinVandy:
Create an issue with a sandbox reproducing this error?

thejanasatan:
https://github.com/TanStack/table/issues/5455
done :+1: Thanks again.

packages/qwik-table/src/index.tsx
```diff
@@ -0,0 +1,72 @@
...
+    onStateChange: () => {},
+    renderFallbackValue: null,
+    ...options,
+  }
+
+  // Create a new table instance and store it in a Qwik store
+  const table = Qwik.useStore<{
+    instance: Qwik.NoSerialize<Table<TData>>
+  }>({
+    instance: Qwik.noSerialize(createTable(resolvedOptions)),
```
wmertens, line 48: This should be inside a function. Every time this runs it creates an instance which gets ignored

packages/qwik-table/src/index.tsx
```diff
@@ -0,0 +1,72 @@
...
+  })
+
+  // By default, manage table state here using the table's initial state
+  const state = Qwik.useSignal(table.instance!.initialState)
```
wmertens, line 52: Why copy the state to a signal? It's already a store

packages/qwik-table/src/index.tsx
```diff
@@ -0,0 +1,72 @@
...
+  table.instance!.setOptions(prev => ({
+    ...prev,
+    ...options,
+    state: {
+      ...state.value,
+      ...options.state,
+    },
+    // Similarly, we'll maintain both our internal state and any user-provided
+    // state.
+    onStateChange: updater => {
```
wmertens, line 65: Every time this hook runs it will change this callback. Better to only set it once, no?

wmertens (commented):
I heard there were issues with this integration and vite but I can't see anything that would cause that. In any case, some comments

wmertens:
@KevinVandy not sure if you saw my comments above

KevinVandy:
@wmertens we'd love help remaking the quick adapter the right way. Especially for v9 in the alpha branch.

KevinVandy:
@wmertens The Qwik adapter has kind of been updated for V9 here: https://github.com/TanStack/table/blob/alpha/packages/qwik-table/src/useTable.ts

However, there's a lot of potential to use TanStack table in much different ways that Qwik might be able to better take advantage of, but I might want to discuss with a Qwik expert first

Issue #5423, opened 2024-03-20
Columns created by columnHelper.accessor doesn't know accessorFn or accessorKey


slhsxcmy:
### TanStack Table version

8.14.0

### Framework/Library version

18

### Describe the bug and the steps to reproduce it

Columns created by columnHelper.accessor has type ColumnDef, which is a union of AccessorColumnDef, DisplayColumnDef, and GroupColumnDef. As a result, they don't know that the accessorFn or accessorKey property exists.

Typing them as narrower types like AccessorFnColumnDef or AccessorKeyColumnDef would allow access to these properties. This is helpful for our current use case where we want to directly call the accessorFn on fake data in unit tests.

[...]

PR #5424, opened 2024-03-20
fix(table-core): narrow types for columns created by ColumnHelper

slhsxcmy:
Fixes https://github.com/TanStack/table/issues/5423

Updates return type of the columns created by
- ColumnHelper.accessor to AccessorFnColumnDef or AccessorKeyColumnDef
- ColumnHelper.display to DisplayColumnDef
- ColumnHelper.group to GroupColumnDef

Screenshot 2024-03-19 at 7 34 38 PM

### DIFF
--- a/packages/table-core/src/columnHelper.ts
+++ b/packages/table-core/src/columnHelper.ts
@@ -1,5 +1,7 @@
 import {
   AccessorFn,
+  AccessorFnColumnDef,
+  AccessorKeyColumnDef,
   ColumnDef,
   DisplayColumnDef,
   GroupColumnDef,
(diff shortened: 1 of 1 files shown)


KevinVandy (approved):

edgarbagagem:
Hello everyone,

While this works with helper.accessor({}).accessorKey, it doesn't seem to work for something like this 

``` typescript
const columns: ColumnDef<T>[] = [
helper.accessor('example' ,{ example})
...
]

columns.map(c => c.accessorKey)

Error:
Property 'accessorKey' does not exist on type 'ColumnDef<TData>'.
Property 'accessorKey' does not exist on type 'ColumnDefBase<TData, unknown> & StringHeaderIdentifier'.
```
My use case here is that I pass columns as props for a wrapper component of the table and I have some options where I need to access the accessorKey of the columns

Sorry for bad english

slhsxcmy:
@edgarbagagem try typing the columns `AccessorKeyColumnDef<T>[]` or leave out the type to let Typescript infer

`AccessorKeyColumnDef` has `accessorKey` but `ColumnDef` doesn't

PR #5430, opened 2024-03-22
fix: id building was not consistent when using deep accessorKey (4754)

### LINKED ISSUE #4754: id building is not consistent when using accessorKey
### Describe the bug

When using the accessorKey to define an id as declared in https://github.com/TanStack/table/blob/main/packages/table-core/src/core/column.ts#L39 the logic is not consistent.
```
let id =
    resolvedColumnDef.id ??
    (accessorKey ? accessorKey.replace('.', '_') : undefined) ??
    (typeof resolvedColumnDef.header === 'string'
      ? resolvedColumnDef.header
      : undefined)
```
That rule is fine when data is only one level deep, but does not work as expected when accessorKey digs deeper.
For example, `accessorKey: 'key1.key2'` will produce `id = 'key1_key2'` while `accessorKey: 'key1.key2.key3'` will produce `id = 'key1_key2.key3'`. 
A simple fix would be to use the `g` flag in a regexp for the replace, i.e. `(accessorKey ? accessorKey.replace(/\./g, '_') : undefined)`

http://not.needed.com

### Steps to reproduce

all is said in the description

### Expected behavior

a consistent id

Every time

### Screenshots or Videos

_No response_

### Platform

all

### react-table version

8.7.9

### TypeScript version

_No response_

### Additional context

_No response_

### Terms & Code of Conduct

Roman86:
fixes #4754


KevinVandy (approved):

xkomiks:
@KevinVandy why are you prefer use `_` for `id`? And why is this included in library standard if the previous version had `.`

KevinVandy:
This PR just changed from using replace to replaceAll

PR #5452, opened 2024-03-29
fix: properly mark type of arrays that are not modified by the librar…

pcorpet:
…y as readonly

Fixes https://github.com/TanStack/table/discussions/3648

This is a new attempt after #4909 was closed.

### DIFF
--- a/examples/react/kitchen-sink/src/components/CustomTable.tsx
+++ b/examples/react/kitchen-sink/src/components/CustomTable.tsx
@@ -14,7 +14,7 @@ type TableGroup = 'center' | 'left' | 'right'
 function getTableHeaderGroups<T extends RowData>(
   table: Table<T>,
   tg?: TableGroup
-): [HeaderGroup<T>[], HeaderGroup<T>[]] {
+): [readonly HeaderGroup<T>[], readonly HeaderGroup<T>[]] {
   if (tg === 'left') {
     return [table.getLeftHeaderGroups(), table.getLeftFooterGroups()]
   }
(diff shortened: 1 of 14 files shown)


KevinVandy:
I don't really understand why people want `readonly` added to the TanStack Table types. Is it necessary? Nice to have? I guess can see the case for nicely integrating `data` that was already marked as `readonly` from some state managers or graphql clients without needing to cast, but I don't understand the advantage for all of these other changes.

pcorpet:
Most of the time when handing over an array you don't expect the function or the component that takes that array to be able to modify them. So you're actually thinking of a readonly array. One could think that it should be the default for all arrays in most code with the mutable arrays being the exception.

Here are two use cases that benefit of readonly array in addition to state managers and graphql clients:

1. Using `as const`to strong type quickly. In our code we regularly write static const values with the [`as const`](https://www.typescriptlang.org/docs/handbook/release-notes/typescript-3-4.html#const-assertions), this helps us restrict the type of the constant to its bare minimum. However this creates readonly arrays.

2. Stabilizing props: in order to improve performance, we recommend our developers to treat data as immutable. See [this article](https://medium.com/@esamatti/react-js-pure-render-performance-anti-pattern-fb88c101332f).

If my proposal is a no-go for you, I can try to restrict to the bare minimum and restrict only the interface of TanStack's library.

KevinVandy:
The builds are not passing after these changes

rstrand:
@KevinVandy would this PR (or another PR with the same changes) be considered if the build was fixed?

KevinVandy:
Or, let's make the changes in the v9 alpha

deammer:
Is this feature request tracked anywhere? `readonly` support would be great for compatibility with other libraries. https://github.com/TanStack/table/discussions/3648#discussioncomment-15672794

PR #5471, opened 2024-04-09
fix(docs): delete unnecessary title, fix invalid links

steadily-worked:
hi, 

1. when i enter to [Columns Guide](https://tanstack.com/table/latest/docs/guide/column-defs), i found that links of `header groups`, `headers`, `footers` returns 404. so i fixed it.

2. remove unnecessary title inside [Headers Guide](https://tanstack.com/table/latest/docs/guide/headers).


KevinVandy:
There were a bunch of bad links as we transitioned from rewriting the docs in remix to TanStack Start, but these new links won't work. Closing in favor of other fixes.

PR #5499, opened 2024-04-19
fix: getIsSomeRowsSelected on all selection

iamswain25:
Selecting all rows and run `getIsSomeRowsSelected`, it return `false` instead of `true`.


sijonelis:
Is it possible to get this merged? The documentation [already states](https://tanstack.com/table/v8/docs/api/features/row-selection#getissomerowsselected) that this is how the row selection check should work!

packages/table-core/src/features/RowSelection.ts
```diff
@@ -436,7 +436,7 @@ export const RowSelection: TableFeature = {
       ).length
       return (
         totalSelected > 0 &&
-        totalSelected < table.getFilteredRowModel().flatRows.length
+        totalSelected <= table.getFilteredRowModel().flatRows.length
```
KevinVandy, line 439: This is not the correct fix. We should just get rid of all the logic that returns false if all rows are selected.

KevinVandy:
To truly fix this I think some breaking changes might be needed. I'm going to point this PR to the alpha branch probably

KevinVandy:
we implemented this in v9

PR #5506, opened 2024-04-23
Fix on[State]Change example

geryogam:


KevinVandy:
This doesn't make a difference?

geryogam:
Are you sure? Then why creating variable `newSortingValue` if we are not using it.

KevinVandy:
To show that you can do something with that value.

Issue #5352, opened 2024-02-16
When using virtual tables, high memory usage has always been unable to be reduced.


KonghaYao:
### TanStack Table version

v8.12.0

### Framework/Library version

Solid v1.8.1 and React v18.2.0

### Describe the bug and the steps to reproduce it

<img width="1125" alt="image" src="https://github.com/TanStack/table/assets/68177907/4b8f5b36-38d0-40f4-ad40-b698d2edeae3">

When I was trying to create a virtual table using tanstack, I accidentally discovered that it was using a very high amount of memory when scrolling, and even after GC, the page remained with high memory for about 3 minutes. Then I looked at the official React Demo and found the same issue.

But，when I use pure data and `@tanstack/virtual-core` to build a grid list, it is ok. So I think it's a @tanstack/solid-table's problem.

[Demo is here.](https://cn-ui.netlify.app/iframe.html?args=&id=data-%E6%95%B0%E6%8D%AE%E5%B1%95%E7%A4%BA-virtuallist-%E8%99%9A%E6%8B%9F%E5%88%97%E8%A1%A8--grid-render&viewMode=story)

[Souce Code is here](https://github.com/KonghaYao/cn-ui/blob/e2ce18c16cb2201434cb9715fe1644c738fad807/src/virtualList/VirtualGrid.tsx)

<img width="1450" alt="image" src="https://github.com/TanStack/table/assets/68177907/77aad96e-7ca4-43cc-b839-1d4b97df5778">

### Your Minimal, Reproducible Example - (Sandbox Highly Recommended)

https://tanstack.com/table/latest/docs/framework/react/examples/virtualized-columns

### Screenshots or Videos (Optional)

https://github.com/TanStack/table/assets/68177907/bca93236-b418-49db-b2f0-8df0aac3d5a9

### Do you intend to try to help solve this bug with your own PR?

No, because I do not have time to dig into it

### Terms & Code of Conduct

- [X] I agree to follow this project's Code of Conduct
- [X] I understand that if my bug cannot be reliable reproduced in a debuggable environment, it will probably not be fixed and this issue may even be closed.

KevinVandy:
memory usage us much lower in tanstack table v9 now

PR #5514, opened 2024-04-27
fix: reduce memory leak by remove memo function

KonghaYao:
I recently used table-core to complete a virtual table. I rendered it with a dataset of 1000*1000, and found that the memory usage of table-core was unreasonably high. Since no one had fixed this issue（#5352）, I attempted to debug it myself. Upon investigation, I discovered that memo was causing memory leaks in many places. This prevented JavaScript from reclaiming a massive number of rows and cells. Once I removed them, memory could be reclaimed again.

https://github.com/TanStack/table/assets/68177907/4ed24f52-c93f-48c4-bead-3d5d93ba5b80

### DIFF
--- a/packages/table-core/src/core/row.ts
+++ b/packages/table-core/src/core/row.ts
@@ -101,7 +101,7 @@ export const createRow = <TData extends RowData>(
   subRows?: Row<TData>[],
   parentId?: string
 ): Row<TData> => {
-  let row: CoreRow<TData> = {
+  const row: CoreRow<TData> = {
     id,
     index: rowIndex,
     original,
@@ -109,9 +109,9 @@ export const createRow = <TData extends RowData>(
     parentId,
     _valuesCache: {},
     _uniqueValuesCache: {},
-    getValue: columnId => {
-      if (row._valuesCache.hasOwnProperty(columnId)) {
-        return row._valuesCache[columnId]
+    getValue(columnId) {
+      if (this._valuesCache.hasOwnProperty(columnId)) {
+        return this._valuesCache[columnId]
       }
 
       const column = table.getColumn(columnId)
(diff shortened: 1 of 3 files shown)


KevinVandy:
removing `memo` in these places actually has a lot of performance implications here. If there is a memory problem, we should try to address that, but not at the expense of losing all of the render performance optimizations.

Also, this PR changes a lot of the code style for no reason, as far as I can tell. Why function() with `this`? Was that part of the memory leak fix, or just personal preference?

KonghaYao:
I'm very sorry. At first, I thought that creating a loop reference to itself during object creation would cause a memory leak. After removing the memo, I forgot to handle it. Now I have tested it and found that the reason for the memory leak was the use of memo. The new commit has removed the use of function and style.

KonghaYao:
Deleting the memo did indeed cause a high memory usage issue. I will investigate the reason for the memory leak caused by the memo.

KonghaYao:
In the example /examples/react/virtualized-columns, horizontal scrolling does not cause any memory issues, but vertical scrolling leads to memory leaks. The function getVisibleCells is called, which accesses several functions that I modified. However, the specific location and trigger of the memory leak have not yet been identified.

attaboy:
This problem (or a similar one) seems to be affecting our product as well. We've tried reducing a specific table down to just a single column with a simple string value in each row's cell, but with 200+ rows, we run into major memory problems that are making the page unusable by our customers — on some computers, the page crashes entirely.

KevinVandy:
In the next version of TanStack Table, it's going to be easy to opt out of memoization for any of the APIs. I'm not sure how we can address it in the current version right now though.

curvecatchdev:
We are also having the same issue, and if we garbage collect it once the table is rendered - it runs fine for a while. @KevinVandy in which version do you plan to release a flag to opt out of memoization btw? Can;t find it in the docs meaning it's not here yet, right?

raidenmiro:
any updates? 🤔

yxxung:
any updates?

KevinVandy:
in v9, you can use static function variants without memoization, but TanStack Table V9 also uses much less memory at scale in general. I don't anticipate anyone actually wanting to opt out of memoization anymore, but it's a lot easier now nonetheless.

PR #5519, opened 2024-04-29
ensure columnPinning.[property] is not accessed unsafely

rustysys-dev:
Had issues with the qwik examples etc... things just didn't really work at all, and it seems like unnecessary processing is being done in the lib.. 

In anycase, I decided to build my own table using the `core` lib directly for qwik.. however I am running into a bug, where if you provide an empty state to start off with the entire app crashes on trying to read the `left` property on `columnPinning` as shown:

image

I notice that most (not all) other functions in the code which do something similar are not accessing it safely with `columnPinning?.[property]`

### DIFF
--- a/packages/table-core/src/features/ColumnPinning.ts
+++ b/packages/table-core/src/features/ColumnPinning.ts
@@ -243,8 +243,8 @@ export const ColumnPinning: TableFeature = {
     row.getCenterVisibleCells = memo(
       () => [
         row._getAllVisibleCells(),
-        table.getState().columnPinning.left,
-        table.getState().columnPinning.right,
+        table.getState().columnPinning?.left,
+        table.getState().columnPinning?.right,
       ],
       (allCells, left, right) => {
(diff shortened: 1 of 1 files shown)


KevinVandy:
Maybe it would make more sense for us to make sure all of the typescript types enforce the full object is always present?

rustysys-dev:
@KevinVandy I think that would satisfy the requirement as well, but I suppose I should ask, is column pinning really required to be defined? If it must be defined for other features etc to work then enforcement via typescript is definitely best.

If not then a thought in the back of the head is that it would cause unnecessary development required when using the core lib.

KevinVandy:
> @KevinVandy I think that would satisfy the requirement as well, but I suppose I should ask, is column pinning really required to be defined? If it must be defined for other features etc to work then enforcement via typescript is definitely best.
> 
> If not then a thought in the back of the head is that it would cause unnecessary development required when using the core lib.

Yes, it is assumed that all of the `getInitialState` calls for each feature have been called successfully.

PR #5525, opened 2024-05-05
feat(angular-table): added injector optional parameter for more flexibility

merto20:
Following are the changes:

- Added optional `injector` parameter for more flexibility. Consumer can `createAngularTable` anywhere.

- Replace runInInjectionContext with an injector in the `effect`.  This has the same behaviour since only `effect` requires to be in the injector context.

- Remove unncessary untracked and signals.

- Remove proxifyTable as I think this is unncessary, since table is not required to be a signal. `laziInit` already proxified the table object. Pls correct me if I'm wrong.

- Refactored grouping example to include pagination  state.

### DIFF
--- a/examples/angular/grouping/src/app/app.component.ts
+++ b/examples/angular/grouping/src/app/app.component.ts
@@ -3,19 +3,19 @@ import {
   ChangeDetectionStrategy,
   Component,
   computed,
-  effect,
   signal,
 } from '@angular/core'
 import {
-  createAngularTable,
   FlexRenderDirective,
+  GroupingState,
+  PaginationState,
+  Updater,
+  createAngularTable,
   getCoreRowModel,
   getExpandedRowModel,
   getFilteredRowModel,
   getGroupedRowModel,
   getPaginationRowModel,
-  GroupingState,
-  Updater,
 } from '@tanstack/angular-table'
 import { columns } from './columns'
 import { makeData } from './makeData'
@@ -31,6 +31,7 @@ export class AppComponent {
   title = 'grouping'
   data = signal(makeData(10000))
   grouping = signal<GroupingState>([])
+  pagination = signal<PaginationState>({ pageIndex: 0, pageSize: 10 })
 
   stringifiedGrouping = computed(() => JSON.stringify(this.grouping(), null, 2))
 
@@ -39,6 +40,7 @@ export class AppComponent {
     columns: columns,
     state: {
       grouping: this.grouping(),
+      pagination: this.pagination(),
     },
     onGroupingChange: (updaterOrValue: Updater<GroupingState>) => {
       const groupingState =
@@ -46,6 +48,13 @@ export class AppComponent {
           ? updaterOrValue([...this.grouping()])
           : updaterOrValue
       this.grouping.set(groupingState)
+    },
+    onPaginationChange: (updaterOrValue: Updater<PaginationState>) => {
+      const paginationState =
+        typeof updaterOrValue === 'function'
+          ? updaterOrValue({ ...this.pagination() })
+          : updaterOrValue
+      this.pagination.set(paginationState)
     },
     getExpandedRowModel: getExpandedRowModel(),
     getGroupedRowModel: getGroupedRowModel(),
@@ -66,5 +75,6 @@ export class AppComponent {
 
   refreshData() {
     this.data.set(makeData(10000))
+    this.pagination.set({ pageIndex: 0, pageSize: 10 })
   }
 }
--- a/examples/angular/grouping/src/app/app.component.ts
+++ b/examples/angular/grouping/src/app/app.component.ts
@@ -33,7 +33,7 @@ export class AppComponent {
 
   stringifiedGrouping = computed(() => JSON.stringify(this.grouping(), null, 2))
 
-  table = createAngularTable(() => ({
+  tableOptions = computed(() => ({
     data: this.data(),
     columns: columns,
     state: {
@@ -53,6 +53,8 @@ export class AppComponent {
     getFilteredRowModel: getFilteredRowModel(),
     debugTable: true,
   }))
+
+  table = createAngularTable(this.tableOptions)
 
   onPageInputChange(event: any): void {
     const page = event.target.value ? Number(event.target.value) - 1 : 0
--- a/examples/angular/grouping/src/app/app.component.ts
+++ b/examples/angular/grouping/src/app/app.component.ts
@@ -8,7 +8,6 @@ import {
 import {
   FlexRenderDirective,
   GroupingState,
-  PaginationState,
   Updater,
   createAngularTable,
   getCoreRowModel,
@@ -31,7 +30,6 @@ export class AppComponent {
   title = 'grouping'
   data = signal(makeData(10000))
   grouping = signal<GroupingState>([])
-  pagination = signal<PaginationState>({ pageIndex: 0, pageSize: 10 })
 
   stringifiedGrouping = computed(() => JSON.stringify(this.grouping(), null, 2))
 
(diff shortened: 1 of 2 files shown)


KevinVandy:
Should this PR be made into the feat-angular-table branch instead of main?

KevinVandy:
Can you help review this @riccardoperra?

packages/angular-table/src/index.ts
```diff
@@ -1,96 +1,61 @@
...
-        if (firstRender) {
-          return (firstRender = false)
-        }
-        untracked(() => {
-          updateOptions()
-          notifier.set([])
-        })
-      })
+    // set table options again when options are updated
+    effect(() => updateOptions(), { injector })
```
riccardoperra, line 54: Consider that now this portion of code was present in order to avoid the twich `setOptions` call during the first table initialization.

Screenshot 2024-05-05 alle 19 54 09

packages/angular-table/src/index.ts
```diff
@@ -1,96 +1,61 @@
...
-          return table
-        },
-        {
-          equal: () => false,
-        }
-      )
+    // set table options for the first time
+    updateOptions()
 
-      return proxifyTable(tableSignal)
```
riccardoperra, line 93: Removing proxifyTable doesn't allow in any way to listen to the table changes if you are using computed/effect. In a more realistic example, you may risk to use `ngDoCheck` or `getters` instead of `computed`

```ts
effect(() => {
  this.table() // on every table option/state change
})

effect(() => {
  this.table.getState() // called on every state change
})

effect(() => {
  this.table.getCenterRows() // called on every center rows change
})

File: examples/angular/grouping/src/app/app.component.ts [hunk 3]
```diff
@@ -39,6 +40,7 @@ export class AppComponent {
     columns: columns,
     state: {
       grouping: this.grouping(),
+      pagination: this.pagination(),
```
riccardoperra, line 43: was this broken? I did this example like the `react` one, which doesn't use controlled pagination

KevinVandy:
I started a thread in the TanStack discord where this can be discussed more: https://discord.com/channels/719702312431386674/1003326939899113492/1236741193619083355

packages/angular-table/src/index.ts
```diff
@@ -1,96 +1,61 @@
...
 import { lazyInit } from './lazy-signal-initializer'
 
 export * from '@tanstack/table-core'
 
-export {
-  FlexRenderDirective,
-  FlexRenderComponent,
-  injectFlexRenderContext,
-} from './flex-render'
+export { FlexRenderDirective } from './flex-render'
```
riccardoperra, line 14: This should be reverted, both `FlexRenderComponent` and `injectFlexRenderContext` are needed to be able to render components via columns definition

packages/angular-table/src/index.ts
merto20: 'updateOptions' will be triggered twice everytime options is changed from outside. This is because 'effect' is listening to 2 signals. I was thinking of handling in all scenarios, if we want to fix this, instead of handling only the initialisation part.

examples/angular/grouping/src/app/app.component.ts
merto20: Table state should not changed if table data changes? With this PR, table states won't change when table data is reset so I added this resetting of pagination in this example.

riccardoperra: Following the react approach we should have a specific example for pagination and controlled pagination approach I think. wdyt @KevinVandy?

If you're not controlling pagination, the `onStateChange` which is responsible to update the table signal will be updated. That the reason we can't just return a simple `table` value and we have to wrap the methods / return a signal

packages/angular-table/src/index.ts
merto20: Got it. Working to put it back again. Thanks

packages/angular-table/src/index.ts
riccardoperra: Yeah you're right. I've also seen that currently the svelte adapter has the issue, since it's subscribing to each options and states.

packages/angular-table/src/index.ts
```diff
@@ -1,96 +1,78 @@
...
-        }
-      )
+    // set table options for the first time
+    updateOptions()
+
+    // convert table instance to signal for proxify to listen to any table state and options changes
+    const tableSignal = computed(
+      () => {
+        void [state(), resolvedOptionsSignal()]
+        return table
```
riccardoperra, line 68: I just have a doubt, considering that we have an effect which listen on both state and resolvedOptions and then we call `updateOptions`, does this `computed` get called **after** the table has been updated? The `notifier` variable was present in order to trigger this signal manually after all operations has been done

merto20: `effect` immediately executes while `computed` lazily executed

riccardoperra (approved):

riccardoperra:
@KevinVandy @merto20 I approved these changes but I just tried the "signal input" example which seems broken

KevinVandy:
@merto20 @riccardoperra So all conversations in this PR should be resolved. We're on the same page here?

One other thing, is that I want to make sure that the Angular table state docs are still accurate after these changes, or if we need to add more there to explain alternative approaches to state and signals. Either of you want to help there?

https://github.com/TanStack/table/pull/5432/files#diff-1c56bcf0dd442fe40b1fcf583f1b82bbbc2a24bb78a0d54201ae9d1ea9e36e7dR2

riccardoperra:
I would wait before merging this, the `injector` parameter is fine but there is a regression by this change related to the notifier signal.

- https://github.com/TanStack/table/pull/5525/files/363bac432c6a1364a8ba180d3ff76a601ed516e4#r1591242971

About the doc, the api changed a little (we now have to pass a function to the adapter), some examples should be rewrited. I may can look on it tomorrow evening

KevinVandy:
Ok, we'll wait for that regression fix, or if you know the exact fix, @riccardoperra , we can merge this and then a subsequent PR to fix that issue.

merto20:
@riccardoperra thanks for the video. I'm finishing up the fix for multiple `updateOptions` calls. Will push the changes in 1 hour.

merto20:
@riccardoperra I updated the `Grouping` example to show we can pass a signal as parameter to `createAngularTable`. Let me know if we need to create separate example project for this.

packages/angular-table/src/index.ts
```diff
@@ -1,96 +1,72 @@
...
-          return table
-        },
-        {
-          equal: () => false,
-        }
-      )
+    // convert table instance to signal for proxify to listen to any table state and options changes
+    const tableSignal = computed(
+      () => {
+        table.setOptions(updatedOptions())
```
riccardoperra, line 61: I’d go back with the previous implementation, updating the options inside a signal is theoretically wrong

merto20: I disagree. Doing side effects not related to value being returned by the computed signal is theoritically wrong. But updating the object before it will be return inside a computed is not. IMHO

merto20: I wanted to get rid of the `effect` from the start because of its high requirements. But was in doubt because I was not so sure if  I'm doing it wrong to call `updateOptions` method. Adding a `updatedOptions` signal simplifies that and more obvious that is is not wrong to update the Table object before it will returned in the computed signal. Calling `updateOptions` functions seems wrong because we cannot control if someone modifies that method and do something more than updating the table options.

riccardoperra:
@KevinVandy I htink this is ready to be merged 😄

KevinVandy (approved):

PR #5543, opened 2024-05-12
docs(angular-table): Add documentation for FlexRenderDirective

riccardoperra:
Hey there, in this PR i'm adding some documentation about rendering cell/header/footer content via FlexRender.

In addition to to render strings, we can also render TemplateRefs and Components, which is the corrispective of using a react jsx component into the cell. Hope everything is clear

### DIFF
--- a/docs/framework/angular/angular-table.md
+++ b/docs/framework/angular/angular-table.md
@@ -2,7 +2,8 @@
 title: Angular Table
 ---
 
-The `@tanstack/angular-table` adapter is a wrapper around the core table logic. Most of it's job is related to managing state the "angular signals" way, providing types and the rendering implementation of cell/header/footer templates.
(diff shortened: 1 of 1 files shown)


docs/framework/angular/angular-table.md
```diff
@@ -42,23 +50,168 @@ Example:
...
+    {
+      id: 'customCell',
+      header: () => this.customHeader(),
+      cell: () => this.customCell(),
+    },
+  ]
+}
+```
+
+#### Rendering a TemplateRef
```
KevinVandy, line 175: why does this have the same title as the heading above?

riccardoperra: wrong copy paste, fixed it

KevinVandy (approved):

PR #5545, opened 2024-05-12
docs(angular-table) Angular table state docs

riccardoperra:
I've updated the table state docs according to the angular implementation. 

I had to refactor a bit the example of the `Individual Controlled State` combining rxjs and signals instead of using `angular-query`, but I'm not sure about that. It's certainly a good way to make Angular Query known, but at the moment it's a library with the experimental flag. I didn't have any issue using it but I don't know if everything is clear to a user who sees this documentation 

I would like to get also some feedback from @merto20 about my `fully controlled state` approach, which in my opinion could differ from other implementations due to some constraints.

This is what the other frameworks does:

```jsx
const table = createSolidTable({
  columns,
  get data() {
    return data()
  },
  //... Note: `state` values are NOT passed in yet
})

const [state, setState] = createSignal({
  ...table.initialState
})

// **This is our constructor**
table.setOptions(prev => ({
  ...prev,
  get state() {
    return state()
  },
  onStateChange: setState
}))
```

This is what I did

```ts
class TableComponent {
  // create an empty table state, we'll override it later
  readonly state = signal({} as TableState);

  // create a table instance with default state values
  readonly table = createAngularTable(() => ({
    columns: this.columns,
    data: this.data(),
    // our fully controlled state overrides the internal state
    state: this.state(),
    onStateChange: updater => {
      // any state changes will be pushed up to our own state management
      this.state.set(
        updater instanceof Function ? updater(this.state()) : updater
      )
    }
  }))

  constructor() {
    // set the initial table state
    this.state.set({
      // populate the initial state with all of the default state values
      // from the table instance
      ...this.table.initialState,
      pagination: {
        pageIndex: 0,
        pageSize: 15, // optionally customize the initial pagination state.
      },
    })
  }
}
```

In angular, we may have to pass both `state` and `onStateChange` properties into the `createAngularTable` initialization. Using `setOptions` that introduces a new signal into the dependency graph breaks our rules to mark the view dirty. So the signal will be updated outside, but it's not reflected into the view and the `proxified` properties

### DIFF
--- a/docs/framework/angular/guide/table-state.md
+++ b/docs/framework/angular/guide/table-state.md
@@ -62,51 +62,62 @@ In order to control a particular state, you need to both pass in the correspondi
 
 Let's take filtering, sorting, and pagination as an example in a "manual" server-side data fetching scenario. You can store the filtering, sorting, and pagination state in your own state management, but leave out any other state like column order, column visibility, etc. if your API does not care about those values.
 
-```jsx
-readonly columnFilters = Angular.signal<ColumnFiltersState>([]) //no default filters
-readonly sorting = Angular.signal<SortingState[]>([
-  {
-    id: 'age',
-    desc: true, //sort by age in descending order by default
-  }
-])
-readonly pagination = Angular.signal<PaginationState>({
-  pageIndex: 0,
-  pageSize: 15
-})
+```ts
+import {signal} from '@angular/core';
+import {SortingState, ColumnFiltersState, PaginationState} from '@tanstack/angular-table'
+import {toObservable} from "@angular/core/rxjs-interop";
+import {combineLatest, switchMap} from 'rxjs';
 
-//Use our controlled state values to fetch data
-tableQuery = injectQuery(() => ({
-  queryKey: ['users', columnFilters.value, sorting.value, pagination.value],
-  queryFn: () => fetchUsers(columnFilters.value, sorting.value, pagination.value),
-  //...
-}))
+class TableComponent {
+  readonly columnFilters = signal<ColumnFiltersState>([]) //no default filters
+  readonly sorting = signal<SortingState[]>([
+    {
+      id: 'age',
+      desc: true, //sort by age in descending order by default
+    }
+  ])
+  readonly pagination = signal<PaginationState>({
+    pageIndex: 0,
+    pageSize: 15
+  })
 
-table = createAngularTable(() => ({
-  columns: this.columns,
-  data: tableQuery.data(),
-  //...
-  state: {
-    columnFilters: this.columnFilters(), //pass controlled state back to the table (overrides internal state)
-    sorting: this.sorting(),
-    pagination: this.pagination(),
-  },
-  onColumnFiltersChange: updater => { //hoist columnFilters state into our own state management
-    updater instanceof Function
-      ? this.columnFilters.update(updater)
-      : this.columnFilters.set(updater)
-  },
-  onSortingChange: updater => {
-    updater instanceof Function
-      ? this.sorting.update(updater)
-      : this.sorting.set(updater)
-  },
-  onPaginationChange: updater => {
-    updater instanceof Function
-      ? this.pagination.update(updater)
-      : this.pagination.set(updater)
-  },
-}))
+  //Use our controlled state values to fetch data
+  readonly data$ = combineLatest({
+    filters: toObservable(this.columnFilters),
+    sorting: toObservable(this.sorting),
+    pagination: toObservable(this.pagination)
+  }).pipe(
+    switchMap(({filters, sorting, pagination}) => fetchData(filters, sorting, pagination))
+  )
+  readonly data = toSignal(this.data$);
+
+  readonly table = createAngularTable(() => ({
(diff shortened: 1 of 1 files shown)


riccardoperra:
> In angular, we may have to pass both state and onStateChange properties into the createAngularTable initialization. Using setOptions that introduces a new signal into the dependency graph breaks our rules to mark the view dirty. So the signal will be updated outside, but it's not reflected into the view and the proxified properties

@KevinVandy @merto20 A solution I was thinking, but that need to be tried, is to override the setOptions behavior (or maybe add an implementation of `mergeOptions` which looks for all given getters, and handle them notifying our signals in orders to mark the view as dirty

KevinVandy (approved):

KevinVandy:
We should eventually just make the "fully controlled" example like the react one to go along with these docs. Our docs can evolve.

merto20:
@KevinVandy @riccardoperra Is it sensible to convert all documentation and examples to utilize signals? I am considering this if it proves to be a logical step.

merto20:
> I've updated the table state docs according to the angular implementation.
> 
> I had to refactor a bit the example of the `Individual Controlled State` combining rxjs and signals instead of using `angular-query`, but I'm not sure about that. It's certainly a good way to make Angular Query known, but at the moment it's a library with the experimental flag. I didn't have any issue using it but I don't know if everything is clear to a user who sees this documentation
> 
> I would like to get also some feedback from @merto20 about my `fully controlled state` approach, which in my opinion could differ from other implementations due to some constraints.
> 
> This is what the other frameworks does:
> 
> ```js
> const table = createSolidTable({
>   columns,
>   get data() {
>     return data()
>   },
>   //... Note: `state` values are NOT passed in yet
> })
> 
> const [state, setState] = createSignal({
>   ...table.initialState
> })
> 
> // **This is our constructor**
> table.setOptions(prev => ({
>   ...prev,
>   get state() {
>     return state()
>   },
>   onStateChange: setState
> }))
> ```
> 
> This is what I did
> 
> ```ts
> class TableComponent {
>   // create an empty table state, we'll override it later
>   readonly state = signal({} as TableState);
> 
>   // create a table instance with default state values
>   readonly table = createAngularTable(() => ({
>     columns: this.columns,
>     data: this.data(),
>     // our fully controlled state overrides the internal state
>     state: this.state(),
>     onStateChange: updater => {
>       // any state changes will be pushed up to our own state management
>       this.state.set(
>         updater instanceof Function ? updater(this.state()) : updater
>       )
>     }
>   }))
> 
>   constructor() {
>     // set the initial table state
>     this.state.set({
>       // populate the initial state with all of the default state values
>       // from the table instance
>       ...this.table.initialState,
>       pagination: {
>         pageIndex: 0,
>         pageSize: 15, // optionally customize the initial pagination state.
>       },
>     })
>   }
> }
> ```
> 
> In angular, we may have to pass both `state` and `onStateChange` properties into the `createAngularTable` initialization. Using `setOptions` that introduces a new signal into the dependency graph breaks our rules to mark the view dirty. So the signal will be updated outside, but it's not reflected into the view and the `proxified` properties

I would re-write the example to this:
```ts
class TableComponent {
  readonly pagination = signal<PaginationState>({ pageIndex: 0, pageSize: 15 })

  tableOptions = computed(() => ({
    data: this.data(),
    columns: columns,
    state: {
      pagination: this.pagination(),
    },
  }))

  table = createAngularTable(this.tableOptions)

  constructor() {}

  onNextPage() {
    const pagination = this.pagination()
    if (pagination.pageIndex < pagination.pageSize) {
      pagination.pageIndex++
      this.pagination.set({ ...pagination })
    }
  }

  onPrevPage() {
    const pagination = this.pagination()
    if (pagination.pageIndex > 0) {
      pagination.pageIndex--
      this.pagination.set({ ...pagination })
    }
  }
}
```

Passing the `onStateChange` handler when invoking `createAngularTable` seems unnecessary. Ideally, examples should avoid suggesting the use of `setOptions` or `table.initialState`, as these are managed internally.

riccardoperra:
You use onStateChange/state in order to have a single object with the entire state of the table, which allows you to have a full state controlled approach. We have also in another section the example to individually handle the state (https://tanstack.com/table/latest/docs/framework/angular/guide/table-state#individual-controlled-state)

Also, you may want to use next/back methods from tanstack table to update your signals in order to have an already working implementation to update the state instead of re-writing you from scratch, so you'll end to use the onStateChange/onPaginationChange handlers to synchronize them.

merto20:
I specifically mention `onStateChange`, not the individual state, since we are already handling the entire state internally.  For me, I don't see any necessity handling the `onStateChange` when I create table instance. 

We are always re-writing the state object, that's the only way we notify the table instance to update the options internally. The `computed` that listens to `options` will not be notified if its object address stays the same.

I use `pagination` as my example as I can control the values of my pagination and inform the table about the changes of the pagination. The flow is only 1 way, since the change is coming from my component not from the table. If I set the pagination change using `table.setPageIndex`, the flow is routed from the table, then to my component (through `onPaginationChange` handle) and going back to the table.

PR #5550, opened 2024-05-13
fix(angular-table): handle null and number values with flex-render

riccardoperra:
@KevinVandy grouping example was broken due to flexRenderDirective not handling null content (e.g. when a cell is grouped)

Screenshot 2024-05-13 alle 08 47 41

Here a screenshot of the grouping example after the fix


riccardoperra:
@jrgokavalsa Thanks I've forgot it. I've added a new check in order to support both string and numbers

KevinVandy (approved):

PR #5568, opened 2024-05-19
test(angular-table): add test case

riccardoperra:
I'm introducing some test cases to check the FlexRenderDirective and createAngularTable behaviors.

The setup is the same of @tanstack/angular-query, so we are currently using vitest and @analogjs plugin. We can may add in the future some test with angular-testing-library to do some rendering testing like the react package

I have found somes issues to the proxy implementation:
- we were transforming to computed signals also functions like `getCoreRowModel`
- Object.keys(table) and `key in table` were throwing an error due to the function prototype, so I've updated the proxy target to an arrow function

In a previous commit we also removed the Signal signature to the `createAngularTable` return type. I've reintroduced it since our table object is both signal and proxy


KevinVandy:
I can't figure out why I'm getting this error when running the project locally now

image

riccardoperra:
Fyi this seems a similar issue
https://github.com/angular/angular/issues/55709

riccardoperra:
@KevinVandy I don't know what happened, but I can confirm to you this is an issue related to some installed deps issue

After doing the below steps in order, I'm now able again to run the examples without any issues :/

1. remove all node_modules and build folder
```bash
find . -name 'node_modules' -type d -prune
find . -name 'build' -type d -prune
```
2. remove pnpm-lock.yaml
3. remove .nx folder

4. run `pnpm store prune --force`
5. run `pnpm install`
6. run `pnpm build`

I've had to recrete from scratch the pnpm-lock.yaml, and I updated all package.json in order to have the same installed angular version (17.3.9)

KevinVandy (approved):

PR #5578, opened 2024-05-23
Revert "fix(useTable): work around React 18 useReducer eager bailout change"

chatpong-v:
Reverts TanStack/table#4207


KevinVandy:
I'm not sure we had planned on releasing any more patches for v7. Is this absolutely necessary for it to work?

Issue #5363, opened 2024-02-19
`getValue` cache not invalidating when `accessorFn` is updated


danny-does-stuff:
### TanStack Table version

v8.12.0

### Framework/Library version

React v18.2.0

### Describe the bug and the steps to reproduce it

When the `accessorFn` for a column is updated to return a new value, the new value is never displayed in the table. Through a sophisticated series of `console.log`s, I discovered that when `getValue` is called on the rerender, the new `accessorFn` is never called, presumably due to caching on column id.

The use case is a column that displays the name of the person for the current row. We want the user to be able to choose whether to display first or last name first (e.g. `Tanner Linsley` or `Linsley, Tanner`). We could just change the column id when the `accessorFn` is updated, but then we lose the current sort state for that column, ie if the table is currently sorted by name, we want to stay sorted by name after the name format is changed.

Steps to reproduce:
1. Define a column where the `accessorFn` returns a different value based on some piece of state
2. Update that piece of state
3. The values in the column don't change

### Your Minimal, Reproducible Example - (Sandbox Highly Recommended)

https://stackblitz.com/edit/tanstack-table-rnjqkx?file=src%2Fmain.tsx

### Do you intend to try to help solve this bug with your own PR?

No, because I do not know how

### Terms & Code of Conduct

- [X] I agree to follow this project's Code of Conduct
- [X] I understand that if my bug cannot be reliable reproduced in a debuggable environment, it will probably not be fixed and this issue may even be closed.

badench:
Any updates on this? I'm seeing the same issue where a piece of state changes the value returned by the `accessorFn` but the new value isn't being shown in the table

andbjer:
We had the same issue in our project. The suggested fix in #5582 fixes the issue for us. We've applied the fix with `yarn patch` until it is merged and released.

aliceprobstexternalairbus:
Any news on this ? We're having the same issue.

PR #5582, opened 2024-05-25
fix: invalidate cached values after updating accessor functions

takoshi:
# What's this PR

https://github.com/TanStack/table/issues/5363

The above issue reported that `getValues` returns stale cached values after changing accessor functions.

This PR fixes it.

[8cd8449](https://github.com/TanStack/table/pull/5582/commits/8cd84491002e8033318165779d4abfaa9f2e6b13): add unit tests
[f66f4e6](https://github.com/TanStack/table/pull/5582/commits/f66f4e6ea6096355ddfb45695318840aa01da5c9): accessor functions are also held as cache, and it determines whether the cached data are fresh or stale.

### DIFF
--- a/packages/table-core/src/core/row.ts
+++ b/packages/table-core/src/core/row.ts
@@ -1,4 +1,4 @@
-import { RowData, Cell, Row, Table } from '../types'
+import { Cell, Row, RowData, Table } from '../types'
 import { flattenBy, getMemoOptions, memo } from '../utils'
 import { createCell } from './cell'
 
@@ -6,6 +6,7 @@ export interface CoreRow<TData extends RowData> {
   _getAllCellsByColumnId: () => Record<string, Cell<TData, unknown>>
   _uniqueValuesCache: Record<string, unknown>
   _valuesCache: Record<string, unknown>
+  _accessorFnsCache: Record<string, unknown>
   /**
    * The depth of the row (if nested or grouped) relative to the root row array.
    * @link [API Docs](https://tanstack.com/table/v8/docs/api/core/row#depth)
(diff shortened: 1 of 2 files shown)


EvanBarnesAZ:
Is this good to merge?

nickzelei:
Would love to see this merged!

lifeiscontent:
@tannerlinsley can we get some eyes on this? :)

packages/table-core/src/core/row.ts
```diff
@@ -107,19 +108,24 @@ export const createRow = <TData extends RowData>(
     original,
     depth,
     parentId,
+    _accessorFnsCache: {},
     _valuesCache: {},
     _uniqueValuesCache: {},
     getValue: columnId => {
-      if (row._valuesCache.hasOwnProperty(columnId)) {
+      const column = table.getColumn(columnId)
```
dylanhillier, line 115: The cost of this lookup is probably negligible, however the lookup isn't necessary if the following if-statement returns. Moving it to line 123 avoids this.

jaxxreal:
Hi @tannerlinsley, may you have a look at this one?

adamtowerz:
@KevinVandy , (sorry to tag, saw you're active merging fixes on this repo) is it possible to get a review on this PR? I'm running into this problem and would like to avoid running a fork if this has a path to merge to main

erlandsona:
I feel like I'm also running into this one fwiw.

aliceprobstexternalairbus:
Any news on this ? Any workaround maybe ? We're getting into the same issue.

PR #5603, opened 2024-06-14
refactor(svelte-table): improve type inference

thenbe:
`svelte-check` was having a difficult time inferring the props types when using svelte 5. This PR fixes that by swapping out the deprecated type utilities to improve type inference for `renderComponent(component, props)`.

Also updates svelte from `svelte@5.0.0-next.78` to `svelte@5.0.0-next.155`.

### DIFF
--- a/examples/lit/basic/package.json
+++ b/examples/lit/basic/package.json
@@ -17,8 +17,8 @@
     "lit": "^3.1.3"
   },
   "devDependencies": {
-    "@rollup/plugin-replace": "^5.0.5",
+    "@rollup/plugin-replace": "^5.0.7",
     "typescript": "5.4.5",
-    "vite": "^5.3.0"
+    "vite": "^5.3.1"
   }
 }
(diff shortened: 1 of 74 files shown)


KevinVandy:
@thenbe I can't get the sorting example types to compile, but also cannot get the error to appear in the editior.

thenbe:
Yeah I see the same issue. After running `pnpm update --latest` in the root of the repo and the sorting example types can compile again. So I guess some dependencies in the lockfile must be old.

thenbe:
I'm not sure how to find the exact culprit though.

thenbe:
The sorting example is the only one failing (out of the other svelte examples) is because it's the only that actually uses `renderComponent`.

thenbe:
Updated deps for svelte examples using `pnpm update --filter './examples/svelte/*'`. Both `build` and `test:types` now work in all svelte examples.

KevinVandy:
be sure to run `pnpm test` before pushing up. Still some type issues.

KevinVandy:
Think we got all the right versions upgraded repo wide now

KevinVandy (approved):

PR #5625, opened 2024-06-26
docs(examples): Add example using TanStack Query and Router

Balastrong:
Here's a slightly adapted version of the example shown on [youtube](https://youtu.be/F4zshDInsJY).

Finding a name wasn't easy, I think _Query Router Search Params_ fits with the other namings and highlights the key features but I'm happy to change it if we find a better one. Same goes for the code, all feedback is welcome :)


lachlancollins:
@KevinVandy FYI the prettier v4 alpha issue has been reported: https://github.com/prettier/prettier-cli/issues/14

PR #5657, opened 2024-07-12
fix(getGroupedRowModel): assign/unassign parentId 

PiR1:
Fixes https://github.com/TanStack/table/issues/4881
And Fixes https://github.com/TanStack/table/pull/4882

### DIFF
--- a/nx.json
+++ b/nx.json
@@ -1,6 +1,7 @@
 {
   "$schema": "./node_modules/nx/schemas/nx-schema.json",
   "defaultBase": "main",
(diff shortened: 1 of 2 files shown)


KevinVandy:
I'm wondering if this is quite right yet.

image

image

It appears that the rows with a depth of 2 are referencing the row at the depth of 0 as their parentId, just like the row with a depth of 1.

Using this console log
```ts
console.log('  '.repeat(row.depth), {
  name: row.original.firstName,
  id: row.id,
  parentId: row.parentId,
})
```

PR #5676, opened 2024-07-26
Fix getFacetedMinMaxValues

vincerubinetti:
I noticed a bug in getFacetedMinMaxValues that I'm surprised no one else has reported or the type checks haven't flagged. Actually, I'm sure I've run into this issue before, but I guess I never got around to reporting it...

The `getUniqueValues` function says it returns (and in testing, does return) an array of values. Thus, `firstValue` becomes e.g. `[123]`, and  `facetedMinMaxValues` will get initialized to `[[123], [123]]`, which sometimes results in the function returning e.g. `[123, [123]]`, depending on the row values. I'm guessing people don't always notice a bug here because an array is being `<`/`>` compared to a numbe, which I think coerces the array to a string?

My first commit makes the minimal change that I think fixes the issue. The second commit reworks the function to be a bit safer imo.

### DIFF
--- a/packages/table-core/src/utils/getFacetedMinMaxValues.ts
+++ b/packages/table-core/src/utils/getFacetedMinMaxValues.ts
@@ -10,11 +10,11 @@ export function getFacetedMinMaxValues<TData extends RowData>(): (
       () => [table.getColumn(columnId)?.getFacetedRowModel()],
       facetedRowModel => {
         if (!facetedRowModel) return undefined
-        
+
         const uniqueValues = facetedRowModel.flatRows
-          .flatMap((flatRow) => flatRow.getUniqueValues(columnId) ?? [])
+          .flatMap(flatRow => flatRow.getUniqueValues(columnId) ?? [])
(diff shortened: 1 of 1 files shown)


packages/table-core/src/utils/getFacetedMinMaxValues.ts
```diff
@@ -11,31 +11,22 @@ export function getFacetedMinMaxValues<TData extends RowData>(): (
...
-        let facetedMinMaxValues: [any, any] = [firstValue, firstValue]
-
-        for (let i = 0; i < facetedRowModel.flatRows.length; i++) {
-          const values =
-            facetedRowModel.flatRows[i]!.getUniqueValues<number>(columnId)
+        if (!uniqueValues.length) return
 
-          for (let j = 0; j < values.length; j++) {
-            const value = values[j]!
+        let facetedMinValue = uniqueValues.at(0)!
```
KevinVandy, line 21: might want to only use es2020 syntax here

PR #5687, opened 2024-08-01
feat: add support for reactivity in Vue-adapter

OlaAlsaker:
This PR updates the `useVueTable` composable in the TanStack Table Vue adapter to support reactive data. Now, when you pass a reactive ref to the data option, the table will automatically update when the data changes.

## Changes
- Extended the types to handle reactive data.
- Modified `useVueTable` to check if the data option is a ref and handle it correctly.
- Updated documentation with examples showing how to use the new feature.

## Example usage

```ts
const dataRef = ref([
  { id: 1, name: 'John' },
  { id: 2, name: 'Jane' }
])

const columns = [
  { accessor: 'id', Header: 'ID' },
  { accessor: 'name', Header: 'Name' }
]

const table = useVueTable({
  columns,
  data: dataRef, // Pass the reactive data ref
})

// Update dataRef later to automatically update the table
dataRef.value = [
  { id: 1, name: 'John' },
  { id: 2, name: 'Jane' },
  { id: 3, name: 'Doe' }
]
```

## Related issues or discussions

This will resolve discussion #5102


jariz:
Why not add support for the other options as well? as of writing I'm really struggling with making it respect changes to the `pageCount` property as well. 
It just will not use it's updated value, only after a hot reload it will actually update to the correct value. (clearly indicating it not correclty picking up changes to the options)
Adding a getter doesn't even do anything!

KevinVandy:
We've been looking into improvements for both Vue and Solid reactivity. Would want to confirm that this brings no breaking changes for v8.

Alternatively, we could have breaking changes if larger parts of the reactivity needed to be rethought for v9 in the alpha branch.

OlaAlsaker:
- [ ] Based on my simple (and limited) testing, this should not be a breaking change. The only thing it does is _allow_ and add support for reactivity, using Vue's `MaybeRef`. But it is not required, and the "old" ways should work as normal.

I have just started using this library, so I created this PR to fix some bugs I experienced when using JS getters (as per the documentation) together with Tanstack Query. Adding support for reactivity fixed my issue, so I made a PR for that 😄

I do agree that the library should have reactivity on the other options as well, and I might be able to fix that in another PR another time in the future 🚀

Maybe we should create an issue with a list of improvements that should be made to the Vue-adapter? I would be happy to have a look at it!

KevinVandy:
@OlaAlsaker @jariz I think we should look at making all possible table options that need to be reactive reactive, but possibly save it for v9. A pr to alpha could be made now for that though. The alpha branch is in heavy development right now.

Can ship the change for reactive data now.

9mm:
I don't know if it's this PR, or https://github.com/TanStack/table/pull/5694 this one, but this has made vue-datatable beyond usuable. My table crashes regularly in chrome with 100+ GB of memory. I have ~20,000-50,000 rows which is handled with absolute zero effort normally.

So i dont know if this is technically "breaking" but i had to revert this, and havent gotten a chance to look at how to actually turn this feature off, or if i need to make any changes. This is the current way I use it:


```javascript
    const table = useVueTable({
      get data() {
        return data.value;
      },
      get columns() {
        return columns.value;
      },
      state: {
        get expanded() {
          return expanded.value;
        },
        get sorting() {
          return sorting.value;
        },
        get rowSelection() {
          return rowSelection.value;
        },
      },
      initialState: {
        pagination: {
          pageIndex: 0,
          pageSize: 250,
        },
      },
      enableRowSelection: true,
      getExpandedRowModel: getExpandedRowModel(),
      getPaginationRowModel: getPaginationRowModel(),
      getRowId: row => row[rowKey.value],
      getSortedRowModel: getSortedRowModel(),
      getSubRows: row => row.sub_rows,
      onExpandedChange,
      onSortingChange,
      onRowSelectionChange,
      debugTable: false,
    });
```

OlaAlsaker:
> I don't know if it's this PR, or https://github.com/TanStack/table/pull/5694 this one, but this has made vue-datatable beyond unusable. My table crashes 100% of the time in chrome with 100+ GB of memory. I have ~20,000-50,000 rows which is handled with absolute zero effort normally.
> 
> 
> 
> IMO the benefit of tanstack table is this feature... it's the same as if i use a `shallowRef` when working with massive data objects. If i have nearly 100K rows I definitely dont want that array to be fully reactive. At the very least maybe we can have a `refresh` function similar to shallowRef, instead of making it fully reactive?
> 
> 
> 
> So yeah - I dont know if this is technically "breaking" but i had to revert this, and it definitely broke my current use case. I havent gotten a chance to look at how to actually turn this feature off or read the code/changes, but just wanted to make a comment now while im here reverting it. I'll look more in detail later. This is the current way I use it:
> 
> 
> 
> 
> 
> ```javascript
> 
>     const table = useVueTable({
> 
>       get data() {
> 
>         return data.value;
> 
>       },
> 
>       get columns() {
> 
>         return columns.value;
> 
>       },
> 
>       state: {
> 
>         get expanded() {
> 
>           return expanded.value;
> 
>         },
> 
>         get sorting() {
> 
>           return sorting.value;
> 
>         },
> 
>         get rowSelection() {
> 
>           return rowSelection.value;
> 
>         },
> 
>       },
> 
>       initialState: {
> 
>         pagination: {
> 
>           pageIndex: 0,
> 
>           pageSize: 250,
> 
>         },
> 
>       },
> 
>       enableRowSelection: true,
> 
>       getCoreRowModel: getCoreRowModel(),
> 
>       getExpandedRowModel: getExpandedRowModel(),
> 
>       getPaginationRowModel: getPaginationRowModel(),
> 
>       getRowId: row => row[rowKey.value],
> 
>       getSortedRowModel: getSortedRowModel(),
> 
>       getSubRows: row => row.sub_rows,
> 
>       onExpandedChange,
> 
>       onSortingChange,
> 
>       onRowSelectionChange,
> 
>       debugTable: false,
> 
>     });
> 
> ```

Thanks for the report! Can you check if the latest version (#5698) fixes the problem?

If not, I think I will just revert the reactivity-changess all together and instead push it to the v9 branch. 

This was of course not meant to be a breaking change. Sorry for the inconvenience 😇

(cc @KevinVandy )

romansp:
@OlaAlsaker first thanks for looking into adding reactivity support for Vue. It would be amazing to eventually get to a point where all table properties can be reactive (e.g. expanded, selection, etc.).

I quickly tried the latest version with reactive data support and something is still feels off about reactivity. Here's my stackblitz: https://stackblitz.com/edit/tanstack-table-dv5fcj?file=src%2FApp.vue&preset=node

1. Table doesn't update after new element is pushed to ref. 
2. Table doesn't update after new element was added with immutability.

But strangely table *will* update if data is coming from `computed` derived from original `ref`.

Although in my actual project data flows through computed, Table still doesn't seem to re-render even though computed value is changing. We use `vue-query` so I will try to create a minimal repro with vue-query as well.

OlaAlsaker:
> @OlaAlsaker first thanks for looking into adding reactivity support for Vue. It would be amazing to eventually get to a point where all table properties can be reactive (e.g. expanded, selection, etc.).
> 
> I quickly tried the latest version with reactive data support and something is still feels off about reactivity. Here's my stackblitz: https://stackblitz.com/edit/tanstack-table-dv5fcj?file=src%2FApp.vue&preset=node
> 
> 1. Table doesn't update after new element is pushed to ref.
> 2. Table doesn't update after new element was added with immutability.
> 
> But strangely table _will_ update if data is coming from `computed` derived from original `ref`.
> 
> Although in my actual project data flows through computed, Table still doesn't seem to re-render even though computed value is changing. We use `vue-query` so I will try to create a minimal repro with vue-query as well.

Thanks for report! Took a look at your Stackblitz, and looks like you have small error in your example:

```vue
  <button class="border p-2" @click="addRow">Add row</button>
  <button class="border p-2" @click="addRow">Add row immutable</button>
```

Both of your test-buttons use the same code. Also your function `addRowImmutable` is trying to reference `data.value`, which does not exist.

Test out your `addRowImmutable`-function (and fix the bug), and you will see that is works like expected 😄 

We use [`shallowRef`](https://vuejs.org/api/reactivity-advanced.html#shallowref) under the hood for performance reasons. This means that the data is not deeply reactive, only the `.value` is.

```ts
const data = ref([{ id: 1, name: 'John' }]);

// This will NOT work ❌
data.value.push({ id: 2, name: 'Doe' });

// This will work ✅
data.value = [...data.value, { id: 2, name: 'Doe' }];
```

9mm:
Hey @OlaAlsaker cool deal! I will have to try that later as I just came down with covid so now im lying in bed. Hopefully someone else is able to take over with help in the meantime. thanks for your quick response checking into it!

Scutua (approved):

PR #5694, opened 2024-08-05
fix: improve reactivity for data in Vue-adapter

OlaAlsaker:
Further improves reactivity for data in the Vue-adapter, which was introduced in #5687.

It mostly fixes some bugs and makes the reactivity more stable and simple. Also fixes a bug which would occur if initial `ref`-data was empty.

### DIFF
--- a/packages/vue-table/src/index.ts
+++ b/packages/vue-table/src/index.ts
@@ -1,9 +1,4 @@
-import {
-  TableOptions,
-  createTable,
-  TableOptionsResolved,
-  RowData,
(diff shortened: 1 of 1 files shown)


KevinVandy (approved):

9mm:
I made a comment here, not sure which commit might be related, this one or the original

https://github.com/TanStack/table/pull/5687#issuecomment-2276894591

Scutua (approved):

Scutua (approved):

PR #5698, opened 2024-08-07
fix: major performance bug in Vue-adapter

OlaAlsaker:
The `mergeOptions` causes huge performance issues in the Vue-adapter whenever the state of the table is updated. This is probably because it merges proxies within proxies, within proxies etc.

As far as I know, there is no need for any special handling of option-merging in the Vue-adapter. Removing it fixes the problem, and the library works like it should.

### DIFF
--- a/packages/vue-table/src/index.ts
+++ b/packages/vue-table/src/index.ts
@@ -13,7 +13,7 @@ import { mergeProxy } from './merge-proxy'
 
 export * from '@tanstack/table-core'
 
-type TableOptionsWithReactiveData<TData extends RowData> = Omit<
+export type TableOptionsWithReactiveData<TData extends RowData> = Omit<
   TableOptions<TData>,
   'data'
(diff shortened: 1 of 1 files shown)


KevinVandy:
@Mokshit06 Do you remember why a special proxy merge was needed in the vue adapter?

Mokshit06:
@KevinVandy If I remember correctly, It was so because without `mergeOptions` based as merging proxies, we would just spread the defaultOptions with the other options, making them no longer reactive. `mergeOptions` was creating a proxy of this so that it lazily accesses the latest value from the options.

@OlaAlsaker Can you check if the options still stay reactive even after this change?

OlaAlsaker:
I later found out that it was needed to make the "old" way work (using JS-getters and proxy). But when using Vue's reactivity, it is not needed. So I ended up just disabling it when the data is reactive 😄

PR #5725, opened 2024-08-24
feat: update deprecated type change

tobySolutions:
Worked with @chibuike-19 to update `never` types in the type ternaries across the core table code with Jsdoc's `@deprecated` block tags by creating interfaces for the  unavailable options.

- [x] TableState
- [ ] TableOptions
- [ ] ColumnDefBase


packages/table-core/src/features/column-filtering/ColumnFiltering.types.ts
```diff
@@ -222,6 +222,16 @@ export interface TableOptions_ColumnFiltering<
 > extends ColumnFiltersOptionsBase<TFeatures, TData>,
     ResolvedFilterFns<TFeatures, TData> {}
 
+export interface TableOptions_ColumnFiltering_Unavailable<
```
KevinVandy, line 225: I think we're going to have to make much more in-depth changes for table options. This PR can focus just on the state type changes.

KevinVandy: I'm checking out the branch and removing this part. The needed change in a future pr will be to mark each of the `ColumnFiltersOptionsBase` interface properties as deprecated with jsdoc, though still with their type, when their associated feature is not present

tobySolutions: Sorry, how do you mean @KevinVandy? Should we do away with all work done on `TableOptions`? I wasn't really sure of our work there though.

Is it okay to merge now with all that work stripped out?

KevinVandy (approved):

packages/table-core/src/features/column-filtering/ColumnFiltering.types.ts
KevinVandy: I just reverted those parts and added the commit to this pr

tobySolutions: Awesome! Thanks Kevin!!

PR #5791, opened 2024-11-05
Fix tsx in column-sizing.md

allevo:


KevinVandy:
this was meant to be an opening tag

PR #5815, opened 2024-11-23
Fix solid adapter in alpha version

riccardoperra:
Reintroduce the `mergeOptions` property which is needed to merge props without losing reactivity when calling `setOptions`

### DIFF
--- a/packages/solid-table/src/createTable.ts
+++ b/packages/solid-table/src/createTable.ts
@@ -23,17 +23,18 @@ export function createTable<
     getInitialTableState(_features, tableOptions.initialState),
   )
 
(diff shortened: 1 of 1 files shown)


packages/solid-table/src/createTable.ts
```diff
@@ -38,6 +41,7 @@ export function createTable<
   createComputed(() => {
     table.setOptions((prev) => {
       return mergeProps(prev, tableOptions, {
+        _features: { ...coreFeatures, ...tableOptions._features },
```
riccardoperra, line 44: Not sure about this, without that I think the new options will not contains the coreFeatures

KevinVandy (approved):

PR #5817, opened 2024-11-24
feat(angular-table): Support Tanstack v9 alpha and Angular 19, fix examples

riccardoperra:
I'm updating the Angular adapter and related examples to support the new helpers and `_features` property. It's mostly finished, but here's some points:

## injectTable types

Due to the new TableOptions types, I've seen that there are some typescript issues while passing the _rowModels property, probably since we pass a function as an argument and _rowModels is using `NoInfer` :/

<details>
<summary>
The error
</summary>

Screenshot 2024-11-24 alle 14 48 15
</details>

Screenshot 2024-11-24 alle 14 50 24

Overriding the properties in that way seems resolving the issue, but could there be some contraindications doing this? Would it make sense to have a less strict type?

```ts
export type AngularTableOptions<
  TData extends RowData,
> = Omit<TableOptions<TFeatures, TData>, '_rowModels'> & {
  _rowModels?: CreateRowModels_All<TFeatures, TData>
  // This type is not currently exported, but it should have the same issues
  // _rowModelsFns: RowModelFns<TFeatures, TData>
}
```

## constructTableHelper/createTableHelper

Since we need to pass a `function` as an argument to the `injectTable/createAngularTable` fn, I think for Angular we should have our own implementation of the core "constructTableHelper" function.

In the future we might need to update this function to be invoked within the context of the angular app, for example if we need to create the helper as a constant and still allowing to inject data/options from a token or service.

I've then added a `constructTableHelper.ts` file the updated signature. This helper is used internally by `createTableHelper`, which is exported publicly.

---

Continue on https://github.com/TanStack/table/pull/5816 for the new proxy implementation


examples/angular/row-selection-signal/src/app/filter.ts
```diff
@@ -39,10 +39,10 @@ import type { Column, Table } from '@tanstack/angular-table'
   standalone: true,
   imports: [CommonModule],
 })
-export class FilterComponent<T> implements OnInit {
+export class FilterComponent<T extends RowData> implements OnInit {
```
KevinVandy, line 42: Might be changing these generics to include TFeatures since it effects what APIs can be used in the logic of this component.

packages/angular-table/src/constructTableHelper.ts
```diff
@@ -0,0 +1,109 @@
...
+  ) => Table<TFeatures, TData> & Signal<Table<TFeatures, TData>>,
+  tableHelperOptions: TableHelperOptions<TFeatures, TData>,
+): TableHelper_Core<TFeatures, TData> {
+  const { TData: _TData, ..._tableHelperOptions } = tableHelperOptions
+  return {
+    columnHelper: createColumnHelper<TFeatures, TData>(),
+    createColumnHelper,
+    features: tableHelperOptions._features,
+    options: _tableHelperOptions as any,
+    tableCreator: (tableOptions) =>
```
KevinVandy, line 66: If this is the only difference, we might explore just making some TS overloads and an option for how it needs to work for Angular in the table-core `constructTableHelper` function. Good to be here for now.

KevinVandy (approved):

Issue #5822, opened 2024-12-03
Pinning aggregation rows then ungrouping causes a hard error


Mike-Dax:
### TanStack Table version

8.20.5

### Framework/Library version

React 17.0.2

### Describe the bug and the steps to reproduce it

This reproduction is a combination of the pinning and grouping example sandboxes.

1. Group some rows, eg 'last names' by clicking on the "open hand emoji".
2. Pin a grouped aggregation row
3. Ungroup the rows.

In this code:

https://github.com/TanStack/table/blob/6b4d616dd7c8917616eb4fecaf09dda7030fd115/packages/table-core/src/features/RowPinning.ts#L233-L247C30

`table.getRow(rowId, true)` is called on an aggregation row that no longer exists, and the component errors ungracefully. 

### Your Minimal, Reproducible Example - (Sandbox Highly Recommended)

https://stackblitz.com/edit/tanstack-table-s94mcy?file=src%2Fmain.tsx

### Do you intend to try to help solve this bug with your own PR?

Yes, I am also opening a PR that solves the problem along side this issue

### Terms & Code of Conduct

- [x] I agree to follow this project's Code of Conduct
- [x] I understand that if my bug cannot be reliable reproduced in a debuggable environment, it will probably not be fixed and this issue may even be closed.

PR #5823, opened 2024-12-03
fix: crash when ungrouping after aggregation row was pinned

Mike-Dax:
This is a fix for https://github.com/TanStack/table/issues/5822


KevinVandy:
fixed in v9

PR #5837, opened 2024-12-18
Fix/tanstack v9 feature test fixes

will-short:
- updated table-core/tsconfig to include /tests directory
- fixes lint and type errors in core/tests/
- fixes test cases so all pass in core/tests/

@KevinVandy Ran into a couple things I had questions on but wasn't sure the best place to ask them. figured we could just have a thread here. 
1. Is it expected to need to pass `coreFeatures` to `constructTable`? - without it lots of core methods I was expecting to work did not exist
2. tried to keep these mostly the same and just update them to work with v9. let me know if there was anything you wanted changed or added to these files and I'm happy to update.

### DIFF
--- a/packages/table-core/tests/RowPinning.test.ts
+++ b/packages/table-core/tests/RowPinning.test.ts
@@ -1,25 +1,27 @@
-import { describe, expect, it } from 'vitest'
+import { beforeEach, describe, expect, it } from 'vitest'
 import {
-  ColumnDef,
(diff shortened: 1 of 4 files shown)


KevinVandy:
This is great. We should discuss your questions maybe in the discord table-v9 channel. Your questions address some decisions that I have not totally decided on, actually.

KevinVandy (approved):

PR #5841, opened 2024-12-21
refactor: fixes and moves all tests in core to new structure

will-short:
- moves all tests in core package to new structure
- removes old helpers
- adds TODO's in existing tests to get them to use new helpers
- fixes up all existing tests to pass


packages/table-core/tests/unit/core/table/constructTable.test.ts
```diff
@@ -37,7 +38,6 @@ describe('constructTable', () => {
     expect(table).toHaveProperty('getRow')
 
     // table APIs
-    expect(table).toHaveProperty('_queue')
```
will-short, line 40: @KevinVandy seems like _queue was removed.  just wanted to flag this to make sure I saw that correctly

KevinVandy: yeah, a couple of apis were removed

KevinVandy (approved):

PR #5849, opened 2024-12-27
chore: starting to test filter types

tylersayshi:


examples/react/filters/src/main.tsx
```diff
@@ -42,6 +43,8 @@ const _features = tableFeatures({
   rowPaginationFeature,
 })
 
+type TestType = ExtractFeatureTypes<typeof _features, 'Column'>
```
tylersayshi, line 46: This is currently unknown. It seems like more needs to be done to override the interface of the Columns maybe, is this correct?

KevinVandy:
Where would attest be installed?

KevinVandy:
maybe another branch that actually uses the `ExtractFeatureTypes` should be made? We'll uncomment that code in that branch, run attest on both the alpha branch and that one.

tylersayshi:
> Where would attest be installed?

Probably the root package.json devDeps. This way we can test examples, packages/* and anything else from there.

I don't have any strong opinion on this though. I did a global install of it locally and I am just sampling as I `cd` around the repo right now.

tylersayshi:
> maybe another branch that actually uses the `ExtractFeatureTypes` should be made? We'll uncomment that code in that branch, run attest on both the alpha branch and that one.

yeah, this would be really helpful. Is it just a matter of uncommenting all the commented out types? 

I was unsure if not all of them were commented for the same reasons or not.

KevinVandy:
assuming we want to close this one

PR #5851, opened 2024-12-30
refactor tdata to infer from full list then use list items after

tylersayshi:
this is WIP and the goal of this is a few things:

- reduce the surface area of where TData is inferred from to just the `data` field
- allow other fields to be less specific, but use TData when it is inferred
  - currently the issue with _rowModels is that it is less specific but demands to
    still match TData from data. The fix should be to not use the return type of
    `createFilteredRowModel(filterFns)`

it's a significant refactor, that I plan to re-evaluate before marking the PR ready,
but I think the goals of inferring TData from just `data` and allowing other fields to
be looser without affecting TData are each worthwile here.

The part I will re-evaluate is just organization and thinking a bit more critically about when we should use TDataList vs TData. For now, the goal is to get this approach working.


packages/angular-table/src/constructTableHelper.ts
```diff
@@ -14,13 +14,13 @@ import type { Signal } from '@angular/core'
  */
 export type TableHelperOptions<
   TFeatures extends TableFeatures,
-  TData extends RowData = any,
+  TDataList extends Array<RowData> = Array<any>,
```
KevinVandy, line 17: just checking, it shouldn't be `= Array<RowData>`? instead of `any`?

tylersayshi: I was carrying over the `= any` from before for now. I'm not yet sure what is correct here though.

tylersayshi:
@KevinVandy this works and passes `pnpm build` now!

I need to iterate on it, by:
- hopefully cleaning things up/refactoring to simplify
- test the changes (including benchmarking)

Feel free to read through the changes so far though or try them out! I'm happy that this is working haha

KevinVandy:
Looks like you were able to simplify it a lot?

tylersayshi:
> Looks like you were able to simplify it a lot?

A lot from the initial version for sure 👍 gonna take another look at the coffee shop in the morning. Happy new year!

KevinVandy:
I was trying to experiment and build a little bit on top of this branch. Ran into an issue when TData is not an array like we are assuming in this branch.

Here is the TData type:

image

It can work if we get rid of the Record in the Union. It's probably a < 1% use-case anyway.

tylersayshi:
Can you share the branch?

KevinVandy:
> Can you share the branch?

Decided to leave it alone for now. This only became a visible problem internally when I had tried to use the same TDataList strategy on the `Table_Internal` type when trying something out.

Take a look at this branch: https://github.com/TanStack/table/tree/tdata-to-list-and-static-bagged
or the commit: https://github.com/TanStack/table/commit/b24171856d1deebf704e175f14e3b07cd1a8da77
where I think I just fixed a bunch of typescript performance problems by moving "bagged" generics to static interfaces. I think the TS server can cache the results of the interfaces once, so better performance

Issue #5767, opened 2024-10-01
Angular:  Issue with FlexRenderComponent not updating row references in context (injectFlexRenderContext)


d9nny:
### TanStack Table version

v8.20.5

### Framework/Library version

Angular 17

### Describe the bug and the steps to reproduce it

Description:

Expected behaviour: When using FlexRenderComponent and injectFlexRenderContext in column defs (like a checkbox for row selection), the row context should update reactively in sync with the selection model.

Actual behaviour: The model updates correctly (e.g., selection count is accurate), but the row reference in the context does not, causing methods like row.getIsSelected() to return stale data (I assume).

Workaround: Rendering the checkboxes directly in the template without FlexRenderComponent works as expected.

Reproduce:

It can be seen in your example
https://tanstack.com/table/latest/docs/framework/angular/examples/row-selection

### Your Minimal, Reproducible Example - (Sandbox Highly Recommended)

https://tanstack.com/table/latest/docs/framework/angular/examples/row-selection

### Do you intend to try to help solve this bug with your own PR?

None

### Terms & Code of Conduct

- [x] I agree to follow this project's Code of Conduct
- [x] I understand that if my bug cannot be reliable reproduced in a debuggable environment, it will probably not be fixed and this issue may even be closed.

riccardoperra:
@d9nny Was looking into the issue now and tried to check it.

Does the error still persist? I'm not able to reproduce this in the examples on my local machine (from both dev server or production build) and the applications where I have tanstack. This could be a change detection issue related to the OnPush strategy.

riccardoperra:
@d9nny @KevinVandy this issue has been fixed with the latest update. also example is working now

PR #5856, opened 2025-01-02
feat(angular-table): Refactor Flex render implementation - Zoneless, Better type safety, allows reactive values into cell content, re-render when cell context changes, allow to pass signal inputs into custom components

riccardoperra:
I refactored how the FlexRender directive applies updates internally during change detection. This will resolve issues like #5767 without necessarily waiting for v9 (so this is not a breaking change for users). It now should also work with zoneless and it's reactive if someone use signals into the cell content. Also there are some dx improvements that helps writing column custom content easily

**tldr**
- [Updated vitest config with new analog version. Resolve some issues we had with tests. Add some "integration" tests. Old tests are all passing with the newest implementation without changes](https://github.com/TanStack/table/pull/5856#update-vitest-configuration:~:text=Update%20vitest%20configuration)
- [Better type safety](https://github.com/TanStack/table/pull/5856#update-vitest-configuration:~:text=Enhanced%20Type%20Safety%20for%20FlexRenderComponent)
- [The cell definition content may update more than before, support zoneless and signals into cell definition](https://github.com/TanStack/table/pull/5856#update-vitest-configuration:~:text=Refactor%20FlexRenderDirective%20rendering%20mechanism) 
- Some code has been moved into separate files

For those who enjoy reading 😅:

## Update vitest configuration 

I've updated vitest test setup with latest version of @analog/vitest-angular and @analog/vite-plugin-angular in order to compile the angular components correctly. 

We are now able to use angular signal inputs during test. I removed all workaround utils we had to use before. Now we could also opt for switching the FlexRenderDirective to signal inputs instead of old-based @Input decorators, even if we cannot rely on effects, we should always use ngOnChanges).

## Enhanced Type Safety for FlexRenderComponent 

The FlexRenderComponent inputs are now typed based on the given component instance
```typescript
@Component()
class MyComp {   
  readonly input1 = input<string>();
  readonly input2 = input.required<number>();
}

new FlexRenderComponent(MyComp, { ... }) // correctly typed as { input1: string | undefined, input2: number }
```

I have also added a `flexRenderComponent` function util that can be used to enhance typesafety and which use an option object instead of multiple arguments to pass data. This allows us to add new arguments in the future without breaking that signature (e.g. in the future we will probably need to bind angular "outputs").

In v9 we may want to have only one of the two available to avoid confusion?

```typescript
@Component()
class Example {   
  readonly input1 = input<string>();
}

//  Example doesn't have required inputs, so 2nd argument is optional. `input1` is typed as `string | undefined`
flexRenderComponent(Example)

@Component()
class ExampleRequired {   
  readonly input1 = input.required<string>();
}

//  Example doesn't have required inputs, so 2nd argument is mandatory. `input1` is typed as `string`
flexRenderComponent(ExampleRequired, { inputs: { input1: "value" } )


```

## Injection context for cell content

Cell content now runs within an injection context. This allows users to inject services and using signals into the cell definition, even outside angular components.

```typescript
// file outside angular
const columns = [ 
  {
      accessorFn: row => row.lastName,
      id: 'lastName',
      header: () => {
        return inject(MyService).returnSomething(...)
      }
  }
]
```

## Refactor FlexRenderDirective rendering mechanism

[tldr](https://github.com/TanStack/table/pull/5856#update-vitest-configuration:~:text=Now%2C%20we%20can%20do%20the%20same%20directly%20into%20the%20column%20definitions)

This is the biggest part. With the previous implementation, we were only relying on `ngOnChanges` hook.

`ngOnChanges` is called by Angular when any data-bound property change, but it's not enough for some reasons:
- `content` property is a reference of the cell/header definition, which could be a function. Doing `content(props)` may return different values, but since we are passing a reference, Angular cannot evaluate it for us automatically.
- `props` most of the time will be `(cell|header).getContext()`, which is memoized for cell (but no headers). Then even in this case the reference may not change

When table data change, we are sure that at least both properties changes. Currently this is the only case were we are re-rendering the cell view.

With this PR, first I am introducing some logic to split some rendering-related code on separate classes (those are not exposed to the library consumers):
- `FlexRenderComponentRef`: a wrapper for Angular ComponentRef, which expose some internal utils to detect changes, set inputs via angular `setInput` api etc.
- `FlexRenderComponentFactory`: an angular service that is responsible to create instances of FlexRenderComponentRef.
- `FlexRenderView`: an abstract class that encapsulate the view state and some information, and defines some logic that we invoke in specific phases of the view checker. It has two implementations:
  - `FlexRenderTemplateView`: a class that define update methods for content rendered via EmbeddedViewRef (which we use to render primitives like string, number or even TemplateRef instances)
  - `FlexRenderComponentView`: a class that define update methods for content rendered via FlexRenderComponentRef (ComponentRef)
 - Code related to FlexRenderComponent injectable context has been moved into a separate file (those declarations are exported for library consumers, but they have not changed)

I've then revisited the rendering business logic via `ngDoCheck` hook and implementing Bit fields to handle the rendering phases.

Link: [angular-table/src/flex-render/flags.ts](https://github.com/riccardoperra/table/blob/feat/angular-flex-render-granular-update/packages/angular-table/src/flex-render/flags.ts)

```ts
/**
 * Flags used to manage and optimize the rendering lifecycle of content of the cell,
 * while using FlexRenderDirective.
 */
export enum FlexRenderFlags {
  /**
   * Indicates that the view is being created for the first time or will be cleared during the next update phase.
   * This is the initial state and will transition after the first ngDoCheck.
   */
  ViewFirstRender = 1 << 0,
  /**
   * Represents a state where the view is not dirty, meaning no changes require rendering updates.
   */
  Pristine = 1 << 1,
  /**
   * Indicates the `content` property has been modified or the view requires a complete re-render.
   * When this flag is enabled, the view will be cleared and recreated from scratch.
   */
  ContentChanged = 1 << 2,
  /**
   * Indicates that the `props` property reference has changed.
   * When this flag is enabled, the view context is updated based on the type of the content.
   *
   * For Component view, inputs will be updated and view will be marked as dirty.
   * For TemplateRef and primitive values, view will be marked as dirty
   */
  PropsReferenceChanged = 1 << 3,
  /**
   * Indicates that the current rendered view needs to be checked for changes.
   */
  DirtyCheck = 1 << 4,
  /**
   * Indicates that a signal within the `content(props)` result has changed
   */
  DirtySignal = 1 << 5,
  /**
   * Indicates that the first render effect has been checked at least one time.
   */
  RenderEffectChecked = 1 << 6,
}
```

Regardless of the implementation, there are two new behaviors:

#### Initializing an effect during the first render 
During the view first render, we initialize an `effect` that set the `DirtySignal` flag every time the `content(props)` changes, if signals are used into the cell definition. This allows to react to changes even without ngZone, and opens up some possibilities in v9 to further improve rendering, for example detecting also when the `table` state signals changes

#### Checking changes in ngDoCheck

During the ngDoCheck lifecycle hook, we determine whether the view needs to be updated by evaluating a set of flags and comparing the result of content(props) with its previous value. At the moment, this check occurs during every doCheck cycle. However, in a future we could use store slices (v10?+) and signals (v9 with the reactivity feature i was working on) to improve this phase updating more efficently.

---

### What changes for users

The new implementation is retrocompatible with previous versions, but improves dx of columns definition while using Angular since this let now to render content conditionally, using signals inside etc.

Most of the time the code I saw to render custom content was something like this, splitting how the table is rendered in both template and typescript

```html
<tbody>
  @for (row of table.getRowModel().rows; track row.id) {
    <tr>
      @for (cell of row.getVisibleCells(); track cell.id) {
        <td>
          @switch (cell.id) {
            <!-- this could have be done also in column def, but what if we want to add some custom logic? -->
            @case ('user') {
              {{ cell.row.original.firstName }}
              {{ cell.row.original.lastName }}
            }
            @case ('status') {
              @if (colors().length) { 
                <app-colored-badge [colors]="colors()" [status]="cell.row.original.status"></app-colored-badge>
              } @else {
                 {{ 'status'. + cell.row.original.status | translate }}
              }
            }
           <!-- This is for columns that can just use the `renderValue` :/ -->
            @default {
              <ng-container
                *flexRender="
                  cell.column.columnDef.cell;
                  props: cell.getContext();
                  let cell
                "
              >
                <div [innerHTML]="cell"></div>
              </ng-container>
            }
          }
        </td>
      }
    </tr>
  }
</tbody>
```

Now, we can do the same directly into the column definitions 😄 

```ts
const helper = createColumnHelper<Person>()
const columns = [
  helper.display({
    id: 'user',
    cell: context => {
      const { firstName, lastName } = context.row.original
      // Just an example of a service that may concat non-empty string
      const stringHelper = inject(StringHelper)
      return stringHelper.joinNonEmpty([firstName, lastName], ' ')

      // or maybe i want to render a component in some specific conditions
      const { firstName, lastName, email } = context.row.original
      if (!email) { 
        return `${firstName} ${lastName}`
      }
      return new FlexRenderComponent(
        CustomCell, 
        { content: `${firstName} ${lastName}`, tooltip: email }
      )
    },
  }),
  helper.accessor('status', {
    id: 'status',
    cell: ({ getValue }) => {
      const status = getValue()
      // this may be a signal, then everything here will be revaluated also when its value change.
      const colors = appColors()
      if (!colors) {
        // just an example
        return translate(`status.${status}`)
      }
      return new FlexRenderComponent(AppColoredBadge, {
        status,
        colors,
      })
    },
  }),
]
```

### DIFF
--- a/docs/framework/angular/angular-table.md
+++ b/docs/framework/angular/angular-table.md
@@ -40,39 +40,169 @@ FlexRender supports any type of content supported by Angular:
 - A [TemplateRef](https://angular.dev/api/core/TemplateRef)
 - A [Component](https://angular.dev/api/core/Component) wrapped into `FlexRenderComponent`
 
-Example:
+You can just use the `cell.renderValue` or `cell.getValue` APIs to render the cells of your table. However,
+these APIs will only spit out the raw cell values (from accessor functions).
+If you are using the `cell: () => any` column definition options, you will want to use the `FlexRenderDirective` from the adapter.
+
+Cell column definition is **reactive** and runs into an **injection context**, then you can inject services or make use of signals to automatically modify the rendered content.
+
+#### Example
 
 ```ts
 @Component({
   imports: [FlexRenderDirective],
   //...
 })
+class YourComponent {}
 ```
 
 ```angular-html
 
 <tbody>
 @for (row of table.getRowModel().rows; track row.id) {
-<tr>
-  @for (cell of row.getVisibleCells(); track cell.id) {
-<td>
-  <ng-container
-    *flexRender="
+  <tr>
+    @for (cell of row.getVisibleCells(); track cell.id) {
+      <td>
+        <ng-container
+          *flexRender="
               cell.column.columnDef.cell;
               props: cell.getContext();
               let cell
             "
-  >
-    <!-- if you want to render a simple string -->
-    {{ cell }}
-    <!-- if you want to render an html string -->
-    <div [innerHTML]="cell"></div>
-  </ng-container>
-</td>
-}
-</tr>
+        >
+          <!-- if you want to render a simple string -->
+          {{ cell }}
+          <!-- if you want to render an html string -->
+          <div [innerHTML]="cell"></div>
+        </ng-container>
+      </td>
+    }
+  </tr>
 }
 </tbody>
+```
+
+#### Rendering a Component
+
+To render a Component into a specific column header/cell/footer, you can pass a `FlexRenderComponent` instantiated with
+your `ComponentType, with the ability to include parameters such as inputs and an injector.
+
+```ts
+import {flexRenderComponent} from "./flex-render-component";
+import {ChangeDetectionStrategy} from "@angular/core";
+
+@Component({
+  template: `
+    ...
+  `,
+  standalone: true,
+  changeDetectionStrategy: ChangeDetectionStrategy.OnPush
+})
+class CustomCell {
+  readonly content = input.required<string>();
+
+  readonly type = input<MyType>()
+}
+
+class AppComponent {
+  columns: ColumnDef<unknown>[] = [
+    {
+      id: 'custom-cell',
+      header: () => {
+        const translateService = inject(TranslateService);
+        return translateService.translate('...');
+      },
+      cell: (context) => {
+        return flexRenderComponent(
+          MyCustomComponent,
+          {
+            injector, // Optional injector
+            inputs: {
+              // Mandatory input since we are using `input.required()
+              content: context.row.original.rowProperty,
+              type // Optional input
+            }
+          }
+        )
+      },
+    },
+  ]
+}
+```
+
+Underneath, this utilizes
+the [ViewContainerRef#createComponent](https://angular.dev/api/core/ViewContainerRef#createComponent) api.
+Therefore, you should declare your custom inputs using the @Input decorator or input/model signals.
+
+You can still access the table cell context through the `injectFlexRenderContext` function, which returns the context
+value based on the props you pass to the `FlexRenderDirective`.
+
+```ts
+
+@Component({
+  // ...
+})
+class CustomCellComponent {
+  // context of a cell component
+  readonly context = injectFlexRenderContext<CellContext<TData, TValue>>();
+  // context of a header/footer component
+  readonly context = injectFlexRenderContext<HeaderContext<TData, TValue>>();
+}
+```
+
+Alternatively, you can render a component into a specific column header, cell, or footer by passing the component type
+to the corresponding column definitions. These column definitions will be provided to the `flexRender` directive along
+with the `context`.
+
+```ts
+class AppComponent {
+  columns: ColumnDef<Person>[] = [
+    {
+      id: 'select',
+      header: () => TableHeadSelectionComponent<Person>,
+      cell: () => TableRowSelectionComponent<Person>,
+    },
+  ]
+}
+```
+
+```angular-html
+<ng-container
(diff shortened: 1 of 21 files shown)


KevinVandy:
I think we need to do something to fix this: 

image

We have the typescript version pinned, but it's still installing the wrong version in some peer deps somehow.

cc @riccardoperra @lachlancollins

riccardoperra:
@KevinVandy pr is ready 😄

riccardoperra:
@KevinVandy pr is now ready. I've also added a fix to proxy that was preventing to have `getRowModel` reactive in some cases.

KevinVandy (approved):

PR #5868, opened 2025-01-12
fix: inject data type to features

tylersayshi:
this removes the need for the union to intersection.

there are a few more type errors that I'll look to solve today after my walk 🐕‍🦺 🚀


tylersayshi:
marking draft to fix the issues created since rebasing alpha

tylersayshi:
closing this - my approach here was flawed

Issue #5026, opened 2023-08-14
Can't perform a React state update on a component that hasn't mounted yet


tacomanator:
### Describe the bug

Error received when navigating to a table using `getGroupedRowModel()` such as:

```ts
  const table = useReactTable({
    data,
    columns,
    getCoreRowModel: getCoreRowModel(),
    getGroupedRowModel: getGroupedRowModel(),
  });
```

As seen in this example, navigating to the basic table works fine where as navigating to the grouped table produces an error in the console. The only difference between the two is `getGroupedRowModel()`.
 
I am using the code from the examples pretty much as is (stripped down for simplicity).

### Your minimal, reproducible example

https://github.com/tacomanator/tanstack-table-unmounted

### Steps to reproduce

1. Clone example, install dependencies, and run
2. While viewing the console, navigate to the grouped table
3. Observe error

### Expected behavior

No error output

### How often does this bug happen?

Every time

### Screenshots or Videos

https://github.com/TanStack/table/assets/63794/30edbbcf-8f7e-459f-b624-b7d82aac3f0a

### Platform

macOS, Chrome 115.0.5790.170 arm64

### react-table version

8.9.3

### TypeScript version

5.1.6

### Terms & Code of Conduct

- [X] I agree to follow this project's Code of Conduct
- [X] I understand that if my bug cannot be reliable reproduced in a debuggable environment, it will probably not be fixed and this issue may even be closed.

andrelillvede:
I am having the same problem in version 8.9.1 when I call table.getRowModel() to conditionally render rows in tbody and I can confirm that it probably has something to do with getGroupedRowModel.

If i remove getGroupedRowModel: getGroupedRowModel() from the hook the error disappears.

timredband:
I'm seeing this exact same issue today.

timredband:
I investigated this issue this morning and discovered that the error only occurs when in dev mode. Same with @tacomanator's example. When I run the example in dev mode I get the error, but if I do a `npm run build` and `npm start` then the error disappears.

bastibuck:
I think this is due to the fact that these errors don't show up in production builds but still happen. 

Happy to be proven wrong.

timredband:
I think it has something to do with pagination. When you add `getGroupedRowModel` it ends up calling [setPagination](https://github.com/TanStack/table/blob/main/packages/table-core/src/features/Pagination.ts#L107) in the render and that ends up calling `setState`. 

[...]

PR #5879, opened 2025-01-19
fix(react-table): fix improper state update on unmounted component

jmtoung:
This commit addresses https://github.com/TanStack/table/issues/5026/, where react-table is performing a state update on an unmounted component. To address this, we add a useEffect to track whether the component on which the table is referenced is mounted.


jmtoung:
I verified this fixes the console.error message React spits out during StrictMode in my local setup

alexander-densley:
any updates on this pr?

spencermarx:
Would love a status update on this when possible 🙂

Lantianyou:
@tannerlinsley would be lovely if this PR is accepted.

PR #5884, opened 2025-01-21
fix dynamic data updates in the Lit Table Adapter

lschierer:
this fixes an issue I discussed in discord where with the lit table 
adapter, updating a data array did not get reflected by the table.

It is a one-line change to the TableController, and a new example that demonstrates the difference.

### DIFF
--- a/packages/lit-table/src/index.ts
+++ b/packages/lit-table/src/index.ts
@@ -55,6 +55,7 @@ export class TableController<TData extends RowData>
     this.tableInstance.setOptions(prev => ({
       ...prev,
       state: { ...this._tableState, ...options.state },
+      data: options.data,
       onStateChange: (updater: any) => {
         this._tableState = updater(this._tableState)
         this.host.requestUpdate()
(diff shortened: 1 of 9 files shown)


KevinVandy:
@kadoshms We've had problems in the Lit adapter with data not being reactive. This fixes it. Wonder if we need to make more/all table options reactive? Maybe at least columns too?

lschierer:
I'm fairly sure making the columns reactive is a similar one line fix, but I didn't test that. I didn't need it.

lschierer:
and many thanks to @KevinVandy  for walking me through this.

kadoshms:
LGTM

packages/lit-table/src/index.ts
```diff
@@ -55,6 +55,7 @@ export class TableController<TData extends RowData>
     this.tableInstance.setOptions(prev => ({
       ...prev,
       state: { ...this._tableState, ...options.state },
+      data: options.data,
```
kadoshms, line 58:
```suggestion
      data: options.data,
      columns: options.columns,
```

KevinVandy (approved):

PR #5890, opened 2025-01-28
Add `header.getAfter` API

joshuajaco:
To allow column pinning on the right side with header groups there should be a `header.getAfter` API, similar to `header.getStart` and `column.getAfter`.

I also copied the unused `position` parameter of `header.getStart`, not sure if this is needed or why it is there in the first place.


KevinVandy:
added in v9

PR #5915, opened 2025-02-10
refactor: remove unnecessary code

pavan-sh:
- Removed `HTMLAttributes` from imports in the React row selection example as it was not being used.
- Removed `ref` from the dependency array of the `IndeterminateCheckbox` function in the React row selection and expanding examples, as React refs don't change and are not needed in the dependency array.


pavan-sh:
@KevinVandy could you please review this PR I raised a while ago? Thanks.

pavan-sh:
Hi @KevinVandy.. can you please take a look? its been a while

PR #5918, opened 2025-02-12
fix: broken hyperlink in row-selection.md for getRowId

pavan-sh:
Fixed broken hyperlink for getRowId link, which previously led to a 404 page.


pavan-sh:
@KevinVandy could you please review this PR I raised a while ago? Thanks.

PR #5919, opened 2025-02-14
chore(svelte-table): cleanup package.json

benmccann:
Svelte only supports ESM, so there's no reason to list CJS or UMD files in the `package.json`


benmccann:
did not realize there's an alpha. will close this and fix it there

PR #5921, opened 2025-02-14
feat(angular): improvements to angularReactivityFeature, remove proxies

riccardoperra:
Some experiments with `angularReactivityFeature`

Previously createAngularTable/injectTable return value was a proxy into a proxy :
- lazyInit -> needed to support required input signals. This basically schedule the table object in a microtask
- proxifyTable -> needed to override `table` properties and make them reactive when the table signal changes

Using proxifyTable doesn't directly lead in issues after the new FlexRender implementation, but it's just a different object compared to the one obtained via (header|cell).getContext(), which is not reactive.

Thanks to table features added in v9, we could move away from `proxyTable` by just overriding table/cell/header/column/row props recursively via `TableFeature` interface.

This allows to make all properties reactive via the same logic used by the table proxy right now using `toComputed`.

Overriding all properties can be costly because if there are a lot of elements, a potential reactive signal would be created for each property of each header/row/column/cell. To work around this, we can disable reactive properties via table options.

---

- `skipBaseProperties` now check via string indexs instead of endsWith/startsWith to improve performance a little when there are many rows
- Removed `enableExperimentalReactivity` with a new `reactivity` object that allows to make reactive properties based on the type of object 
- `setReactiveProp` will now override with a getter that will create the computed lazily. This may improve performance a bit since previously a computed was created synchronously, while now it is created only once the getter is invoked for the first time


riccardoperra:
<p align="center"></p>

riccardoperra:
<p align="right"></p>

riccardoperra:
<p align="center"></p>

riccardoperra:
aaaaaa

Issue #5926, opened 2025-02-20
Performance issue: `createRow` in tanstack/table-core uses excessive memory


afreix:
### TanStack Table version

v8.21.2

### Framework/Library version

tanstack/table-core

### Describe the bug and the steps to reproduce it

`createRow` in row.ts ([src](https://github.com/TanStack/table/blob/db745afdb867511ddd04a79150f89d3894a03532/packages/table-core/src/core/row.ts#L95)) allocates new closures for _every_ row in the table. Each table features also allocate closures for each row. These functions could be defined once on a shared prototype.

We've played around with a [few solutions](https://github.com/mleibman-db/table/commit/2d6e91620a3f02b97c2bbb6b0a74624a1fa7bfdb), namely using a `WeakMap<Table, rowProto>` in order to support a shared prototype for all rows in a given table. The `WeakMap` appears to be necessary given the dynamic application of table features (i.e. we can't share a row prototype across all table instances).

A JS Heap snapshot shows a **memory reduction** from 141mb --> 11.6mb in `examples/vanilla/basic` example with [this change](https://github.com/mleibman-db/table/commit/2d6e91620a3f02b97c2bbb6b0a74624a1fa7bfdb). Note: there are a few small issues to fix. We expect it to be productionalizable

It's worth reiterating that users pay the cost regardless of whether they interact with the table, since it happens during initialization.

We'd like to collaborate with the maintainers on productionalizing ideas defined in [this commit](https://github.com/mleibman-db/table/commit/2d6e91620a3f02b97c2bbb6b0a74624a1fa7bfdb). Thank you!

### Your Minimal, Reproducible Example - (Sandbox Highly Recommended)

Reproduction:
- apply [this commit](https://github.com/mleibman-db/table/commit/2d6e91620a3f02b97c2bbb6b0a74624a1fa7bfdb) and run the `examples/vanilla/basic` example
- compare JS heap dumps before/after

Attached are screenshots of the before/after that we observed locally:

**Before**
![Image](https://github.com/user-attachments/assets/3cf32cb1-6a48-4871-ad23-75b75a0b4c83)

**After**
![Image](https://github.com/user-attachments/assets/a058a5ed-1d94-456e-91e6-93424c5fa3be)

### Screenshots or Videos (Optional)

An image of tanStack table initialization closures accounting for 59% of retained memory in our application:

![Image](https://github.com/user-attachments/assets/b1ab8284-9ace-42b0-bdbf-7f7e81d128e2)

[...]

PR #5927, opened 2025-02-21
perf, memory: Improve performance and memory use for large datasets

mleibman-db:
This PR moves all duplicate row instance methods to row proto, drastically reducing memory use for large datasets.
Fixes issue #5926.

For a table of 50'000x5, this reduces table memory use from **136 Mb** to **4.8 Mb** (**28x**).
The initialization time (`accessRows()`) is also reduced from **132ms** to **18ms** (**7x**).
It also speeds up grouping (using modified `getGroupedRowModel.test.ts`) from **1169ms** to **276ms** (**4x**).

**Before**
before

**After**
after

**Deltas**
delta

### DIFF
--- a/packages/table-core/src/core/row.ts
+++ b/packages/table-core/src/core/row.ts
@@ -102,119 +102,114 @@ export function getRowProto<TData extends RowData>(table: Table<TData>) {
   let rowProto = rowProtosByTable.get(table)
 
   if (!rowProto) {
-    const obj: CoreRow<TData> = {
-      // props are here only for typing; they are set on the instance at runtime
-      id: 'unused',
-      depth: 0,
-      index: -1,
-      original: undefined as TData,
-      subRows: [],
-      _valuesCache: {},
-      _uniqueValuesCache: {},
+    const proto = {} as CoreRow<TData>
 
-      getValue(columnId: string) {
-        if (this._valuesCache.hasOwnProperty(columnId)) {
-          return this._valuesCache[columnId]
-        }
+    // Make the default fallback value available on the proto itself to avoid duplicating it on every row instance
+    // even if it's not used. This is safe as long as we don't mutate the value directly.
+    proto.subRows = [] as const
 
-        const column = table.getColumn(columnId)
+    proto.getValue = function (columnId: string) {
+      if (this._valuesCache.hasOwnProperty(columnId)) {
+        return this._valuesCache[columnId]
+      }
 
-        if (!column?.accessorFn) {
-          return undefined
-        }
+      const column = table.getColumn(columnId)
 
-        this._valuesCache[columnId] = column.accessorFn(
-          this.original as TData,
-          this.index
-        )
+      if (!column?.accessorFn) {
+        return undefined
+      }
 
-        return this._valuesCache[columnId] as any
-      },
+      this._valuesCache[columnId] = column.accessorFn(
+        this.original as TData,
+        this.index
+      )
 
-      getUniqueValues(columnId: string) {
-        if (!this.hasOwnProperty('_uniqueValuesCache')) {
-          // lazy-init cache on the instance
-          this._uniqueValuesCache = {}
-        }
-
-        if (this._uniqueValuesCache.hasOwnProperty(columnId)) {
-          return this._uniqueValuesCache[columnId]
-        }
-
-        const column = table.getColumn(columnId)
-
-        if (!column?.accessorFn) {
-          return undefined
-        }
-
-        if (!column.columnDef.getUniqueValues) {
-          this._uniqueValuesCache[columnId] = [this.getValue(columnId)]
-          return this._uniqueValuesCache[columnId]
-        }
-
-        this._uniqueValuesCache[columnId] = column.columnDef.getUniqueValues(
-          this.original as TData,
-          this.index
-        )
-
-        return this._uniqueValuesCache[columnId] as any
-      },
-
-      renderValue(columnId: string) {
-        return this.getValue(columnId) ?? table.options.renderFallbackValue
-      },
-
-      getLeafRows() {
-        return flattenBy(this.subRows, d => d.subRows)
-      },
-
-      getParentRow() {
-        return this.parentId ? table.getRow(this.parentId, true) : undefined
-      },
-
-      getParentRows() {
-        let parentRows: Row<TData>[] = []
-        let currentRow = this
-        while (true) {
-          const parentRow = currentRow.getParentRow()
-          if (!parentRow) break
-          parentRows.push(parentRow)
-          currentRow = parentRow
-        }
-        return parentRows.reverse()
-      },
-
-      getAllCells: memo(
-        function (this: Row<TData>) {
-          return [this, table.getAllLeafColumns()]
-        },
-        (row, leafColumns) => {
-          return leafColumns.map(column => {
-            return createCell(table, row, column, column.id)
-          })
-        },
-        getMemoOptions(table.options, 'debugRows', 'getAllCells')
-      ),
-
-      _getAllCellsByColumnId: memo(
-        function (this: Row<TData>) {
-          return [this.getAllCells()]
-        },
-        allCells => {
-          return allCells.reduce(
-            (acc, cell) => {
-              acc[cell.column.id] = cell
-              return acc
-            },
-            {} as Record<string, Cell<TData, unknown>>
-          )
-        },
-        getMemoOptions(table.options, 'debugRows', 'getAllCellsByColumnId')
-      ),
+      return this._valuesCache[columnId] as any
     }
 
-    rowProtosByTable.set(table, obj)
-    rowProto = obj
+    proto.getUniqueValues = function (columnId: string) {
+      if (!this.hasOwnProperty('_uniqueValuesCache')) {
+        // lazy-init cache on the instance
+        this._uniqueValuesCache = {}
+      }
+
+      if (this._uniqueValuesCache.hasOwnProperty(columnId)) {
+        return this._uniqueValuesCache[columnId]
+      }
+
+      const column = table.getColumn(columnId)
+
+      if (!column?.accessorFn) {
+        return undefined
+      }
+
+      if (!column.columnDef.getUniqueValues) {
+        this._uniqueValuesCache[columnId] = [this.getValue(columnId)]
+        return this._uniqueValuesCache[columnId]
+      }
+
+      this._uniqueValuesCache[columnId] = column.columnDef.getUniqueValues(
+        this.original as TData,
+        this.index
+      )
+
+      return this._uniqueValuesCache[columnId] as any
+    }
+
+    proto.renderValue = function (columnId: string) {
+      return this.getValue(columnId) ?? table.options.renderFallbackValue
+    }
+
+    proto.getLeafRows = function () {
+      return flattenBy(this.subRows, d => d.subRows)
+    }
+
+    proto.getParentRow = function () {
+      return this.parentId ? table.getRow(this.parentId, true) : undefined
+    }
+
+    proto.getParentRows = function () {
+      let parentRows: Row<TData>[] = []
+      let currentRow = this
+      while (true) {
+        const parentRow = currentRow.getParentRow()
+        if (!parentRow) break
+        parentRows.push(parentRow)
+        currentRow = parentRow
+      }
+      return parentRows.reverse()
+    }
+
+    proto.getAllCells = memo(
+      function (this: Row<TData>) {
+        return [this, table.getAllLeafColumns()]
+      },
+      (row, leafColumns) => {
+        return leafColumns.map(column => {
+          return createCell(table, row, column, column.id)
+        })
+      },
+      getMemoOptions(table.options, 'debugRows', 'getAllCells')
+    )
+
+    proto._getAllCellsByColumnId = memo(
+      function (this: Row<TData>) {
+        return [this.getAllCells()]
+      },
+      allCells => {
+        return allCells.reduce(
+          (acc, cell) => {
+            acc[cell.column.id] = cell
+            return acc
+          },
+          {} as Record<string, Cell<TData, unknown>>
+        )
+      },
+      getMemoOptions(table.options, 'debugRows', 'getAllCellsByColumnId')
+    )
+
+    rowProtosByTable.set(table, proto)
+    rowProto = proto
   }
 
   return rowProto as CoreRow<TData>
--- a/packages/table-core/src/core/row.ts
+++ b/packages/table-core/src/core/row.ts
@@ -108,7 +108,7 @@ export function getRowProto<TData extends RowData>(table: Table<TData>) {
     proto.clone = function () {
       return Object.assign(Object.create(Object.getPrototypeOf(this)), this)
     }
-      
+
     // Make the default fallback value available on the proto itself to avoid duplicating it on every row instance
     // even if it's not used. This is safe as long as we don't mutate the value directly.
     proto.subRows = [] as const
--- a/packages/table-core/src/core/row.ts
+++ b/packages/table-core/src/core/row.ts
@@ -92,6 +92,136 @@ export interface CoreRow<TData extends RowData> {
   subRows: Row<TData>[]
 }
 
+const rowProtosByTable = new WeakMap<Table<any>, any>()
+
+
+/**
+ * Creates a table-specific row prototype object to hold shared row methods, including from all the
+ * features that have been registered on the table.
+ */
+export function getRowProto<TData extends RowData>(table: Table<TData>) {
+  let rowProto = rowProtosByTable.get(table)
+
+  if (!rowProto) {
+    const obj: CoreRow<TData> = {
+      // props are here only for typing; they are set on the instance at runtime
+      id: 'unused',
+      depth: 0,
+      index: -1,
+      original: undefined as TData,
+      subRows: [],
+      _valuesCache: {},
+      _uniqueValuesCache: {},
+
+
+      getValue(columnId: string) {
+        if (this._valuesCache.hasOwnProperty(columnId)) {
+          return this._valuesCache[columnId]
+        }
+
+        const column = table.getColumn(columnId)
+
+        if (!column?.accessorFn) {
+          return undefined
+        }
+
+        this._valuesCache[columnId] = column.accessorFn(
+          this.original as TData,
+          this.index
+        )
+
+        return this._valuesCache[columnId] as any
+      },
+
+      getUniqueValues(columnId: string) {
+        if (!this.hasOwnProperty('_uniqueValuesCache')) {
+          // lazy-init cache on the instance
+          this._uniqueValuesCache = {};
+        }
+
+        if (this._uniqueValuesCache.hasOwnProperty(columnId)) {
+          return this._uniqueValuesCache[columnId]
+        }
+
+        const column = table.getColumn(columnId)
+
+        if (!column?.accessorFn) {
+          return undefined
+        }
+
+        if (!column.columnDef.getUniqueValues) {
+          this._uniqueValuesCache[columnId] = [this.getValue(columnId)]
+          return this._uniqueValuesCache[columnId]
+        }
+
+        this._uniqueValuesCache[columnId] = column.columnDef.getUniqueValues(
+          this.original as TData,
+          this.index
+        )
+
+        return this._uniqueValuesCache[columnId] as any
+      },
+
+      renderValue(columnId: string) {
+        return this.getValue(columnId) ?? table.options.renderFallbackValue
+      },
+
+      getLeafRows() {
+        return flattenBy(this.subRows, d => d.subRows)
+      },
+
+      getParentRow() {
+        return this.parentId ? table.getRow(this.parentId, true) : undefined
+      },
+
+      getParentRows() {
+        let parentRows: Row<TData>[] = []
+        let currentRow = this
+        while (true) {
+          const parentRow = currentRow.getParentRow()
+          if (!parentRow) break
+          parentRows.push(parentRow)
+          currentRow = parentRow
+        }
+        return parentRows.reverse()
+      },
+
+      getAllCells: memo(
+        function (this: Row<TData>) {
+          return [this, table.getAllLeafColumns()]
+        },
+        (row, leafColumns) => {
+          return leafColumns.map(column => {
+            return createCell(table, row, column, column.id)
+          })
+        },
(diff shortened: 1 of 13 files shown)


KevinVandy:
This idea is awesome, and it's something that I've been thinking about for a long time.

However, it's a bigger change and I would personally like to see this PR targeted at alpha so we can test the pattern there and even take this idea further.

V9 probably won't be stable for a few more months, but I hope that you still would be able to help implement this there.

We should follow this same pattern for not only rows, but also cells, headers, and header groups. The table object is the only one where I don't see this as necessary since there is only 1 table created.

What do you think?

mleibman-db:
@KevinVandy Thank you for the quick response! 

Is there any way we can apply this fix to V8 (pretty please)?

This is a bit of a pressing issue for us. We use the table in Notebooks in our product (Databricks), where it's not unusual to have dozens of tables on a page with up to 10k rows each, which results in the page taking up multiple gigabytes of memory, with over 70% of it coming from the table. We have some ideas for workarounds in our code, but they are somewhat brittle/hacky, and I would much prefer to contribute to improving the table itself. 

We could increase the scope of this change and apply the pattern more broadly (I saw a lot of areas where we could improve scalability wrt efficiency), but my goal here was to make a minimally invasive fix w/o changing the API that gets us most of the way there. As is, this PR reduces table memory use by ~10x, which is more than enough to address our current needs. 

In terms of applying the same pattern to headers and header groups, yes, that would have a similar impact on tables with a very large number of columns, but that is a lot less common than tables with lots of rows. We're talking about tens of thousands of columns here. It does happen, but is something that can easily be added in a separate PR.

Applying this to cells is not as effective however, since cells are created on-demand as they are rendered, so unlike rows, they don't use up memory unless you render (or scroll through in case of virtualization) tens of thousands of cells.

KevinVandy:
I'm definitely still open to merging this to v8, I'll just need to do some extensive regression testing

mleibman-db:
@KevinVandy I don't do much OSS development, so I'm going to need some help / hand-holding here :) Do I need to do a separate PR to apply the changes to the alpha branch? Is that instead of this one, or in addition to? Not sure what I need to do here.

Re: doing the same thing for `column` objects. As I mentioned, in most cases it wouldn't be as impactful since tables with tens of thousands of columns are much less rare than tables with lots of rows, but I'm happy to make that change as well. I'd probably do that in a separate PR though to limit the scope.

KevinVandy:
The scope of this pr to the main v8 branch is fine/good.

However, the alpha v9 branch has been heavily refactored with new approaches to assigning APIs to these objects. In the v9 alpha branch, I'd hope to find an approach that follows this new strategy for everything as much as possible.

mleibman-db:
Ok, so IIUIC, I'll leave this PR as-is to proceed with code review, testing, and inclusion in v8, and will look at v9 to see how things are different there and what I need to do to re-apply them there.

KevinVandy:
> Ok, so IIUIC, I'll leave this PR as-is to proceed with code review, testing, and inclusion in v8, and will look at v9 to see how things are different there and what I need to do to re-apply them there.

That would be awesome. I realize the follow up for the v9 alpha work is a big extra ask, but hopefully a fun and interesting thing for you to look at helping us out. It will need a slightly different approach.

One of the main goals of v9 is to strip down the bundle sizes (and memory usage) of table instances down to just the features that apps are actually using. This PR is very much on theme for that.

mleibman-db:
Will do!

packages/table-core/src/core/row.ts
```diff
@@ -92,6 +92,134 @@ export interface CoreRow<TData extends RowData> {
...
+  if (!rowProto) {
+    const obj: CoreRow<TData> = {
+      // props are here only for typing; they are set on the instance at runtime
+      id: 'unused',
+      depth: 0,
+      index: -1,
+      original: undefined as TData,
+      subRows: [],
+      _valuesCache: {},
+      _uniqueValuesCache: {},
```
KevinVandy, line 113: Just wondering if we might get in trouble by storing these values on the proto? The wisdom I heard last time I worked on stuff using the prototypes was to only store functions on the prototype

packages/table-core/src/features/ColumnFiltering.ts
```diff
@@ -411,6 +403,23 @@ export const ColumnFiltering: TableFeature = {
 
       return table._getFilteredRowModel()
     }
+
+    Object.assign(getRowProto(table), {
```
KevinVandy, line 407: At one point, we had removed most usages of `Object.assign` in favor of direct assignment as a performance improvement at scale. Wonder if that's still applicable to consider here.

mleibman-db: It wouldn't be an issue here since it's only called once per table anyway. Your question would apply more to `createRow()` in `row.ts` since we call it once per row there, but AFAIK, there are no known performance issues around `Object.assign()`. There have been some many years ago when it was just introduced and browser support was fresh (plus there were polyfills), but that hasn't been the case in quite some time.

packages/table-core/src/core/row.ts
thomasmarr: @KevinVandy beat me to it. I like the idea but am not a big fan of typing the prototype as `CoreRow<TData>` which is not strictly accurate.

@mleibman-db did you try making the createRow function into a constructor function, adding the methods directly to the prototype? I haven't tried it myself but intuitively it feels like it should work. Would need to always call createRow with the `new` keyword I think.

packages/table-core/src/features/ColumnFiltering.ts
tombuntus: @KevinVandy beat me to it. I like the idea but am not a big fan of typing the prototype as CoreRow<TData> which is not strictly accurate, (and requires us to create these dummy values to keep typescript happy).

@mleibman-db did you try making the createRow function into a constructor function, adding the methods directly to the prototype? I haven't tried it myself but intuitively it feels like it should work. Would need to always call createRow with the new keyword I think.

packages/table-core/src/core/row.ts
mleibman-db: With the exception of `subRows`, they are all unused (never referenced on the proto) and are only here for TypeScript type checking.

The advice you heard is most likely referring to potential confusion if one ends up modifying a property on the proto instead of an instance, which results in the prop changing in all instances. This is not happening here for `subRows` since it is never modified directly and is only reassigned, which sets the value on the instance.

packages/table-core/src/features/ColumnFiltering.ts
mleibman-db: 1. Typing the row proto as `CoreRow` is actually very useful since it provides type safety and makes sure the methods only access defined props there. The use of default unused values there doesn't strike me as concerning, but we could try to replace them with some purely TypeScript type annotations, though IMHO that would be more hacky.

2. I'm not sure I understand what you're proposing. Could you elaborate?

packages/table-core/src/features/ColumnFiltering.ts
```diff
@@ -362,14 +362,6 @@ export const ColumnFiltering: TableFeature = {
     }
   },
 
-  createRow: <TData extends RowData>(
```
tombuntus, line 365: In the core `createRow` function, we still call these `feature.createRow` functions if they exist, passing them the row and table instance. That should prevent breaking changes for existing custom features, but we may want to recommend custom features to take the same approach (i.e. extend the prototype). @KevinVandy what do you think about this?

I haven't thought all the details through but something like retaining a `createRow` function in each feature, and in the core `createRow` function both calling the `feature.createRow` function with the row and table instances (to prevent breaking changes for existing custom features), and also merging its prototype onto the core `createRow` prototype.

That way we could also retain the `createRow` functions in the core features, (just move the methods onto the prototype), and wouldn't need the `getRowProto` and `Object.assign()` approach I think.

packages/table-core/src/features/ColumnFiltering.ts
tombuntus: > since it provides type safety

It's the wrong type though, isn't it? The prototype shouldn't have the instance properties on it.

> Could you elaborate?

I am imagining something approximately like the below. I haven't tried but _think_ it should work, happy to be corrected. The naming would be a bit weird though. `createRow` should probably become just `Row`, but that would be a breaking change - not sure what to do about that.

```typescript
const createRow = <TData>(
  this: CoreRow<TData>,
  table: Table<TData>,
  id: string,
  original: TData,
  rowIndex: number,
  depth: number,
  subRows?: Row<TData>[],
  parentId?: string
) => {
  this.id = id
  this.original = original
  // etc.
}

createRow.prototype.getValue = (columnId: string) => {
      
    // ...

    return this._valuesCache[columnId] as any
}
```

elsewhere:
```typescript
const row = new createRow(...)
```

packages/table-core/src/features/ColumnFiltering.ts
mleibman-db: +1 to generally recommending people use the same approach for implementing custom features. I considered making things more explicit by adding methods like `initRowProto()` to `TableFeature` interface, but decided against it for simplicity's sake, plus this is more of an internal implementation detail than a public API.

packages/table-core/src/features/ColumnFiltering.ts
tombuntus: > there are no known performance issues around Object.assign()

I don't remember what Kevin is referring to specifically, but my instinct says performance issues would've likely been caused by garbage collection rather than `Object.assign()` itself. Since it's almost always called with a freshly created object in the second argument. Calling it that way many times consecutively can leave a pile of objects lying around for garbage collection.

mleibman-db: > It's the wrong type though, isn't it? The prototype shouldn't have the instance properties on it.

No, it should be `CoreRow`...  Instance props are not present on the proto object itself, but TypeScript needs to know about them in order to do proper type checking.

For your example (second part), this approach results in having to re-assign `createRow.prototype.getValue`

mleibman-db: > > there are no known performance issues around Object.assign()
> 
> I don't remember what Kevin is referring to specifically, but my instinct says performance issues would've likely been caused by garbage collection rather than `Object.assign()` itself. Since it's almost always called with a freshly created object in the second argument. Calling it that way many times consecutively can leave a pile of objects lying around for garbage collection.

I'm sure it's pretty well optimized by all browser engines since it's so common, and I haven't seen any evidence to support it. But happy to change if you feel strongly about it - this is your guys' codebase :)

tombuntus: > No, it should be `CoreRow`

I still think technically this is not the correct type for the prototype. I can understand the pragmatism of using it that way, but it still seems technically incorrect. With the suggestion I made above, I _think_ that the methods you add to the `createRow.prototype` would be able to access the instance properties without complaint from TypeScript, either because you specified the type of `this` in the constructor function args, or because you assigned the properties in the constructor function body. I am not 100% sure though.

> this approach results in having to re-assign createRow.prototype.getValue et al for each row in a loop

Really? I am suggesting to attach them to the prototype of the constructor function itself (once only, at the module level), and then when you call `new createRow(...)` the object returned by that constructor function has those methods on its prototype, because that's the way constructor functions work. Or at least, I _think_ so! As far as my knowledge goes, they are not reassigned anywhere in that process.

mleibman-db: > ...I am suggesting to attach them to the prototype of the constructor function itself (once only, at the module level)...

Don't forget that row protos are created dynamically for each table based on provided table features, so you can't just create the proto once at the module level. I suppose we could make it work, but I don't fully understand how, and things are likely to get very hairy as far as type safety is concerned. TypeScript is not very friendly to constructor functions in general, vastly preferring classes, and making them work requires declaring a separate interface for the constructor function. Considering that we create different row prototypes per table config w/ different features and apply different mixins dynamically, this is bound to get overcomplicated. TypeScript does support the mixin pattern (https://www.typescriptlang.org/docs/handbook/mixins.html), but I'm afraid that going down that route is a typing rabbit hole.

KevinVandy: Anywhere where we are thinking that an alternative would be cleaner, but it's a breaking change, can be reserved for a v9 pr. So far this PR looks mostly good. We don't have to assign dummy vars to the prototype just to satisfy TypeScript. A cast could be acceptable there.

If the Object.assign only gets called once, that is negligible and something we don't need to worry about. Direct assignment was a performance improvement in [this pr](https://github.com/TanStack/table/pull/5048) that sped up rendering when creating 10k+ rows. This PR is solving the memory side of that same issue. In conclusion, I'm not worried about this after you explained more.

packages/table-core/src/features/ColumnFiltering.ts
KevinVandy: This kind of pattern will be useful to think about in the alpha branch though

packages/table-core/src/features/ColumnFiltering.ts
tombuntus: > so you can't just create the proto once at the module level

But I think you can merge the `feature.createRow` prototypes into the prototype of the object returned by the core `createRow` function at runtime, when `new createRow()` is called. In the same loop where we currently call `feature.createRow` in the core `createRow()` function body. I haven't tested this though. In this case the prototypes methods would be created at module level in each of the features' `createRow` functions.

> vastly preferring classes

Personally I am not opposed to using a class if it makes typing easier.

tombuntus: (and just for anyone reading this ... the code snippet in [this](https://github.com/TanStack/table/pull/5927#discussion_r1966274178) comment should be using `function createRow() {}`, not an arrow function!)

tombuntus: > But happy to change if you feel strongly about it

I was actually agreeing with you that since it's not called many times, it wouldn't be likely to cause issues. I was just trying to explain the likely cause of perf issues - not due to `Object.assign()` itself, but rather the fact that it it often called like this:

```typescript
Object.assign(
  targetObject, // <-- existing object
  {
    // new source object which will be garbage collected eventually
  },
)
```

If it's used this way in a loop with many thousands of iterations, you can run into perf issues due to garbage collection.

mleibman-db: > We don't have to assign dummy vars to the prototype just to satisfy TypeScript. A cast could be acceptable there.

Done.

mleibman-db:
Friendly ping

mleibman-db:
That's ok. We can upgrade to the next v8 release with this fix.

riccardoperra:
dshaduishdishudis hduish duish sui hs uish dsui hsiu d

mleibman-db:
> dshaduishdishudis hduish duish sui hs uish dsui hsiu d

@riccardoperra eh?

riccardoperra:
I can definitely look at the implementation better tomorrow. 

It's just an opinion taking a quick look, could an implementation with classes leveraging mixins might be more readable? Achieving the same result. (I don't remember if classes turned out to be more performant for those cases)

mleibman-db:
> It's just an opinion taking a quick look, could an implementation with classes leveraging mixins might be more readable? Achieving the same result. (I don't remember if classes turned out to be more performant for those cases)

I started using that approach initially, but got quickly discouraged with TypeScript semantics, general verbosity, and a bunch of other issues I can't recall right now. Plus having multiple different versions of a `Row` object with different captured values and different implementations (one per table) just hurt my brain, and a more direct/manual approach ended up much more concise and readable to IMHO.

mleibman-db:
I found a potential risky area. Since rows now have prototypes, we can no longer clone them using `{...row}` or any other method that doesn't copy the prototype. I added a `clone()` method to the row and fixed the only two places I could find where this was happening, but user code outside of the component may be doing the same. Alas, there is no other way of achieving the improvement we're looking for without changing the public API, so this will have to do. Definitely worth mentioning in the release notes though.

packages/table-core/src/utils/getSortedRowModel.ts
```diff
@@ -45,7 +45,7 @@ export function getSortedRowModel<TData extends RowData>(): (
         const sortData = (rows: Row<TData>[]) => {
           // This will also perform a stable sorting using the row index
           // if needed.
-          const sortedData = rows.map(row => ({ ...row }))
+          const sortedData = rows.map(row => row.clone())
```
KevinVandy, line 48: If this is the only potential breaking change, I think we may still be able to ship it. It's on the edge of what's acceptable. We don't have much of a way to knowing how many people would rely on this.

KevinVandy:
If anyone wants to install and test this PR in their own codebase, you can install the preview package down below. Find all of the other adapter preview versions under "Continuous Releases" under the "All checks have passed" dropdown.

`@tanstack/react-table@9cf94c1` 

@mleibman-db Not to beat a dead horse by over-explaining this, but you can install this instead of the actual tanstack table package in your own code-base right now so that your team won't be blocked regardless of whether or not we are able to verify that there are no breaking changes or regressions here for a wide release.

mleibman-db:
@KevinVandy We tried that yesterday, and that's how we found the row cloning issue. With it fixed, all unit tests pass now, but I'm not sure about the integration tests. Checking.

mleibman-db:
@KevinVandy So all our tests are passing, with the exception of several instances where we essentially run into a similar issue I listed as a risk above. 

```typescript
const { id, getVisibleCells } = row;
const cells = getVisibleCells();  // ERROR: can only call this as a method on a row
```

mleibman-db:
> You're saying Destructiring doesn't work with prototype methods?

No, destructuring works just fine, but you can't invoke an object's method separately since it wouldn't have access to scope.
For example:

```typescript
const mySet = new Set();

// this works
mySet.has('entry');

// this doesn't
const setHas = mySet.has;
setHas('entry');

// or this
const { setHas } = mySet;
setHas('entry');
```

KevinVandy:
That makes sense. But to be clear, you are saying that anyone who is currently destructuring methods from rows and then tries to call them is going to get unexpected results/errors? That would be a pretty large breaking change. Totally acceptable for v9, but that brings a lot of doubt to me for whether or not we would be able to safely release this for v8.

mleibman-db:
> That makes sense. But to be clear, you are saying that anyone who is currently destructuring methods from rows and then tries to call them is going to get unexpected results/errors?

Yes.

> That would be a pretty large breaking change. Totally acceptable for v9, but that brings a lot of doubt to me for whether or not we would be able to safely release this for v8.

The only way around it I can see for v8 is to make it optional/configurable and keep the default behavior backwards compatible.

riccardoperra:
@mleibman-db @KevinVandy may proxies can help in this case? Something like this should allow to call methods with destructuring without losing `this` scope. I haven't investigated performance and memory usage in depth though.

```ts
const rowProxyHandler: ProxyHandler<Row<any>> = {
  get(target, p, receiver) {
    const res = Reflect.get(target, p, receiver);
    return typeof res === 'function' ? res.bind(receiver) : res;
  }
}

export const createRow = <TData extends RowData>(
  table: Table<TData>,
  id: string,
  original: TData,
  rowIndex: number,
  depth: number,
  subRows?: Row<TData>[],
  parentId?: string
): Row<TData> => {
  const row: CoreRow<TData> = Object.create(getRowProto(table))
  // [...]

  const rowProxy = new Proxy(row, rowProxyHandler);

  for (let i = 0; i < table._features.length; i++) {
    const feature = table._features[i]
    feature?.createRow?.(rowProxy as Row<TData>, table)
  }

  return rowProxy as Row<TData>;
}
```

mleibman-db:
> @mleibman-db @KevinVandy may proxies can help in this case? Something like this should allow to call methods with destructuring without losing `this` scope. I haven't investigated performance and memory usage in depth though.
> 
> ```ts
> const rowProxyHandler: ProxyHandler<Row<any>> = {
>   get(target, p, receiver) {
>     const res = Reflect.get(target, p, receiver);
>     return typeof res === 'function' ? res.bind(receiver) : res;
>   }
> }
> ```

@riccardoperra 

In terms of performance overhead, in my tests with the `/examples/react/grouping`, it slows `getGroupedRowModel()` down about 2x compared to the implementation w/o the proxy, though it's still 2x faster if compared to the baseline.

It works, but it only addresses one of the two potential problematic use cases:

```typescript
const { id, getVisibleCells } = row;
const cells = getVisibleCells();  // pattern 1: this works with the proxy
const clone = {... row}; // pattern 2: this still doesn't
```

The second one could of course be handled by the `ownKeys()` proxy handler, but we're getting into pretty muddy territory here. I'm a huge fan of proxies myself, but this feels wrong as we'd be changing the natural expected behavior of JavaScript objects with this (e.g. `obj.fn === obj.fn` is no longer true unless we start caching bound functions, which complicates and slows things down even more, etc).

At this point, I'd be strongly in favor of just listing it as a breaking change in release notes. IMHO, both of the patterns in the example above are intuitively "wrong" in a sense that using them implies making assumptions about `row` being a POJO (plain old javascript object, i.e. w/o a prototype, or with the prototype being `Object.prototype`), and that is not an assumption one should make. The fact that these call patterns have worked so far is a consequence of internal implementation details and in no way constitutes an API contract with the user.

Zertz:
If I may, is there any chance this change could be v9 and the current alpha become v10?

KevinVandy:
> If I may, is there any chance this change could be v9 and the current alpha become v10?

That's actually becoming my plan. We're going to split up the future deliverables a bit. V9 might be focusing on a few small breaking changes around conventions like this, and v10 can focus on tree shaking and signal state

jarikoff:
Waiting for this one 🙏

KevinVandy:
> Waiting for this one 🙏

It will only be in a future major version. Going to make a decision on how/where to incorporate this soon.

GrinZero:
Great, looking forward to merging soon

artchen:
Really looking forward to get this fix in our app. Any updates?

artchen:
Hi @KevinVandy, could you shed some light on when will this improvement be incorporated?

rarenal:
@KevinVandy @mleibman-db is there any way we can help moving it forward? This would be a massive improvement for us.

KevinVandy:
Realized that I never gave an update on this PR.

I implemented a very similar pattern in the v9 alpha. All rows, columns, headers, and cells objects store functions on the prototype. There is a significant memory reduction.

KevinVandy:
Closing this PR to V8 as it is in V9, as the story is told here: https://tanstack.com/blog/tanstack-table-v9-memory-performance

PR #5964, opened 2025-03-13
chore: Remove unnecessary `console.error` log from `table.getColumn(id)` implementation

mikekobit-stripe:
The `getColumn` method is already typed to return `Column<TData, unknown> | undefined`,  indicating to callers that the result might be undefined. The current `console.error` logging when a column is not found:

1. Adds noise in development environments
2. Can mask actual issues
3. Forces users to implement workarounds (as discussed in https://github.com/TanStack/table/discussions/5505)

This change removes the error logging while maintaining the expected return type, allowing consumers to handle undefined columns as already expected by the API.

## Summary by CodeRabbit

* **Refactor**
  * Simplified internal column retrieval implementation, removing redundant development-mode error logging for missing columns.

<sub>✏️ Tip: You can customize this high-level summary in your review settings.</sub>


deepan-ashby (approved):
Just ran into this myself, this seems like a reasonable fix

DominikBoniecki-haptiq (approved):

mattprice09 (approved):

Ombrenoirz:
I'm following this pr carefully 😅

mikekobit-stripe:
It looks like there was some internal refactoring from the time this pull request was opened til now. It doesn't appear at first glance that the public facing API has changed, but the underlying log is still present in a slightly varied form.

https://github.com/TanStack/table/blob/b9c46a784d851ec40a5ab34cc868448528e8d3dc/packages/table-core/src/core/columns/coreColumnsFeature.utils.ts#L263-L288

So it looks like the logs will still be flooded with `warn` instead of `error`, so this will still be a problem. Please consider making the same change to the new version.

KevinVandy:
You shouldn't have code that can try to call invalid column ids. We even used to throw an error here

PR #5984, opened 2025-04-08
fix(solid-table): converge on use for hooks

birkskyum:
Adopt `useSolidTable` syntax for solid, so it's like vue/react/svelte5. This is aligned with how we do for tanstack [router](https://tanstack.com/router/latest/docs/framework/solid/guide/path-params#path-params-in-components) (useParams, useSearch, useRouter ..) and [query](https://tanstack.com/query/latest/docs/framework/solid/reference/useQuery) (useQuery, useMutation ..), [store](https://tanstack.com/store/latest/docs/framework/solid/reference/functions/usestore) (useStore) and the "official" [solid router](https://docs.solidjs.com/solid-router/reference/primitives/use-match) (useNavigate, useMatch ..), so it's nice to have syntax convergence for hooks.

### DIFF
--- a/docs/api/core/table.md
+++ b/docs/api/core/table.md
@@ -2,7 +2,7 @@
 title: Table APIs
 ---
 
-## `createAngularTable` / `useReactTable` / `createSolidTable` / `useQwikTable` / `useVueTable` / `createSvelteTable`
+## `createAngularTable` / `useReactTable` / `useSolidTable` / `useQwikTable` / `useVueTable` / `createSvelteTable`
 
 ```tsx
 type useReactTable = <TData extends AnyData>(
--- a/docs/framework/solid/guide/table-state.md
+++ b/docs/framework/solid/guide/table-state.md
@@ -11,7 +11,7 @@ TanStack Table has a simple underlying internal state management system to store
 You do not need to set up anything special in order for the table state to work. If you pass nothing into either `state`, `initialState`, or any of the `on[State]Change` table options, the table will manage its own state internally. You can access any part of this internal state by using the `table.getState()` table instance API.
 
 ```jsx
-const table = createSolidTable({
+const table = useSolidTable({
   columns,
   get data() {
     return data()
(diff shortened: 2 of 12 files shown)


riccardoperra:
Could be ok for @deprecate, but I'm not sure really about this 🤔. Following a [comment](https://github.com/solidjs/solid/issues/317#issuecomment-767715235) by Ryan a while ago , `create` seems more appropriate in this case.

For example, in @tanstack/store, you're creating a new store via `new Store`, then you make use of it via `useStore`. I think the same can be compared with the other libraries. Also [solid-primitives](https://github.com/solidjs-community/solid-primitives/blob/main/CONTRIBUTING.md#make-non-reactive-vs-create-reactive) follows a similiar approach

In table, `createSolidTable` will construct the table and create a state there, it's not using something defined on the top. 

In v9 (alpha branch) it has been renamed to `createTable`

KevinVandy:
I'm in favor of keeping it as create as per Ryan's comments. As for the TanStack docs switching between frameworks causing 404s, I think we can just set up better redirects

birkskyum:
We'll need the redirects for sure - angular e.g. often has an injectX naming pattern.

I understand the sentiment in Ryan's comment, I just don't agree with it justifying an extra migration step - especially because the exact same logic apply to vue + svelte, which both decided to stick to `use` anyway.

ryansolid:
I see where this came from now. It's interesting. We do use `use___` specifically for things that use context. More recently I've been debating having the `use` signify things that have to live inside of the render tree/reactive graph. Ie.. `useContext`, `useEffect`, `useUniqueId` vs `createSignal`, `createStore`, which can be used outside... However the feedback was that was too subtle. That being said it would be very odd at this point I think to be like `useSignal`. 

`"use server"` is a bit different. It was less the phrasing and the use of those sort of statements we were following suit on.. There is a `"use strict"` after all.

I guess the question is whether to just say all custom primitives use `use`. Svelte and Vue don't use `use` internally to my knowledge so this convention is something coming from third parties. Solid Primitives followed core conventions whereas the ecosystem lead Svelte and Vue who switched to these sort of primitives later on after being established. I don't know this seems like a bigger question than can be answered here.

Of course this is a 3rd party library so you can do whatever you want.

PR #5996, opened 2025-04-24
Prevents deletion of non-selectable rows when calling `onRowSelectionChange`

simon-curtis:
This simple fix prevents deleting non-selectable rows from the RowSelectionState if the row is not selectable. 

In my case I need to show to the user that an item has been added to the list previously and cannot be un-checked via this menu, leaving it in a checked and disabled state.

```tsx
{
    id: "select",
    header: ({ table }) => (
        <Checkbox
            checked={
                table.getIsAllRowsSelected()
                    ? true
                    : table.getIsSomeRowsSelected()
                        ? "indeterminate"
                        : false
            }
            onCheckedChange={(value) => table.toggleAllRowsSelected(!!value)}
            aria-label="Select all"
        />
    ),
    cell: ({ row }) => (
        <Checkbox
            checked={row.getIsSelected()}
            disabled={!row.getCanSelect()}
            onCheckedChange={(value) => row.toggleSelected(!!value)}
            aria-label="Select row"
        />
    ),
    enableSorting: false,
    enableHiding: false,
    meta: { className: "w-0" },
},
```

Unchecking the header checkbox currently removes the checked state, but it should not.


KevinVandy:
closing in favor of #6495

PR #6005, opened 2025-05-06
fix: export GroupColumnDefBase

### LINKED ISSUE #6004: GroupColumnDefBase not being exported, causing type error when using `declaration` tsconfig #5444
### TanStack Table version

v8.21.3

### Framework/Library version

React v18

This is a very similar issue to #5445 
Due to this type not being exported, creating declaration files for objects that uses `columnHelper.group` throws a TypeScript error:

`Exported variable 'foo' has or is using name 'GroupColumnDefBase' from external module "/.../@tanstack/table-core/build/lib/types" but cannot be named.` 

Exporting this variable resolves this issue

*

_No response_

None

### Terms & Code of Conduct

thescripted:
Fixes #6004


InteractiveTimmy (commented):
**Note** - This would improve type safety and alignment in downstream consumers.

PR #6058, opened 2025-07-18
Update React peer dependency to support React 19

yash-walke-solytics:
This PR updates the `react-table` package to officially support React 19 by modifying the peer dependency range in `package.json`:

- Updated `react` peer dependency to include: `^19.0.0`
- Bumped package version from `7.8.0` to `7.9.0`

## Reason

React 19 introduces several updates that improve rendering and developer experience. This change ensures `react-table` remains compatible with projects upgrading to React 19.

## Checklist

- [x] Peer dependency updated
- [x] Version bumped

### DIFF
--- a/package.json
+++ b/package.json
@@ -1,6 +1,6 @@
 {
   "name": "react-table",
-  "version": "7.8.0",
+  "version": "7.9.0",
   "description": "Hooks for building lightweight, fast and extendable datagrids for React",
   "license": "MIT",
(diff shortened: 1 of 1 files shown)


yash-walke-solytics:
Hi @tannerlinsley,
I've updated the React peer dependency to include support for React 19 and bumped the package version accordingly. There are no conflicts with the base branch, and the changes are minimal and scoped.

Could you please review and confirm if we can proceed with merging this?

Thanks!

PR #6060, opened 2025-07-23
fix(sort): treat null values the same as undefined in SortUndefined

marceloverdijk:


marceloverdijk:
@KevinVandy related to:
- https://github.com/TanStack/table/pull/5486
- https://github.com/TanStack/table/pull/5485

maybe there was a good reason to exclude `null`'s?

PR #6080, opened 2025-08-20
Fix table.getIsSomeRowsSelected when filtered out rows are selected

### LINKED ISSUE #6079: Table `getIsSomeRowsSelected()` does not consistently return correct value when column filter is applied
### TanStack Table version

v8.21.3

### Framework/Library version

React v18.3.1

When using TanStack Table with row selection and filtering, the table `getIsSomeRowsSelected()` function does not consistently return true when some visible rows are selected.

Steps to reproduce:

1. Create a table with row selection and filtering
2. Select several rows.
3. Apply a filter so only a subset of the selected rows are visible.
4. Click the header checkbox to select all of the filtered rows.
5. Begin unselecting the filtered rows from the top and observe that the header checkbox becomes unselected instead of indeterminate, indicating that `getIsSomeRowsSelected()` is returning false when it should be returning true.

This appears to be caused by the way the `table.getIsSomeRowsSelected()` is implemented. This is the current function implementation:

```
table.getIsSomeRowsSelected = () => {
  const totalSelected = Object.keys(
    table.getState().rowSelection ?? {}
  ).length
  return (
    totalSelected > 0 &&
    totalSelected < table.getFilteredRowModel().flatRows.length
  )
}
```

The issue with this approach is that it checks if the total number of selected rows (across the whole table, not just the filtered rows) is greater than 0 and less than the number of filtered rows. If you select rows that are filtered out, they will still count towards `totalSelected`. So if you have 3 visible rows and 2 selected (and also 2 more selected that are filtered out), `totalSelected` would be 4, which is greater than the number of visible rows (3), thus causing this function to return `false`, even though some filtered rows are selected.

https://tanstack.com/table/latest/docs/framework/react/examples/row-selection

https://github.com/user-attachments/assets/8f89dac9-32eb-4d67-a8a8-4089294dbcae

Yes, I am also opening a PR that solves the problem along side this issue

### Terms & Code of Conduct

kahopfer:
This pull request is for fixing an issue where table.getIsSomeRowsSelected occasionally returns an incorrect value when filtered out rows are selected.

Fixes https://github.com/TanStack/table/issues/6079

## Summary by CodeRabbit

* **Bug Fixes**
  * Row selection status now reflects only visible (filtered) rows.
  * The header checkbox shows indeterminate state only when some visible rows are selected and not all.
  * “Select all” and partial selection behaviors align with the current filter view.

* **Chores**
  * No changes to public APIs; behavior improved without altering external interfaces.


kahopfer:
Closing this PR since https://github.com/TanStack/table/issues/6079 was closed. It sounds like this will be fixed in v9.

PR #6098, opened 2025-09-20
fix(table-core): CoreOptions columns and data are readonly

Sheraff:
When used as the type for a function parameter, `readonly` ensures that the parameter will not be mutated inside the function.

The `columns` and `data` array inputs are never modified by `table-core`. They should thus be marked as `readonly` to make this contract explicit. This allows developper to rely on the guarantee they can pass in static objects.

Here's an example where this is useful:

```ts
const Component = () => {
  const { data } = useQuery({ ... })
  const table = useReactTable({ data, ... })
```

In a case like this, `data` comes from the `tanstack/query` cache, and as such must be immutable. If `useReactTable` can guarantee that it won't modify the array we pass in, then we can pass `data` directly.

## Summary by CodeRabbit

- Refactor
  - Table options now use readonly arrays. Both data and columns accept readonly arrays, and related getters return readonly arrays. This enforces immutability and improves type safety. If your code mutates these arrays or relies on mutable types, update it to create new arrays instead.

- Documentation
  - API docs updated to reflect readonly types for data and columns in the Table APIs.


rsslldnphy:
Very excited to see this PR! Just run into the problem that it fixes. Is it likely to be merged any time soon do you know?

Sheraff:
> Very excited to see this PR! Just run into the problem that it fixes. Is it likely to be merged any time soon do you know?

Probably no changes are going to be made on this major sadly

rsslldnphy:
oh really? that's a shame :-( - how come? it's a completely backwards compatible change isn't it?

Issue #6115, opened 2025-10-20
Expanding API including rows that dont include a subRow


RichardHpa:
### TanStack Table version

8.21.3

### Framework/Library version

17.0.1

### Describe the bug and the steps to reproduce it

While using the expanding api within tanstack table, I noticed that the expanding doesn't take into consideration rows that **don't** have subrows.

This is only obvious when you open all rows (in my example I have set the initial expanded state to true) and then close just 1 row. If you have a look at the expanded state, it now includes every row's index, including the ones that don't have subrows. 

To show this I have used the default expandable example and made a few edits but the main change is I have made it that only even numbered rows have subrows.

Steps to reproduce
1. View this [sandbox](https://stackblitz.com/edit/tanstack-table-3zzzyjrn?file=src%2Fmain.tsx). Everything should be open and the expanded state is `true`
2. Close the 2nd row. This should now generate the expanded state. And you should see that both **0** and **3** are now `true`, which shouldn't even better in the object as those aren't rows that can expand.

This also happens for deep nested rows which means your `.getExpandedDepth()` is actually incorrect as its giving you a max depth that isn't actually possible.

Steps to reproduce this other issue
1. View this [sandbox](https://stackblitz.com/edit/tanstack-table-3zzzyjrn?file=src%2Fmain.tsx). Everything should be open and the expanded state is `true`
2. Close every expandable row individually. Even though all the rows should be uncollapsed now. The expanded state still includes entries but all of them are invalid entires as their rows dont have sub rows.

I would expect those keys to not be in the object as they dont have a sub row

### Your Minimal, Reproducible Example - (Sandbox Highly Recommended)

https://stackblitz.com/edit/tanstack-table-3zzzyjrn?file=src%2Fmain.tsx

### Screenshots or Videos (Optional)

<img width="360" height="1144" alt="Image" src="https://github.com/user-attachments/assets/9d219a41-ba63-44dc-8577-940efb9af41a" />

<img width="321" height="685" alt="Image" src="https://github.com/user-attachments/assets/1f3b35c5-9738-487c-8567-2d0e2264beeb" />

### Do you intend to try to help solve this bug with your own PR?

None

### Terms & Code of Conduct

- [x] I agree to follow this project's Code of Conduct
- [x] I understand that if my bug cannot be reliable reproduced in a debuggable environment, it will probably not be fixed and this issue may even be closed.

RichardHpa:
[...]

PR #6116, opened 2025-10-21
[BUG] ignore non expandable rows when the expanded state changes

RichardHpa:
## 🎯 Changes

Resolving the issue I bought up in https://github.com/TanStack/table/issues/6115 where if the old expanded state is `true` and you change any of the expanded rows. It will also include rows that don't have expanded rows. This was then causing a situation where the `getExpandedDepth` function was returning the wrong number as the state included rows that shouldn't have been included

I have just added a check to see if that row should be expandable before it gets added to the expandable state.

## ✅ Checklist

## 🚀 Release Impact

## Summary by CodeRabbit

* **Bug Fixes**
  * Fixes the "expand all" behavior so only rows eligible for expansion are marked expanded; non-expandable rows are no longer added to the expanded state. Preserves single-row expansion behavior, external change handlers, and prevents unintended visual/state changes when toggling expand-all.

<sub>✏️ Tip: You can customize this high-level summary in your review settings.</sub>


RichardHpa:
@coderabbitai help

RichardHpa:
@coderabbitai full review

PR #6144, opened 2026-01-08
feat(angular): Add angular table composability

riccardoperra:
Here's the notable changes:

TLDR

- Introduced a new `createTableHook` helper to provide a strongly typed, Angular-friendly table API
- `injectAppTable` now returns an enhanced table instance with:
  - typed `tableComponents`, `cellComponents`, `headerComponents`
  - helper functions (`appCell`, `appHeader`, `appFooter`) for typed templates
- Column definitions can directly reference components defined in `createTableHook` (e.g. `cell.TextCell`)
- Added `tanStackTable`, `tanStackTableCell`, `tanStackTableHeader`, `tanStackTableFooter` directives to enable context injection in rendered components
- `FlexRenderDirective` was refactored to rely entirely on signals. No manual dirty checking anymore!
- Simplified `*flexRenderCell`, `*flexRenderHeader`, and `*flexRenderFooter` APIs
- Added injectable tokens for cell / header / footer contexts when using `flexRender`

## FlexRenderCell simplified directive

You can now pass directly a `cell`, `header` or `footer` instance while using \*flexRenderCell, \*flexRenderHeader or \*flexRenderFooter. `*flexRender`  directive still works and can be used when you need to specifiy different props.

```html
<!-- Old -->
<td *flexRender="cell.column.columnDef.cell; props: cell.getContext(); let cell">{{cell}}</td>
<td *flexRender="header.column.columnDef.header; props: header.getContext(); let header">{{header}}</td>
<td *flexRender="footer.column.columnDef.footer; props: footer.getContext(); let footer">{{footer}}</td> 

<!-- New -->
<td *flexRenderCell="cell; let cell">{{cell}}</td>
<th *flexRenderHeader="header; let header">{{header}}</th>
<th *flexRenderFooter="footer; let footer">{{footer}}</th>
```

## New `createTableHook` helper

Angular doesn’t support higher-order components in the same way React does, but we can still provide a strong, typed authoring experience by leveraging a few Angular-friendly patterns.

```typescript
export const {
  createAppColumnHelper,
  injectAppTable,
  injectTableContext,
  injectTableCellContext,
  injectTableHeaderContext,
  // injectFlexRenderHeaderContext
  // injectFlexRenderCellContext
} = createTableHook({
  tableComponents: {
    TableToolbar // those will be available through `table` (e.g. table.TableToolbar)
  },
  cellComponents: {
    TextCell // those will be available in columns cell definition (e.g. cell.TextCell)
  },
  headerComponents: {
    FooterSum: // those will be available header/footer definition (e.g. header.FooterSum)
  }
})
```

In React, context can be injected via components such as `AppCell` or `AppHeader`, which act as typed wrappers around render logic:

```tsx
<table.AppCell cell={c} key={c.id}>
  {(cell) => (
    <td>
      {/* Cell components are pre-bound via AppCell */}
      <cell.FlexRender />
    </td>
  )}
</table.AppCell>
```

In Angular, the only way to provide contextual bindings to templates is through structural directives. However:

- Structural directives cannot be created dynamically from a factory function
- Defining them statically would prevent us from reusing the “app table” typings produced by `createTableHook`

Because of this, directly emulating the React pattern is not feasible in Angular.

Instead of introducing new directives, we augment the **table instance itself** with typed helper functions.

This works because:

- Component bindings for cells, headers, and footers are already attached during table initialization
- We rely on a custom TanStack Table feature that allows this augmentation at setup time
- No runtime `Object.assign` is needed during rendering

This allow us to just call `*flexRender` and pass that cell in the context, like we did in \<V9.

As a recap:

- Cell object will contain all spreaded `cellComponents`
- Header and Footers objects will contain spreaded `headerComponents`
- Table object will contain spreaded `tableComponents`

Column definitions work the same way as before. If a component class or the result of `flexRenderComponent` is provided, the system renders that component with the correct table context. As a plus, thanks to the new types, you are able to just pass the components defined in `createTableHooks`

```typescript
personColumnHelper.accessor('firstName', {
  header: 'First Name',
  footer: ({ header }) => flexRenderComponent(header.FooterColumnId),
  cell: ({ cell }) => flexRenderComponent(cell.TextCell),
})
```

#### Using cell/header components in template

What if you need to use `header.FooterColumnId` in the template? In the react example we have something like this:

```tsx
<table.AppFooter header={f} key={f.id}>
                      {(footer) => {
                        const columnId = footer.column.id

                        return (
                          <td colSpan={footer.colSpan}>
                            {footer.isPlaceholder ? null : (
                              <>
                                {/* Use FooterSum for numeric columns, FooterColumnId for others */}
                                {columnId === 'age' ||
                                columnId === 'visits' ||
                                columnId === 'progress' ? (
                                    <footer.FooterSum />
                                ) : columnId === 'actions' ? null : (
                                    <footer.FooterColumnId />
                                )}
                              </>
                            )}
                          </td>
                        )
                      }}
                    </table.AppFooter>
```

In Angular, we achieve the same result by using typed helpers exposed on the table instance.

These helpers (`appCell`, `appHeader`, `appFooter`) simply return **casted versions** of the original TanStack objects, enriched with the app-specific typings. They are cheap at runtime and do not introduce additional directives or wrappers.

```html
@for (_footer of footerGroup.headers; track _footer.id) {
  @let footer = table.appFooter(_footer); --> this will return an AppHeader type

  <!-- using flexRender -->
  <ng-container *flexRender="footer.FooterSum; props: footer.getContext()"/>;

  <!-- using angular built-in directive -->
  
  <ng-container [tanStackTableFooter]="footer"> <!-- injectable directive that provides footer -->
    <ng-container *ngComponentOutlet="footer.FooterSum"/>
    <!-- if you want to call it like before, 
         you may want to access that context without passing inputs 
     -->
    <app-footer-sum/>
  </ng-container>
}
```

## Context injection via `tanStackTableCell / Header / Footer` directives

In addition to the typed helpers, this PR introduces the following injectable directives:

- `tanStackTableCell`
- `tanStackTableHeader`
- `tanStackTableFooter`

These directives exist for **one specific reason**: to allow components rendered inside the table (via `ngComponentOutlet` or just using the component tag) to inject the correct TanStack Table context. If you use `*flexRender`, this is already done internally via that directive. This will act like a context in React.

```html
<ng-container [tanStackTable]="table()">
 
 <!-- components can just call injectTableContext() to access table, instead of propagating inputs -->

 <table>
   <thead>
     <th [tanStackTableHeader]="header()">
       <!-- components can just call injectTableHeaderContext() to access header -->
     </th>
   </thead>

   <!-- cell rendering -->
    <tr>
      <td [tanStackTableCell]="cell()"> 
        <!-- components can just call injectTableCellContext() to access cell -->
      </td>
    </tr>
   
 </table>

</ng-container>
```

## FlexRenderDirective rendering refactor

Since the new `injectTable` with `angularReactivityFeature` will wraps every method of row, cells etc to a cached computed, the \*flexRender directive doesn't need to handle anymore the "mark as dirty" using both `ngDoCheck` and effects.

The refactored implementation removes `ngOnChanges` and `ngDoCheck` using a single `effect` called during class initialization, which is responsible to handle the entire flow:

- On the first render, we will create a component/template etc. based on the given inputs. During this phase we will create a nested effect `effect`with `content(props)` as a dependency
- When that effect changes, we will set the `Dirty`flag in order to update the component on the next `` `FlexRender#update` ``  call
- When input content or props changes, the nested effect will be destroyed and the `render` flow will start from scratch. We don't need here to handle component manual dirty checking like before because it will be automatically marked as dirty when you call a computed method inside.

## FlexRender new injectable tokens

While using flexRender, based on the given context, components and template and rendered with a custom injector whith those tokens as providers:

While passing a cell context:

- `CellContextToken` -> `injectTableCellContext()` -> returns a signal of a cell instance

While passing an header or footer context

- `HeaderContextToken` -> `injectTableHeaderContext()` -> returns a signal of an header or footer instance

### DIFF
--- a/examples/angular/composable-tables/src/app/components/cell-components.ts
+++ b/examples/angular/composable-tables/src/app/components/cell-components.ts
@@ -3,6 +3,7 @@ import { injectFlexRenderContext } from '@tanstack/angular-table'
 import { CurrencyPipe } from '@angular/common'
 import { injectTableCellContext } from '../table'
 import type { CellContext, TableFeatures } from '@tanstack/angular-table'
+import type { Person } from '../makeData'
 
 @Component({
   selector: 'span',
@@ -67,7 +68,7 @@ export class ProgressCell {
   `,
 })
 export class RowActionsCell {
-  readonly cell = injectTableCellContext<number>()
+  readonly cell = injectTableCellContext<number, Person>()
 
   view() {
     alert(


riccardoperra:
@coderabbitai full review

riccardoperra:
@coderabbitai review

riccardoperra:
@coderabbitai review

riccardoperra:
@coderabbitai review

PR #6181, opened 2026-02-25
feat(angular,solid,vue): add core signal reactivity feature 

riccardoperra:
Table will expose a `optionsStore` which is a store storing the table options. This remove the necessity to call `table.baseStore.setOptions(prev => ({...prev}))` on every `setOptions` in our adapters.

Since `options` will be reactive and patched with the `notifier` given by the adapter through the`tableReactivityFeature`, solid/vue table will update even if an option change after table construction (properties given as a getter)

## Summary by CodeRabbit

* **New Features**
  * Runtime toggle for row selection in Angular, Solid, and Vue examples; new Solid devtools package and an "Options" inspector panel.

* **Bug Fixes**
  * Row-selection checkboxes now disable when a row cannot be selected.

* **Refactor**
  * Large rework of reactivity and options handling across frameworks; removal of legacy Angular reactivity utilities.

* **Chores**
  * Multiple dependency bumps and added tooling/config for Solid devtools.


packages/table-core/src/core/table/coreTablesFeature.types.ts
```diff
@@ -94,14 +94,22 @@ export interface Table_CoreProperties<
...
+   */
+  optionsStore: Store<TableOptions<TFeatures, TData>>
   /**
    * This is the resolved initial state of the table.
    */
   initialState: TableState<TFeatures>
+  /**
+   * The latest table options value (non reactive).
+   */
+  latestOptions: TableOptions<TFeatures, TData>
```
KevinVandy: I'm trying to follow along. Why is this needed?

riccardoperra: I'm trying to figure out whether we still need it or not. I initially added it because I was running into recursive and infinite access issues in the Angular adapter.

In theory, from TanStack Store I expected to have an `untracked` utility to read a signal’s value without triggering the dependency auto-tracking system, exactly how you can do in solid, angular or like Preact with the `peek` method

- [https://preactjs.com/guide/v10/signals/#reading-signals-without-subscribing-to-them](https://preactjs.com/guide/v10/signals/#reading-signals-without-subscribing-to-them)
- [https://docs.solidjs.com/reference/reactive-utilities/untrack](https://docs.solidjs.com/reference/reactive-utilities/untrack)
- [https://angular.dev/api/core/untracked](https://angular.dev/api/core/untracked)

riccardoperra: had seen some inconsistencies in my tests expecially while using table\_mergeOptions and table\_setOptions utils since they access the "reactive" version of the options, so I've temporarly introduced a non-reactive property that you can access there, until I check better the issues I had

PR #6187, opened 2026-03-04
fix(table-core): consistent alphanumeric sorting for mixed letter/digit strings

### LINKED ISSUE #6174: Alphanumeric sorting is not accurate
### TanStack Table version

8.21.3

### Framework/Library version

React 19.1.0

The alphanumeric table sorting appears to change how it sorts depending on the location in the word of the character in question.

For the first character in a word, the table does alphanumeric sort, with letters being sorted BEFORE numbers. This is strange, but is already reported in https://github.com/TanStack/table/issues/5500. 

This issue is different, is definitely a bug, and describes how the sort behavior changes mid-word.

The first character is sorted, as described above, alphanumerically with letters before numbers. However, subsequent characters are sorted differently and sometimes innacurately. One example provided below.

These two lists are what the table is currently showing as sorted ascending. This is incorrect.
apple1
apple2
appleA
appleB
banana1
banana2
1apple
1banana
2apple
2banana

This is what that list SHOULD look like, assuming alphanumeric sorting with letters coming before numbers. Note that the positions of the first four entries have changed from above.
appleA
appleB
apple1
apple2
banana1
banana2
1apple
1banana
2apple
2banana

This is what the list should look like if sorted alphanumerically where numbers come before letters. This is the ideal sort order, but since that point is already made in ticket 5500, I won't belabor it. Whether letters or numbers come first, this ticket is about keeping that ordering consistent regardless of where in the word the character is located. This list below is, for example, how a (literal physical book-filled) library would sort the items:
1apple
1banana
2apple
2banana
apple1
apple2
appleA
appleB
banana1
banana2

I don't have access to this

Image

None

### Terms & Code of Conduct

isaackaara:
## Problem

The alphanumeric sorting function produces inconsistent ordering for mixed letter/digit strings depending on where digits appear in the string.

As reported in #6174:
- At the start of a string: letters sort before digits (`Aapple` < `1apple`) ✓
- In the middle of a string: digits sort before letters (`apple1` < `appleA`) ✗

Expected consistent behavior: letters should always sort before digits regardless of position.

## Root Cause

The previous implementation used regex-based chunk splitting (`/([0-9]+)/gm`), which produced misaligned chunks when comparing strings with different digit/letter boundaries:

- `"apple1"` splits to `["apple", "1"]`
- `"appleA"` splits to `["appleA"]` (no digit groups)

This caused the comparison to be between `"apple"` and `"appleA"` (string comparison), rather than correctly comparing character-by-character at the divergence point.

## Fix

Replaced the chunk-based approach with a character-by-character comparison that:

1. **Compares non-digit characters lexicographically** — standard string ordering
2. **Extracts and compares full digit sequences as numbers** — natural numeric ordering (e.g., `2` < `10`)
3. **Consistently sorts letters before digits at any position** — the key fix

### Example sort results (ascending):

| Before (inconsistent) | After (consistent) |
|---|---|
| apple1 | appleA |
| apple2 | appleB |
| appleA | apple1 |
| appleB | apple2 |

The exported `reSplitAlphaNumeric` regex is preserved (marked as deprecated) for backwards compatibility.

Closes #6174

### DIFF
--- a/packages/table-core/src/sortingFns.ts
+++ b/packages/table-core/src/sortingFns.ts
@@ -1,5 +1,8 @@
 import { SortingFn } from './features/RowSorting'
 
+/**
+ * @deprecated No longer used internally. Kept for backwards compatibility.
+ */
 export const reSplitAlphaNumeric = /([0-9]+)/gm
 
 const alphanumeric: SortingFn<any> = (rowA, rowB, columnId) => {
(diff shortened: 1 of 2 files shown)


packages/table-core/src/sortingFns.ts
```diff
@@ -67,51 +70,58 @@ function toString(a: any) {
   return ''
 }
 
+function isDigitChar(ch: string): boolean {
```
Christopher96u, line 73: shouldn't we remove the return type? I think it's redundant

PR #6196, opened 2026-03-06
fix: replace Date.now() with performance.now() in memo debug timing

### LINKED ISSUE #6127: DX: Next.JS complain about using `Date.now()` inside a Client Component
### TanStack Table version

v8.21.3

### Framework/Library version

React 19.2.0 Next 16.0.3

It's a pure DX issue, where Next.JS will show an issue when Date.now() is used in a client component, mainly to tell devs not to use it, where it leds to the render on server and client might differ.

TanStack Table used `Date.now()` [in memo function](https://github.com/TanStack/table/blob/02c203afed5865cfec0aa614de9f2861645d6ff9/packages/table-core/src/utils.ts#L150), note there is no way to switch it off, opts.debug is always defined [shows here](https://github.com/TanStack/table/blob/02c203afed5865cfec0aa614de9f2861645d6ff9/packages/table-core/src/utils.ts#L215).

I suggest either making getMemoOptions return undefined, or changing to use performance.now(), which Next.JS will not complain, note need to take care edge case like #3830, tho engine version is set to >= 12.

Extra Next.JS docs: [here](https://github.com/vercel/next.js/blob/5b62dd8504650ab03a960039b86f63c5dcc22ef5/packages/next/src/server/node-environment-extensions/date.tsx#L1-L11)

If a maintainer decided, I can help open a PR for that

N/A

_No response_

### Terms & Code of Conduct

isaackaara:
Next.js warns about `Date.now()` usage in client components because it can cause hydration mismatches between server and client renders. While the `Date.now()` calls in `memo()` are only used for debug timing and guarded behind `opts.debug`, Next.js statically detects `Date.now()` usage and shows the warning regardless.

Switching to `performance.now()` resolves the Next.js warning and also provides more precise sub-millisecond timing measurements for debug output. The `performance.now()` API is available in all modern browsers and Node.js 12+, which aligns with the project's engine requirements.

This is a debug-only change with no impact on runtime behavior.

Fixes #6127

### DIFF
--- a/packages/table-core/src/utils.ts
+++ b/packages/table-core/src/utils.ts
@@ -147,7 +147,7 @@ export function memo<TDeps extends readonly any[], TDepArgs, TResult>(
 
   return (depArgs) => {
     let depTime: number
-    if (opts.key && opts.debug) depTime = Date.now()
+    if (opts.key && opts.debug) depTime = performance.now()
 
     const newDeps = getDeps(depArgs)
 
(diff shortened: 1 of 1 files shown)


KevinVandy:
can't change this for the node versions that are supported in the version

PR #6202, opened 2026-03-12
fix(react-table): fix infinite rerender issue in useLegacyTable

riccardoperra:
This pr removes the `useStore` call in useLegacyTable since it's already used by the `useTable` hook with shallow selector. Table reference will change when that state is updated so we shouldn't need to call it twice in both hooks

[https://github.com/TanStack/table/blob/alpha/packages/react-table/src/useTable.ts#L114-L122](https://github.com/TanStack/table/blob/alpha/packages/react-table/src/useTable.ts#L114-L122)

I've also updated the Subscribe component to use the shallow selector to restore the `default`  behavior we have in other adapters and previous tanstack store versions

### DIFF
--- a/packages/react-table/src/Subscribe.ts
+++ b/packages/react-table/src/Subscribe.ts
@@ -1,6 +1,6 @@
 'use client'
 
-import { useStore } from '@tanstack/react-store'
+import { shallow, useStore } from '@tanstack/react-store'
 import type {
   NoInfer,
   RowData,
@@ -63,7 +63,7 @@ export function Subscribe<
 >(
   props: SubscribeProps<TFeatures, TData, TSelected>,
 ): ReturnType<FunctionComponent> {
-  const selected = useStore(props.table.store, props.selector)
+  const selected = useStore(props.table.store, props.selector, shallow)
 
   return typeof props.children === 'function'
     ? props.children(selected)
--- a/packages/react-table/src/useLegacyTable.ts
+++ b/packages/react-table/src/useLegacyTable.ts
@@ -15,8 +15,8 @@ import {
   sortFns,
   stockFeatures,
 } from '@tanstack/table-core'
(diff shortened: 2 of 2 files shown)


packages/react-table/src/useLegacyTable.ts
```diff
@@ -419,20 +419,21 @@ export function useLegacyTable<TData extends RowData>(
...
 
   return useMemo(
     () =>
       ({
         ...table,
-        getState: () => state,
+        getState: () => table.state,
       }) as LegacyReactTable<TData>,
-    [table, state],
+    [table],
```
KevinVandy, line 437: I'm trying to remember why I even did had a dedicated useStore call here, I think it might have had something to do with the react compiler. I'm fine shipping this though

packages/react-table/src/Subscribe.ts
```diff
@@ -63,7 +63,7 @@ export function Subscribe<
 >(
   props: SubscribeProps<TFeatures, TData, TSelected>,
 ): ReturnType<FunctionComponent> {
-  const selected = useStore(props.table.store, props.selector)
+  const selected = useStore(props.table.store, props.selector, shallow)
```
KevinVandy, line 66: thanks, this was needed.

packages/react-table/src/useLegacyTable.ts
riccardoperra: mmh could it be because previously there were some updates done in useEffect with queueMicrotask to update the baseStore? That's something we removed after `state` store updates automatically when options changes.

riccardoperra: Anyway, I just tested the useLegacyTable example with this configuration and it seems working the same

```
 react({
      babel: {
        plugins: ['babel-plugin-react-compiler'],
      },
    }),
```

PR #6217, opened 2026-04-02
fix(pagination): prevent onPaginationChange from firing when state is unchanged

### LINKED ISSUE #6158: Event `onPaginationChange` is triggered on every history push in RSC
### TanStack Table version

v8.21.3

### Framework/Library version

Next.js@15.5.7, React@19.0.0

#### Describe the issue
The reproduction is quite simple. I need to have parent component that is not a client component & pass prop `data` to the child `Table` component that has enabled pagination.

From now when anything is pushed to the URL (search params/hash) via Next.js's `Link` component or `router.push` from any level, this change will trigger `onPaginationChange` event. 

If I change the `page.tsx`  to client component via `"use client"` directive, it will no longer trigger the `onPaginationChange`, but in my case it needs to be a server component, since in this level I fetch the data from the server.

#### How I discovered it
In my setup I have Tanstack Table states persisted in the URL and when any state changes (sort/search filter) it will trigger `onPaginationChange` as well which is persisted to URL as well and this update updates the pagination search params URL even thought they haven't changed & this change will trigger again `onPaginationChange` (cuz URL has changed) resulting into infinite loop. 

https://codesandbox.io/p/devbox/6l8mtc

https://github.com/user-attachments/assets/0a441d2f-dbaa-48ef-93c6-d8befeeed658

No, because I do not know how

### Terms & Code of Conduct

solssak:
## Summary

Fixes #6158

- **Problem**: `onPaginationChange` fires on every URL navigation in React Server Components, even when pagination state hasn't changed. When a parent RSC re-renders (e.g., due to `router.push` or `Link` navigation), the `data` prop gets a new reference. This triggers `_autoResetPageIndex` → `resetPageIndex` → `setPagination` → `onPaginationChange`, even though `pageIndex` and `pageSize` are identical. When pagination is persisted to URL params, this creates an infinite loop.

- **Fix**: Add a shallow equality check in `setPagination` before invoking `onPaginationChange`. If the resolved new pagination state is identical to the current state, the callback is skipped entirely.

## Changes

- `packages/table-core/src/utils.ts` — Added `shallowEqual` utility function
- `packages/table-core/src/features/row-pagination/rowPaginationFeature.utils.ts` — `table_setPagination` now checks if state actually changed before calling `onPaginationChange`
- `packages/table-core/tests/unit/features/row-pagination/rowPaginationFeature.utils.test.ts` — Added 9 tests covering no-op updates for `table_setPagination`, `table_setPageIndex`, `table_resetPageIndex`, and `table_autoResetPageIndex`

## Verification

- All 239 table-core tests pass (including 9 new ones)
- TypeScript type check passes

## Summary by CodeRabbit

* **Refactor**
  * Optimized pagination state management to prevent redundant callback invocations when pagination values remain unchanged, reducing unnecessary updates and improving application performance with shallow equality checks.

* **Tests**
  * Added comprehensive unit tests for row pagination utilities, including validation of pagination state updates, callback triggers, page index management, reset behavior, and default state expectations.

### DIFF
--- a/packages/table-core/src/features/row-pagination/rowPaginationFeature.utils.ts
+++ b/packages/table-core/src/features/row-pagination/rowPaginationFeature.utils.ts
@@ -1,4 +1,4 @@
-import { functionalUpdate } from '../../utils'
+import { functionalUpdate, shallowEqual } from '../../utils'
 import type { RowData, Updater } from '../../types/type-utils'
 import type { TableFeatures } from '../../types/TableFeatures'
 import type { Table_Internal } from '../../types/Table'
@@ -35,6 +35,14 @@ export function table_setPagination<
     const newState = functionalUpdate(updater, old)
 
     return newState
+  }
+
+  const currentPagination = table.store.state.pagination
+  if (currentPagination) {
+    const newPagination = functionalUpdate(safeUpdater, currentPagination)
+    if (shallowEqual(currentPagination, newPagination)) {
+      return
+    }
   }
 
   return table.options.onPaginationChange?.(safeUpdater)
--- a/packages/table-core/src/utils.ts
+++ b/packages/table-core/src/utils.ts
@@ -9,6 +9,25 @@ export function functionalUpdate<T>(updater: Updater<T>, input: T): T {
     : updater
 }
 
+export function shallowEqual<T>(a: T, b: T): boolean {
+  if (Object.is(a, b)) return true
+  if (
+    typeof a !== 'object' ||
+    typeof b !== 'object' ||
+    a === null ||
+    b === null
+  ) {
+    return false
+  }
+  const keysA = Object.keys(a)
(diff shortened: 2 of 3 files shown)


KevinVandy:
I ended up applying a similar fix in alpha.37 before seeing this. Check if it's solved for you.

We could explore a more broad set state prevention refactor when current state shallow equals the next state update if we want.

solssak:
@KevinVandy, thanks for the heads up! I just verified — yeah, you're right, it's already fixed in alpha.37. My local setup was apparently out of date, sorry for the noise on that.

Really appreciate you taking the time to look into this and share the broader refactor idea too. A more general "skip update when next state shallow-equals current" pass sounds like a nice direction — happy to take a stab at that as a follow-up PR if it'd be useful.

Thanks again for the consistently kind and thoughtful reviews 🙏

PR #6221, opened 2026-04-03
feat(angular): produce build th Vite

riccardoperra:
This PR removes ng-packagr angular build in favor of Vite 8


KevinVandy:
pretty sure we want to close this? re-open as non-draft if it's still a good idea

PR #6240, opened 2026-04-29
fix(react-table): type column contexts with React table

### LINKED ISSUE #6230: HeaderContext and CellContext types should use ReactTable
### TanStack Table version

9.0.0-alpha.32

### Framework/Library version

react@19.1.0

When defining `cell` or `header` in the column definitions, the type of the props provided, `CellContext` and `HeaderContext` respectively, is inaccurate. The type of the `table` prop should be `ReactTable` instead of `Table`.  Without, a cell cannot use `table.Subscribe` that became available in the alpha 9.

In the example, you can see in the logs that the table provided as prop does actually have the properties.

https://codesandbox.io/p/sandbox/yq2lwx

_No response_

No, because I do not have time to dig into it

### Terms & Code of Conduct

cyphercodes:
## 🎯 Changes

Fixes #6230.

- Type React column `cell`, `header`, and `footer` contexts so their `table` prop is the React table instance and exposes helpers like `table.Subscribe`.
- Re-export the React adapter's context, column definition, and table option types so `ColumnDef` from `@tanstack/react-table` uses those React-specific contexts.
- Add a type regression test covering `table.Subscribe` usage from both header and cell render callbacks.

## ✅ Checklist

- [ ] I have tested this code locally with `pnpm test:pr`.

Targeted checks run instead:

- `pnpm exec prettier --check packages/react-table/src/useTable.ts packages/react-table/src/index.ts packages/react-table/tests/column-context.test-d.ts`
- `pnpm --filter @tanstack/react-table test:eslint`
- `pnpm --filter @tanstack/react-table test:types`
- `pnpm --filter @tanstack/react-table test:lib`
- `pnpm --filter @tanstack/react-table build`
- `git diff --check`

## Summary by CodeRabbit

* **New Features**
  * New TypeScript types for column definitions and table contexts with enhanced type inference capabilities
  * Explicit type exports improve developer experience and IDE autocompletion support

* **Refactor**
  * Updated public API export structure to explicitly define the package surface, replacing wildcard exports with named type exports

### DIFF
--- a/packages/react-table/src/index.ts
+++ b/packages/react-table/src/index.ts
@@ -3,4 +3,16 @@ export * from '@tanstack/table-core'
 export * from './FlexRender'
 export * from './Subscribe'
 export * from './createTableHook'
-export * from './useTable'
+export { useTable } from './useTable'
+export type {
+  AccessorColumnDef,
+  AccessorFnColumnDef,
+  AccessorKeyColumnDef,
+  CellContext,
+  ColumnDef,
+  DisplayColumnDef,
+  GroupColumnDef,
+  HeaderContext,
+  ReactTable,
+  TableOptions,
+} from './useTable'
--- a/packages/react-table/src/useTable.ts
+++ b/packages/react-table/src/useTable.ts
@@ -7,10 +7,18 @@ import { FlexRender } from './FlexRender'
 import { Subscribe } from './Subscribe'
 import type {
   CellData,
+  ColumnDefTemplate,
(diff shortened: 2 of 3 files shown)


KevinVandy:
This kind of fixes the issue, but I fear that it can get pretty messy. There are lots more `table`, `column`, `footer`, `cell`, etc. instances that would take way too many type replacements like this if we wanted to go all the way. I'm not against this PR but not sure it's the right way to go.

PR #6242, opened 2026-05-05
WIP: fix(solid-table): adapt to Solid v2 beta reactivity model

brenelz:
- Wrap <Match when> props in FlexRender with accessor functions and use createMemo inside Match callbacks so reactive reads happen in tracking scopes (resolves STRICT_READ_UNTRACKED warnings).
- Update basic-app-table example to call accessor args (row(), cell(), header(), footerGroup()) returned by Solid v2's <For> callbacks.

### DIFF
--- a/packages/table-devtools/package.json
+++ b/packages/table-devtools/package.json
@@ -50,7 +50,7 @@
     "@tanstack/devtools-utils": "^0.4.0",
     "@tanstack/solid-store": "^0.11.0",
     "goober": "^2.1.18",
-    "solid-js": "^1.9.12"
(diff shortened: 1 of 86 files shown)


packages/table-devtools/package.json
```diff
@@ -50,7 +50,7 @@
     "@tanstack/devtools-utils": "^0.4.0",
     "@tanstack/solid-store": "^0.11.0",
     "goober": "^2.1.18",
-    "solid-js": "^1.9.12"
+    "solid-js": "^1.9.7"
```
birkskyum, line 53: why downgrade solid-js here?

PR #6248, opened 2026-05-09
feat: add angular devtools support and fix core devtools reactivity

riccardoperra:
**TanStack Devtools Angular package is not published yet in npm**

- add new @tanstack/angular-table-devtools package
- add devtools integration  in composable tables

## Summary by CodeRabbit

* **New Features**
  * Devtools integration added (dev-mode, lazy-loaded) across frameworks with table devtools panels and wiring for interactive inspection.
* **Bug Fixes / UX**
  * Examples and templates now read from the public table.state surface — fixes debug panels, pagination, resizing, DnD, filters and live bindings.
* **Documentation**
  * Migration and framework guides updated to reference table.state.
* **Chores / Examples**
  * New Angular example apps added and many examples/templates updated.

[Review Change Stack](https://app.coderabbit.ai/change-stack/TanStack/table/pull/6248?utm_source=github_walkthrough&utm_medium=github&utm_campaign=change_stack)


packages/react-table/src/useTable.ts
```diff
@@ -137,6 +141,12 @@ export function useTable<
   tableOptions: TableOptions<TFeatures, TData>,
   selector?: (state: TableState<TFeatures>) => TSelected,
 ): ReactTable<TFeatures, TData, TSelected> {
+  const instanceIdRef = useRef<string | undefined>(undefined)
+  const id = useId()
```
KevinVandy: this technically makes this a React 18+ package now. We could just require an id/key table option in order for devtools to work, and then remove the 2nd arg from useTanStackTableDevtools(table)

riccardoperra: What about removing entirely useId :thinking: ? There is still the generation logic provided next. it was just added as a prefix but probably unnecessary



```diff
-  const id = useId()
   if (!instanceIdRef.current) {
    // eslint-disable-next-line @eslint-react/purity
-     instanceIdRef.current = `${id}-${Math.random().toString(36).slice(2, 9)}`
+     instanceIdRef.current = Math.random().toString(36).slice(2, 9);
   }
```

KevinVandy: Using math.random for anything in a library usually causes security scanners to freak out, even if it's for something trivial like this.

riccardoperra: switched with crypto.randomUUID (if present) or just id++ for this

KevinVandy (approved):

PR #6266, opened 2026-05-16
Performance Fix: Solving Quadratic Slowdown in svelte-table

### LINKED ISSUE #6235: `svelte-table`: nested mergeObjects in setOptions produces quadratic slowdown on rapid state changes
### TanStack Table version

9.0.0-alpha.32

### Framework/Library version

Svelte 5.55.4

## Summary

`mergeObjects` has had this shape for a long time, but the current `@tanstack/svelte-table` adapter makes the consequences prominent: its `$effect.pre` calls `setOptions((prev) => mergeObjects(prev, mergedOptions))` on every reactive tick. Because `mergeObjects` defines getters whose closures reference the previous result, `table.options` after N ticks is an N-deep chain. Reads of any key that lives only at the bottom of the chain — e.g. feature defaults installed by `getDefaultTableOptions` — walk the entire chain, making each read O(N). Row-model pipelines read those defaults many times per row per tick, producing O(N²) aggregate work and user-visible freezes while typing into a global filter input.

We hit this after bumping from `9.0.0-alpha.10` to `9.0.0-alpha.32`. The `mergeObjects` implementation is identical between the two, but the newer adapter wires up the effect so that every reactive state change fires `setOptions`, letting the chain grow much faster in practice.

**Scope:** only measured on `@tanstack/svelte-table`. The `table_mergeOptions` default in `@tanstack/table-core` (`(defaults, next) => mergeObjects(defaults, next)`) is the structural source of the chain growth, so other adapters (React, Vue, Solid, Angular) may be affected to the extent their integrations also re-call `setOptions` on each render/reactive update — not verified.

## Reproduction

https://github.com/AndersRobstad/tanstack-merge-repro ([open in StackBlitz](https://stackblitz.com/github/AndersRobstad/tanstack-merge-repro))

Type 30–50 characters into the filter input and watch **default-only ms** (10k reads of `onExpandedChange`, installed by `rowExpandingFeature.getDefaultTableOptions`) climb. **user-key ms** (`enableRowSelection`, user-provided) stays flat.

Measured on 500 rows, 7 features active:

| keystrokes | chain depth | default-only read (20k) | user-key read (20k) |
|-----------:|------------:|------------------------:|--------------------:|
|          0 |           2 | ~2 ms                   | ~0.2 ms             |
|         50 |         103 | ~94 ms                  | ~0.2 ms             |
|        100 |         203 | ~275 ms                 | ~1.5 ms             |

**Expected:** typing into a global filter stays responsive indefinitely; `table.options` reads are O(1) regardless of how many times the effect has run.

**Actual:** `table.options` is an N-deep `mergeObjects` chain after N reactive ticks. Feature-default-key reads scale linearly with N. A Chrome performance profile on a production app we maintain showed the `mergeObjects` getter dominating CPU, with per-keystroke latency growing 15 → 183 → 374 → 899 → 1854 → 3486 → 7043 → 13923 ms across eight consecutive keystrokes.

## Environment

- Reproduced in StackBlitz (WebContainer, Chrome) and locally (macOS 15.3, Chrome 131, Node 22).

## Package versions

- `@tanstack/svelte-table` 9.0.0-alpha.32
- `@tanstack/table-core` 9.0.0-alpha.32
- `svelte` 5.55.4

## Root cause

`@tanstack/svelte-table/dist/createTable.svelte.js`:

```js
$effect.pre(() => {
  void mergedOptions.state;
  void mergedOptions.data;
  untrack(() => {
    table.setOptions((prev) => {
      return mergeObjects(prev, mergedOptions); // wraps prev into a new getter-chain layer
    });
  });
});
```

Each tick:
1. `functionalUpdate(updater, table.options)` → `mergeObjects(prev, mergedOptions)` — depth + 1
2. `table_mergeOptions` via the adapter-installed default `(defaults, next) => mergeObjects(defaults, next)` — depth + 1
3. Stored in `optionsStore`; `table.options` returns this chain

User-provided options (`data`, `columns`, `state`, `enableX`) are in the `mergedOptions` closure — found on the first source check, O(1) reads. Feature-default options (`onExpandedChange`, `onColumnVisibilityChange`, `onRowSelectionChange`, `onSortingChange`, `onColumnFiltersChange`, `getColumnCanGlobalFilter`, `isMultiSortEvent`) live only in the flat initial store state `{...defaultOptions, ...tableOptions}` at the bottom — every read walks the full chain, O(N).

## Why this was latent

Hot code paths that read user options aren't affected. You only feel it when the row-model pipeline (filter, sort, expand) reads a feature-default option many times per row, and the chain is allowed to grow across many reactive ticks in quick succession (e.g. rapid keystrokes in a filter input). Older adapter versions triggered `setOptions` less aggressively, so the chain didn't grow fast enough for this to be visible in practice.

## Workaround

Override `mergeOptions` at the user level to resolve each key once and store plain data descriptors. Live implementation in the repro: [src/App.svelte#L52-L73](https://github.com/AndersRobstad/tanstack-merge-repro/blob/c347ee92031cc726c06443da3988007b936f997d/src/App.svelte#L52-L73).

```ts
createTable({
  mergeOptions: (prev, next) => {
    const result: Record<PropertyKey, unknown> = {};
    const allKeys = new Set([...Reflect.ownKeys(prev), ...Reflect.ownKeys(next)]);
    for (const key of allKeys) {
      // `undefined` from `next` means "next didn't define this key" (not
      // "next explicitly set undefined"), so we fall through to `prev`.
      const fromNext = (next as Record<PropertyKey, unknown>)[key];
      result[key as string] = fromNext !== undefined ? fromNext : (prev as Record<PropertyKey, unknown>)[key];
    }
    return result;
  },
  // ...rest of options
});
```

Reactivity is preserved because the adapter's `$effect.pre` fires `setOptions` on every state/data change, so `table.options` is repopulated with fresh snapshots before any consumer reads it within the tick.

A descriptor-copy merger that preserves getters (`Object.defineProperty(result, key, { ...desc, configurable: true })`) is **not** sufficient — the chain of getter closures still grows, just with half the constant. The repro's "Use flat `mergeOptions` override (fix)" checkbox toggles this true-flat implementation on, demonstrating that the same column's read time drops from ~275 ms to ~0 ms at 100 keystrokes.

## Suggested upstream fix

Either:

1. **Adapter level**: in `createTable.svelte.js`, write a fresh resolved snapshot instead of wrapping `prev`.
2. **Core level**: change the default `mergeOptions` in `table_mergeOptions` so it resolves and flattens instead of nesting `mergeObjects`.

https://stackblitz.com/~/github.com/AndersRobstad/tanstack-merge-repro

https://github.com/user-attachments/assets/263e05ef-d7e3-489c-b5d3-9a697d35b3ad

No, because I do not know how

### Terms & Code of Conduct

Abyanzhafran:
## 🎯 Changes

Fixes https://github.com/TanStack/table/issues/6235
### Root Cause Analysis (RCA)
The performance degradation observed in the Svelte 5 adapter was caused by a **nested getter chain** created during rapid reactive updates.

1.  **The Mechanism**: The `mergeObjects` utility (inherited from SolidJS patterns) creates a proxy-like object using lazy getters. These getters do not resolve values eagerly; instead, they walk through a `sources` array via a closure every time a property is read.
2.  **The Loop**: In `createTable.svelte.ts`, the `$effect.pre` block calls `setOptions` on every reactive tick (e.g., every keystroke in a filter input). Each call to `setOptions` wraps the *previous* options object in a new layer of `mergeObjects`.
3.  **The Bottleneck**: After **N** reactive updates, `table.options` becomes an **N-deep chain** of closures. Reading a value that lives in the base defaults (like feature configurations) requires traversing all **N** layers.
4.  **Complexity**: Since the row-model pipeline reads these options multiple times per row per tick, the aggregate work scales at **O(N²)**, leading to the reported UI freezes during rapid input.

---

### The Solution: Eager Resolution with `flatMerge`

Introduced a new utility, `flatMerge`, to break the recursion chain in hot reactive paths.

#### `flatMerge` vs `mergeObjects`
*   **`mergeObjects` (Lazy)**: Preserves getters. Great for initial setup where you want to maintain a "live" link to reactive state.
*   **`flatMerge` (Eager)**: Resolves all values immediately and copies them into a plain, flat JavaScript object. This transforms the lookup complexity from **O(N)** back to **O(1)**.

#### Implementation Strategy
I implemented a targeted fix that preserves reactivity while eliminating the slowdown:
1.  **Preserved Initialization**: `mergeObjects` is still used during the initial table setup to ensure the adapter correctly "captures" Svelte's reactive getters.
2.  **Flattened Updates**: Inside the `$effect.pre` loop and the `mergeOptions` handler, we switched to `flatMerge`. 
    *   Since `$effect.pre` re-runs automatically when dependencies change, we don't need the resulting object to have getters; the effect itself acts as the "pump" that pushes fresh, flat data into the table core.

---

### Performance Impact
*   **Before**: Reading a default option after 50 keystrokes = **50 nested function calls** per read. Aggregate work is quadratic.
*   **After**: Reading a default option after 50 keystrokes = **1 direct property access**. Aggregate work is linear/constant relative to state updates.

### Code Summary

```typescript
// New utility to prevent getter-chain accumulation
export function flatMerge(...sources: any): any {
  const result = {}
  for (let source of sources) {
    if (typeof source === 'function') source = source()
    if (!source) continue
    for (const key of Reflect.ownKeys(source)) {
      const value = source[key]
      if (value !== undefined) {
        result[key] = value
      }
    }
  }
  return result
}
```

This fix ensures `@tanstack/svelte-table` remains performant even in complex tables with rapid state changes, while maintaining full compatibility with Svelte 5's reactivity model.

### How to reproduce in our local env
*   Clone this updated solution
*   Use `pnpm link` to connect into your reproducible project

## ✅ Checklist

## Summary by CodeRabbit

* **Refactor**
  * Changed how table options are resolved and merged during table initialization and updates.

[Review Change Stack](https://app.coderabbit.ai/change-stack/TanStack/table/pull/6266?utm_source=github_walkthrough&utm_medium=github&utm_campaign=change_stack)


KevinVandy:
So now that we have `table.optionsStore` which subscribes to changes in data, should the custom `mergeObjects` probably just be removed? @riccardoperra  fyi @walker-tx

KevinVandy:
I see our svelte `basic-external-atoms` example is broken, but it's also broken in alpha apparently. I might have to follow up

KevinVandy (approved):

PR #6295, opened 2026-06-04
fix(table-core): clamp column size to minSize/maxSize during resize drag

### LINKED ISSUE #5997: ColumnSizing's `getResizeHandler` does not respect columnDef minSize nor maxSize
### TanStack Table version

v8.21.3

### Framework/Library version

React 17.0.2

1. Define a table with header and body but override the display from table (or table-cell / table-row) to flex
2. Define columns with `minSize` or `maxSize` 
3. Enable the resize (`enableColumnResizing`)
4. On your header size slider element apply `onMouseDown` prop with the value `header.getResizeHandler()` (where header is the tanstack's header)
5. Try to resize a column

Actual behavior:
Resize does not respect the minSize or maxSize according to the columnDef but resize the column according to the mouse end position

Expected behavior:
Resize SHOULD respect the columnDef minSize and maxSize.

Additional context:
1. You can reproduce it by visiting the columnSizing example from Tanstack official examples: [column-sizing](https://stackblitz.com/github/tanstack/table/tree/main/examples/react/column-sizing) and see by yourself you can resizing the column below the minSize :/

2. tanstack's implementation - tableCore/src/features/columnSizing.ts L390-L395:
```typescript
old.columnSizingStart.forEach(([columnId, headerSize]) => {
              newColumnSizing[columnId] =
                Math.round(
                  Math.max(headerSize + headerSize * deltaPercentage, 0) * 100
                ) / 100
            })
```

IMHO should be:
```typescript
const columnDef = header.column.columnDef;
const newSize = Math.round(
  Math.max(headerSize + headerSize * deltaPercentage, 0) * 100
) / 100;

newColumnSizing[columnId] = Math.min(
  columnDef?.maxSize ?? Infinity,
  Math.max(columnDef?.minSize ?? 0, newSize)
);
```

https://stackblitz.com/github/tanstack/table/tree/main/examples/react/column-sizing

Image

### Terms & Code of Conduct

LeSingh1:
Closes #5997

## What was wrong

When a user drags a resize handle, `header_getResizeHandler` computes new column widths inside a `forEach` over `columnSizingStart`. That computation never consulted `columnDef.minSize` or `columnDef.maxSize`, so dragging far enough in either direction would write a committed size that violated the bounds. `column_getSize` does clamp when reading committed state, but the value gets written into `columnSizing` unclamped, so any code that reads `state.columnSizing[id]` directly (e.g. performant resize examples that skip the full recalculation) sees the out-of-bounds number.

## What the fix does

At drag-start time the handler already iterates `header.getLeafHeaders()` to build `columnSizingStart`. This PR adds a second pass that records each leaf column's `minSize` and `maxSize` from its `columnDef` (falling back to the same defaults `column_getSize` uses) into a closure map called `columnSizingConstraints`. During each drag-move update that map is used to clamp the computed size before it is written into `newColumnSizing`:

```ts
Math.min(maxSize, Math.max(minSize, rawSize))
```

The existing `Math.max(..., 0)` floor is replaced by this — the default `minSize` of 20 is stricter than 0 anyway, and custom `minSize: 0` is preserved correctly.

## Tests

Two new tests in `columnResizingFeature.utils.test.ts`:

- dragging far left on a column with `minSize: 50` keeps the committed size at or above 50
- dragging far right on a column with `maxSize: 150` keeps the committed size at or below 150

Both tests failed against `main` before this change and pass after.

Full suite: 295 tests, 19 test files, all pass.

## Release Notes

* **Bug Fixes**
  * Column resizing now clamps column widths to configured minimum and maximum size constraints.

* **Tests**
  * Added unit tests to verify min/max size constraint behavior during column resizing operations.

### DIFF
--- a/packages/table-core/src/features/column-resizing/columnResizingFeature.utils.ts
+++ b/packages/table-core/src/features/column-resizing/columnResizingFeature.utils.ts
@@ -163,9 +163,8 @@ export function header_getResizeHandler<
           const minSize = constraints?.minSize ?? 0
           const maxSize = constraints?.maxSize ?? Number.MAX_SAFE_INTEGER
           newColumnSizing[columnId] =
-            Math.round(
-              Math.min(maxSize, Math.max(minSize, rawSize)) * 100,
-            ) / 100
+            Math.round(Math.min(maxSize, Math.max(minSize, rawSize)) * 100) /
+            100
         })
 
         return {
--- a/packages/table-core/src/features/column-resizing/columnResizingFeature.utils.ts
+++ b/packages/table-core/src/features/column-resizing/columnResizingFeature.utils.ts
@@ -1,5 +1,6 @@
 import {
   column_getSize,
+  getDefaultColumnSizingColumnDef,
   header_getSize,
   table_setColumnSizing,
 } from '../column-sizing/columnSizingFeature.utils'
@@ -119,6 +120,18 @@ export function header_getResizeHandler<
         column_getSize(leafHeader.column),
       ])
 
+    const defaultSizes = getDefaultColumnSizingColumnDef()
+    const columnSizingConstraints: Record<
+      string,
+      { minSize: number; maxSize: number }
+    > = {}
+    header.getLeafHeaders().forEach((leafHeader) => {
+      columnSizingConstraints[leafHeader.column.id] = {
+        minSize: leafHeader.column.columnDef.minSize ?? defaultSizes.minSize,
+        maxSize: leafHeader.column.columnDef.maxSize ?? defaultSizes.maxSize,
+      }
+    })
+
     const clientX = isTouchStartEvent(event)
       ? Math.round(event.touches[0]!.clientX)
       : (event as MouseEvent).clientX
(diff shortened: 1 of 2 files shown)


MILLERMARRU (commented):
Traced through where `old.columnSizingStart` gets built to check whether `constraints` can actually be undefined at the fallback line. It's populated in the same `header_getResizeHandler` closure, from the same `header.getLeafHeaders()` call that builds `columnSizingConstraints`, at the same mousedown, so the two will always have identical column ids and the `constraints?.minSize ?? 0` / `?? Number.MAX_SAFE_INTEGER` fallback can't actually trigger today. CodeRabbit already flagged this as a nitpick, but it's worth being precise about why it still matters: `getDefaultColumnSizingColumnDef()` returns `minSize: 20`, not `0`, so the fallback you wrote doesn't match the library's real default, it's just unreachable right now. If this constraints map ever gets reused somewhere the leaf header set can diverge (a follow-up refactor, a different call site), a column could silently be allowed to shrink to 0px instead of respecting the actual 20px floor. Since `defaultSizes` is already destructured two lines above, swapping in `defaultSizes.minSize` and `defaultSizes.maxSize` costs nothing and removes that trap.

Setting that aside, the fix itself is right. The old code only clamped the lower bound at `0`, so a column with a `minSize` set in its columnDef could still be dragged smaller than that, and there was no `maxSize` enforcement at all during an active drag, only at the initial size resolution. Reading the raw size first and clamping with `Math.min(maxSize, Math.max(minSize, rawSize))` after is the correct order, and doing it per column id inside the `columnSizingStart.forEach` means multi-column resize (colSpan headers) respects each leaf's own bounds instead of one shared clamp.

The two new tests exercise the actual drag handler end to end through synthetic mouse events rather than calling some inner helper directly, so they're testing what a user would actually trigger, and dragging far enough past the bound in each test (to ~10px and ~600px) makes sure the assertion isn't accidentally passing because the drag distance was too small to matter.

PR #6298, opened 2026-06-06
feat: enable support for native reactivity in solid vue adapters

riccardoperra:
Updated Vue and Solid table reactivity to rely on framework-native primitives and table.atoms instead of the framework-specific TanStack Store adapters.

Summary:

- Removed @tanstack/vue-store and @tanstack/solid-store from the Vue/Solid table adapters.
- Switched Vue/Solid reactivity bindings to native refs/signals/memos while keeping core atom/store compatibility through @tanstack/store.
- Removed the adapter-level table.state convenience API from Angular, Vue, and Solid. Updated examples to read state through table.store.get() / atoms instead of table.state. 
- Simplified Vue/Solid table.Subscribe so it no longer accepts source or selector; it now exposes table.atoms to the child callback, letting the framework track only the atom reads used by that render boundary.
- Removed selector-based Subscribe plumbing and unused selector helpers.

---

- table.atoms now provides the granular subscription model directly.
- This avoids subscribing through full-table selectors and makes reactive reads explicit
- . Vue and Solid can rely on their own reactive primitives for memoization/tracking, reducing adapter dependencies and simplifying the API.

## Summary by CodeRabbit

* **Refactor**
  * Examples (Angular, Solid, Vue) now derive and display table state from internal store/atoms instead of the previous public state view; pagination and debug UIs reflect store/atom values.
  * Solid- and Vue-table APIs simplified: selector-based subscriptions removed; Subscribe now provides atoms to consumers.

* **Chores**
  * Reactivity adapters and manifests unified to use the shared store runtime.

* **Tests**
  * Unit and example tests updated to assert atoms/store-based state and Subscribe behavior.


examples/angular/composable-tables/src/app/components/table-components.ts
```diff
@@ -88,12 +89,12 @@ export class RowCount {
           type="number"
           min="1"
           [max]="table().getPageCount()"
-          [value]="table().state.pagination.pageIndex + 1"
```
KevinVandy, line 91: These should probably just be atom lookups, but fine either way

examples/solid/column-resizing/src/App.tsx
```diff
@@ -97,7 +97,7 @@ function App() {
     header: ReturnType<typeof table.getHeaderGroups>[number]['headers'][number],
   ) => {
     if (columnResizeMode() === 'onEnd' && header.column.getIsResizing()) {
-      const delta = table.state().columnResizing.deltaOffset ?? 0
+      const delta = table.store.get().columnResizing.deltaOffset ?? 0
```
KevinVandy: makes more sense for atom read for this line since it's one slice of state

examples/solid/composable-tables/src/components/table-components.tsx
```diff
@@ -13,7 +13,7 @@ import { useTableContext } from '../hooks/table'
 export function PaginationControls() {
   const table = useTableContext()
 
-  const pagination = createMemo(() => table.state().pagination)
+  const pagination = createMemo(() => table.store.get().pagination)
```
KevinVandy: read pagination atom instead of store

KevinVandy (approved):

PR #6330, opened 2026-06-17
fix(filterFns): use autoRemove instead of resolveFilterValue for comparison filters (#6212)

rafaumeu:
Fixes #6212. Replaces testFalsy() with autoRemove for filterFn_greaterThan, filterFn_greaterThanOrEqualTo, filterFn_lessThan, and filterFn_lessThanOrEqualTo. This fixes incorrect comparisons where filter values were coerced to false/0, causing all non-zero rows to pass filters like 'price > 50'.

**Test Results:**
- 51 tests passed in filterFns.test.ts
- No breaking changes

**Changes:**
- 1 file modified (packages/table-core/src/fns/filterFns.ts)
- 4 lines changed (resolveFilterValue → autoRemove)


KevinVandy:
merge conflicts, probably wrong branch

PR #6331, opened 2026-06-17
fix: AppTable remounting children on table state updates

Abyanzhafran:
App wrappers (AppTable, AppCell, AppHeader, AppFooter) were memoized on the unstable `table` reference, causing them to be recreated every render. React then remounted their entire subtrees on every state change, destroying controlled input focus.

Keep wrappers stable (created once) and read the current table from a ref updated each render. Fixes children remounting while preserving latest table state access.

Fixes : https://github.com/TanStack/table/pull/6331

## 🎯 Changes

# RCA: AppTable remounting children on every table state update

## Problem
`table.AppTable` (and `AppCell`, `AppHeader`, `AppFooter`) created a new function reference on every render. React uses function identity to decide if a component should re-render or remount. A new function = unmount + remount the entire subtree, destroying controlled input focus.

## Root Cause
1. `useTable` returns a fresh `table` reference every render (by design, for React Compiler compatibility)
2. `useAppTable` feeds a new options object literal to `useTable` every render
3. The four App wrappers were memoized with `[table]` as a dependency
4. Since `[table]` changes every render, the wrappers were recreated every render
5. React saw a new component type → unmounted children instead of re-rendering them
6. A controlled input in a toolbar lost focus on every keystroke

## Solution
Decouple **component identity** (must be stable) from **the table reference the component reads** (must be current) using a ref pattern:

1. Import `useRef`
2. Create a ref that tracks the current table:
   ```js
   const tableRef = useRef(table)
   tableRef.current = table


## ✅ Checklist

<!-- This is an auto-generated comment: release notes by coderabbit.ai -->
## Summary by CodeRabbit

* **Bug Fixes**
  * Improved stability of table wrapper components to prevent unnecessary recreation/remounts during frequent table updates by using a stable internal table reference.

* **New Features**
  * Added row-selection “select” columns across React, Preact, Angular, Solid, Svelte, Vue, and Lit examples, including synchronized header/row checkboxes with indeterminate (some selected) behavior.
  * Introduced reusable indeterminate checkbox components in the respective example frameworks.
  * Enhanced example table layout with a fixed-height scroll container and sticky table headers.
<!-- end of auto-generated comment: release notes by coderabbit.ai -->

### DIFF
--- a/examples/angular/composable-tables/src/app/components/cell-components.ts
+++ b/examples/angular/composable-tables/src/app/components/cell-components.ts
@@ -1,4 +1,4 @@
-import { Component, computed } from '@angular/core'
+import { ChangeDetectionStrategy, Component, computed } from '@angular/core'
 import { injectFlexRenderContext } from '@tanstack/angular-table'
 import { CurrencyPipe } from '@angular/common'
 import { injectTableCellContext } from '../table'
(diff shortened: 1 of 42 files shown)


KevinVandy:
Awesome, I'll try this out soon

PR #6337, opened 2026-06-19
fix(table-core): prototype-named column ids poisoning row value caches

artchen-db:
## 🎯 Changes

A column whose `id` equals an `Object.prototype` member (most impactfully `hasOwnProperty`) crashes filtering/sorting:

```
TypeError: this._valuesCache.hasOwnProperty is not a function
```

**Root cause:** the per-row value caches are plain `{}` objects probed with `cache.hasOwnProperty(columnId)`, then written keyed by column id. When `columnId === "hasOwnProperty"`, the write shadows the inherited method, so the next probe invokes a cached value as a function and throws — breaking filtering for every column on that row, and sorting on the offending column.

**Fix:** initialize the caches (`_valuesCache`, `_uniqueValuesCache`, `_groupingValuesCache`) with `Object.create(null)` and read them via `Object.prototype.hasOwnProperty.call(...)`. Also null-prototype `_getAllCellsByColumnId` so ids like `toString`/`constructor` resolve correctly when read via bracket access in `getColumnCanGlobalFilter`. This removes the entire prototype-collision class for these caches.

Added `tests/prototypeColumnId.test.ts` covering `getValue`, `getUniqueValues`, and the cascading-failure case across `hasOwnProperty`, `toString`, `constructor`, `valueOf`, `__proto__`, `isPrototypeOf`. The test fails without the fix and passes with it.

## ✅ Checklist

## 🚀 Release Impact

### DIFF
--- a/.changeset/proud-pets-fix.md
+++ b/.changeset/proud-pets-fix.md
@@ -1,0 +1,7 @@
+---
+'@tanstack/table-core': patch
+---
+
+fix prototype-named column ids (e.g. `hasOwnProperty`, `toString`) crashing filtering/sorting
+
+Row value caches (`_valuesCache`, `_uniqueValuesCache`, `_groupingValuesCache`) were plain objects probed with `cache.hasOwnProperty(columnId)`. A column whose id collided with an `Object.prototype` member shadowed the inherited method on write, so the next probe invoked a cached value as a function and threw `TypeError`. Caches are now created with `Object.create(null)` and read via `Object.prototype.hasOwnProperty.call`, and `_getAllCellsByColumnId` is likewise null-prototyped so prototype-named ids resolve correctly.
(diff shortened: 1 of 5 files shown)


KevinVandy:
why would a column id be named a javascript reserved method name?

artchen-db:
> why would a column id be named a javascript reserved method name?

IMO it is entirely possible and fair to support. Especially if the table is used to display arbitrary data from users. Imagine a database app using tanstack table to display arbitrary query results - the column names are entirely up to the users.

KevinVandy:
I'm going to experiment with this in v9 here: https://github.com/TanStack/table/pull/6338

artchen-db:
@KevinVandy would you be open to applying the fix to v8?

PR #6353, opened 2026-06-27
fix(table-core): guard Array.isArray before index access in range filter autoRemove

### LINKED ISSUE #5949: setColumnFilter(num) fails to set filter value
### TanStack Table version

8.21.2

### Framework/Library version

React 18.2.0

When I set a numeric column filter, it immediately reverts to undefined, but it works if I set it to a string value.

```typescript
const col = table.getColumn('colour');
const onClick = () => {
   col.setColumnFilter('purple');
};
useEffect(() => {
   console.log(col.getColumnFilter()); // prints undefined, undefined, purple
}, [col]);
```

But when I use a numeric column filter value it behaves differently.
```typescript
const col = table.getColumn('colourId');
const onClick = () => {
   col.setColumnFilter(99);
};
useEffect(() => {
   console.log(col.getColumnFilter()); // prints undefined, undefined
}, [col]);
```

The sandbox shows the issue pretty clearly.

* Click the "Red colour" filter button and view that the filter works.
* Click the "Clear filters" button to reset the filters.
* Click the "Red colourId" filter button and see that it does nothing. Essentially the filter set up in `showRedByIdClicked()` has not effect... 

Let me know if there is more I can do to help.

https://codesandbox.io/p/sandbox/kzxcl2

_No response_

No, because I do not know how

### Terms & Code of Conduct

JSap0914:
## Bug

`filterFn_between`, `filterFn_betweenInclusive`, and `filterFn_inNumberRange` each contain:

```ts
autoRemove: (val: any) =>
  testFalsy(val) || (testFalsy(val[0]) && testFalsy(val[1]))
```

When a **scalar** (non-array) value such as `99` is passed, `val[0]` and `val[1]` are both `undefined`. `testFalsy(undefined) === true`, so the condition evaluates to `true && true` = `true` — the filter is immediately auto-removed even though `99` is a perfectly valid, non-empty filter value.

Numeric columns default to the `inNumberRange` filter function (auto-selected by `column_getAutoFilterFn`). Calling `col.setColumnFilter(99)` on a numeric column therefore silently discards the filter.

Closes #5949

## Fix

Add `Array.isArray(val) &&` before accessing `val[0]` / `val[1]`, so the index checks only execute when `val` is actually an array. The leading `testFalsy(val)` still handles `undefined`, `null`, and `''`.

```ts
// Before
autoRemove: (val: any) =>
  testFalsy(val) || (testFalsy(val[0]) && testFalsy(val[1]))

// After
autoRemove: (val: any) =>
  testFalsy(val) || (Array.isArray(val) && testFalsy(val[0]) && testFalsy(val[1]))
```

The same pattern is applied identically to all three range filter functions that shared the bug.

## Verification

```
pnpm vitest run packages/table-core/tests/unit/fns/filterFns.test.ts
```

**Before fix:** 1 test fails — `filterFns.inNumberRange.autoRemove > should NOT auto-remove a scalar number` (`autoRemove(99)` returned `true`, expected `false`)

**After fix:** 358/358 tests pass (349 pre-existing + 9 new for `filterFns.inNumberRange.autoRemove`)

> This fix was developed with AI assistance.

## Summary by CodeRabbit

* **Bug Fixes**
  * Improved range filter cleanup logic to safely handle non-array values without incorrectly treating them as invalid.
  * Preserved auto-removal for empty/missing range endpoints while keeping valid inputs (including `0`) and scalar values from being dropped.
* **Tests**
  * Added unit tests covering `inNumberRange` auto-removal behavior for undefined/null/empty strings, empty endpoint arrays, and valid numeric ranges (including regression checks for scalar `99` and `0`).

### DIFF
--- a/packages/table-core/src/fns/filterFns.ts
+++ b/packages/table-core/src/fns/filterFns.ts
@@ -219,7 +219,7 @@ const filterFn_between = Object.assign(
       filterFn_lessThan(row, columnId, filterValues[1])),
   {
     autoRemove: (val: any) =>
-      testFalsy(val) || (testFalsy(val[0]) && testFalsy(val[1])),
+      testFalsy(val) || (Array.isArray(val) && testFalsy(val[0]) && testFalsy(val[1])),
   },
 )
 
@@ -243,7 +243,7 @@ const filterFn_betweenInclusive = Object.assign(
       filterFn_lessThanOrEqualTo(row, columnId, filterValues[1])),
   {
     autoRemove: (val: any) =>
-      testFalsy(val) || (testFalsy(val[0]) && testFalsy(val[1])),
+      testFalsy(val) || (Array.isArray(val) && testFalsy(val[0]) && testFalsy(val[1])),
   },
 )
 
@@ -287,7 +287,7 @@ export const filterFn_inNumberRange = Object.assign(
       return [min, max] as const
     },
     autoRemove: (val: any) =>
-      testFalsy(val) || (testFalsy(val[0]) && testFalsy(val[1])),
+      testFalsy(val) || (Array.isArray(val) && testFalsy(val[0]) && testFalsy(val[1])),
   },
 )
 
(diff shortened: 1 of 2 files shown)


KevinVandy:
wait, how is a number valid as an array?

KevinVandy:
There was a conflict in this one, but gave co-author credit to the merged PR

PR #6385, opened 2026-07-06
Feature: Allow Manual Override of Pagination canNextPage/canPreviousPage

rtablada:
## 🎯 Changes

This adds explicit `canNextPage` and `canPreviousPage` options to Pagination state.

In large data scale pagination (news feeds, financial, live updates, etc), counts may not be accurate to whether or not pages are available. In these cases it is common for [REST (JSON:API)](https://jsonapi.org/format/#fetching-pagination) or GraphQL requests to add information on how to navigate to next/previous pages and whether or not those pages exist.

## Alternative

For these cases consumers *could* use a separate store with fallback to Tanstack pagination atom/state. However this means having to manage two state objects and in cases of Frameworks where signal evaluation is lazy (Ember/Vue/Solid), it can be easy to return before all reactive state are evaluated and be in a situation where a value is not properly observed in the case of fallback behavior.

## ✅ Checklist

### DIFF
--- a/packages/table-core/src/features/row-pagination/rowPaginationFeature.utils.ts
+++ b/packages/table-core/src/features/row-pagination/rowPaginationFeature.utils.ts
@@ -251,7 +251,14 @@ export function table_getCanPreviousPage<
   TFeatures extends TableFeatures,
   TData extends RowData,
 >(table: Table_Internal<TFeatures, TData>) {
-  return (table.atoms.pagination?.get()?.pageIndex ?? 0) > 0
+  const pagination = table.atoms.pagination?.get()
+  const { canPreviousPage, pageIndex } = pagination ?? {}
+
+  if (canPreviousPage !== undefined) {
+    return canPreviousPage
+  }
+
+  return (pageIndex ?? defaultPageIndex) > 0
 }
 
 /**
@@ -269,9 +276,15 @@ export function table_getCanNextPage<
   TFeatures extends TableFeatures,
   TData extends RowData,
 >(table: Table_Internal<TFeatures, TData>) {
-  const pageIndex = table.atoms.pagination?.get()?.pageIndex ?? defaultPageIndex
-
+  const pagination = table.atoms.pagination?.get()
+  const pageIndex = pagination?.pageIndex ?? defaultPageIndex
+  // Read `pageCount` before any early return so the reactivity system always
+  // subscribes to its dependencies (pageSize atom + pre-paginated row model).
   const pageCount = table_getPageCount(table)
+
+  if (pagination?.canNextPage !== undefined) {
+    return pagination.canNextPage
+  }
 
   if (pageCount === -1) {
     return true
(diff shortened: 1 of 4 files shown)


packages/table-core/src/features/row-pagination/rowPaginationFeature.utils.ts
```diff
@@ -251,7 +251,14 @@ export function table_getCanPreviousPage<
   TFeatures extends TableFeatures,
   TData extends RowData,
 >(table: Table_Internal<TFeatures, TData>) {
-  return (table.atoms.pagination?.get()?.pageIndex ?? 0) > 0
```
rtablada, line 254: Previously this used `?? 0`. However this is now updated to use the same `defaultPageIndex` const as `table_getCanNextPage` for consistency

packages/table-core/src/features/row-pagination/rowPaginationFeature.utils.ts
```diff
@@ -251,7 +251,14 @@ export function table_getCanPreviousPage<
...
+  const pagination = table.atoms.pagination?.get()
+  const { canPreviousPage, pageIndex } = pagination ?? {}
```
rtablada, line 255: Destructuring before any return condition ensures reactivity for lazy eval frameworks are maintained.

This is important in the case where `pageIndex` is declared as a lazy getter in frameworks such as Solid/Angular/Ember.

packages/table-core/tests/unit/features/row-pagination/rowPaginationFeature.utils.test.ts
```diff
@@ -0,0 +1,123 @@
+import { describe, expect, it } from 'vitest'
```
rtablada: Net new unit tests.

rtablada:
In this, I updated only the React docs. I'm not sure if there is a sync script or existing prompt to sync similar API change docs across all FWs

KevinVandy:
I prefer to not have anything that can/should be "derived" in state. I can see the argument for a table option that acts like the `rowGetCanExpand` and similar table option APIs, but this should not be "stateful"

rtablada:
@KevinVandy the difference is that this isn't just a derived state. If total count is indeterminant, then the state can't be fully derived. This is different than `rowGetCanExpand` since external state dictates the capabilities.

rtablada:
For instance in the case of a stats or log it is common to have `hasNext`/`hasPrev` or `next`/`prev` request URL be passed in meta without.

In retail analytics this is particularly common when querying pages can cross partitions or DB page files so counts are very expensive or even not applicable.

--

Another case is an active log stream tail where `hasNext` could be sent as more rows are made available (think Graphana or similar log table where as a job log is running more data is parsed and made available to later pages). Again this is a case where usually counts are not generated because of the nature of the data.

rtablada:
In discussing some with our team some more background on why external state `canNextPage` and `canPreviousPage` are useful for server-side/manual pagination:

1. Data from a "count" meta data may be cached or stale compared to current page pointer/cursor
2. APIs may give count estimates for large data (ie `>100,000`) especially when full db page search is costly
3. Other user or integration will stale counts (similar to the log case, but 


---

## Alternative considered

Something that I assessed would be to not use tanstack for pagination state management at all or rolling a full separate pagination feature plugin.

The downside of this is losing type compatibility and plugin interoperability.

PR #6394, opened 2026-07-09
fix(table-core): guard Array.isArray before index access in range filter autoRemove

sanjibani:
Fixes #6353.

## Bug

`filterFn_between`, `filterFn_betweenInclusive`, and `filterFn_inNumberRange` share an `autoRemove`:

```ts
autoRemove: (val: any) =>
  testFalsy(val) || (testFalsy(val[0]) && testFalsy(val[1]))
```

When a scalar value such as `99` is passed instead of a `[min, max]` tuple, `val[0]` and `val[1]` are both `undefined`, so the condition evaluates to `true && true` and the filter is silently auto-removed.

Numeric columns default to `inNumberRange` via `column_getAutoFilterFn`, so `col.setColumnFilter(99)` on a numeric column discards the filter before the row scan even runs.

## Fix

Wrap the index checks in `Array.isArray(val) &&`:

```ts
autoRemove: (val: any) =>
  testFalsy(val) ||
  (Array.isArray(val) && testFalsy(val[0]) && testFalsy(val[1]))
```

The leading `testFalsy(val)` still handles `undefined`, `null`, `0`, and `''`. Behavior of the existing array-tuple cases is unchanged because `Array.isArray` is true and short-circuits to the same boolean.

## Tests

- New `filterFns.inNumberRange.autoRemove` describe block (the function had no direct unit tests previously).
- One regression test added to each of `between.autoRemove` and `betweenInclusive.autoRemove` covering `autoRemove(99)`.
- Existing array-tuple cases unchanged.

Local trace of all relevant cases:

| `val`              | before | after |
| ------------------ | ------ | ----- |
| `undefined`        | true   | true  |
| `[undefined, undefined]` | true | true |
| `[null, null]`     | true   | true  |
| `['', '']`         | true   | true  |
| `[5, 10]`          | false  | false |
| `[0, 10]`          | false  | false |
| `[undefined, 10]`  | false  | false |
| `99`               | true (BUG) | false |
| `''`               | true   | true  |

2 files changed, +49/-3.

## Summary by CodeRabbit

* **Bug Fixes**
  * Corrected auto-removal logic for range-based filters so non-array/scalar inputs are no longer treated as empty ranges.
  * Fixed range filter auto-removal behavior for between/between-inclusive and numeric range cases.
* **Tests**
  * Added regression tests to ensure scalar inputs don’t trigger auto-removal, and empty-string handling remains consistent.

### DIFF
--- a/packages/table-core/src/fns/filterFns.ts
+++ b/packages/table-core/src/fns/filterFns.ts
@@ -219,7 +219,8 @@ const filterFn_between = Object.assign(
       filterFn_lessThan(row, columnId, filterValues[1])),
   {
     autoRemove: (val: any) =>
-      testFalsy(val) || (testFalsy(val[0]) && testFalsy(val[1])),
+      testFalsy(val) ||
+      (Array.isArray(val) && testFalsy(val[0]) && testFalsy(val[1])),
   },
 )
 
@@ -243,7 +244,8 @@ const filterFn_betweenInclusive = Object.assign(
       filterFn_lessThanOrEqualTo(row, columnId, filterValues[1])),
   {
     autoRemove: (val: any) =>
-      testFalsy(val) || (testFalsy(val[0]) && testFalsy(val[1])),
+      testFalsy(val) ||
+      (Array.isArray(val) && testFalsy(val[0]) && testFalsy(val[1])),
   },
 )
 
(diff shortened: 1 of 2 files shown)


sanjibani:
@coderabbitai fixed in b098522. The test was bundling two cases into one assertion — the regression test for #6353 (non-empty scalar tuple confusion) and the unrelated empty-string falsy behavior. Split into two: `should NOT auto-remove a non-empty scalar number` (covers autoRemove(99) / autoRemove(0)) and `should auto-remove an empty-string scalar (matches existing testFalsy behavior)` (covers autoRemove('')). The empty-string case is intentional — `testFalsy` already short-circuits to true for `''`, which matches the existing comment in the fix: "The leading `testFalsy(val)` still handles `undefined`, `null`, `0`, and `''`." Vitest run: 73/73 passing.

PR #6438, opened 2026-07-18
fix global filtering with undefined first values

### LINKED ISSUE #4783: Global Filter Fails When First Row Value is Undefined
### Describe the bug

If the first row of a table has an undefined value in a column, the global filter won't work on that column, regardless of valid values in subsequent rows.

Take for example:

| ID    | Other |
|-------|-------|
|       | test  |
| hello | test  |
| world | test  |

Searching for hello fails with the global filter fails, even though it's clearly there. As soon as the order of the rows within the data is reversed, the global filters begins to work again.

https://codesandbox.io/p/sandbox/adoring-sid-ro5tmg?file=%2Fsrc%2Fmain.tsx&selection=%5B%7B%22endColumn%22%3A26%2C%22endLineNumber%22%3A56%2C%22startColumn%22%3A26%2C%22startLineNumber%22%3A56%7D%5D

### Steps to reproduce

Create a table where the first row of data has an undefined value. Subsequent rows should have string values. Upon using the global filter to search for an existing value, nothing is shown.

### Expected behavior

Global filter should work, regardless of row order.

Every time

### Screenshots or Videos

https://user-images.githubusercontent.com/91962855/228718000-77c0218e-7e71-4447-8994-41b8ccf247a0.mp4

### Platform

- OS: Windows
- Browser: Firefox

### react-table version

8.8.2

### TypeScript version

_No response_

### Additional context

This may be due to the getColumnCanGlobalFilter implementation: https://github.com/TanStack/table/blob/134414e7c5c6ff732c839a6c8b8f43a933967d3e/packages/table-core/src/features/Filters.ts#L190-L196

It seems to check the first row's values to determine if certain columns can be globally filtered. 

### Terms & Code of Conduct

KevinVandy:
## What changed

- determine default global-filter eligibility from the first non-nullish value in a column
- preserve the existing behavior of enabling string and number columns only
- add a regression test covering an undefined first value followed by searchable strings

## Why

The default eligibility predicate sampled only the first row. When that row had an undefined value, the entire column was excluded from global filtering even when later rows contained valid strings or numbers. This made results depend on row order.

## Validation

- reproduced the failure on beta before applying the implementation change
- table-core test suite: 52 files, 1,005 tests passed
- table-core TypeScript and declaration-emit checks passed
- table-core ESLint passed
- Prettier and git diff checks passed

Closes #4783

## Summary by CodeRabbit

* **Bug Fixes**
  * Global filtering now correctly evaluates columns when the first row has a blank value.
  * Filtering works as expected when matching values appear in later rows.
  * Columns explicitly enabled for global filtering can now be filtered even when they contain object values, using a custom filter.

* **Tests**
  * Added coverage for undefined values in the first row and explicitly enabled object-valued columns.


KevinVandy:
credit to @SAY-5 for PRing a fix in https://github.com/TanStack/table/pull/6252

PR #6443, opened 2026-07-18
fix(table-core): reset expansion when data changes

### LINKED ISSUE #5801: autoResetExpanded not reset the expanded state on data chagne in V8
### TanStack Table version

V8.20.5

### Framework/Library version

18.3.1

The autoResetExpanded can not reset the expanded state even if I set this flag to true.
The below example is a copy of the official document. I even changed the expanded state to uncontrol mode(the table maintains the state)
After expanding some rows and clicking the `refresh data` button, the expanded rows can not be reset.

Worked in V7:  https://codesandbox.io/p/sandbox/trusting-framework-qhl8rl?file=%2Fsrc%2FApp.js%3A37%2C10

Not work in V8: https://codesandbox.io/p/devbox/flamboyant-blackburn-2ctj8j?file=%2Fsrc%2Fmain.tsx%3A130%2C29&workspaceId=52d5059b-b49f-4b59-913f-c07fb465ad14

https://codesandbox.io/p/devbox/flamboyant-blackburn-2ctj8j?file=%2Fsrc%2Fmain.tsx%3A130%2C29&workspaceId=52d5059b-b49f-4b59-913f-c07fb465ad14

_No response_

None

### Terms & Code of Conduct

KevinVandy:
Closes #5801

## Decision

This remains a valid, narrowly scoped bug in the latest v9 beta. The existing automatic expansion reset is attached to the grouped row-model stage, so it happens only when grouping support is installed. A table using row expansion without grouping retains stale expanded row IDs after its data reference changes, despite the autoResetExpanded option contract.

This is not a new API or a behavior redesign. It connects the existing reset helper to the core data-dependent row model, alongside the existing automatic page-index reset.

## Changes

- run the existing expanded-state auto-reset after the core row model recomputes for a new data reference
- add an end-to-end regression test using expansion without the grouping feature
- retain the existing autoResetAll, autoResetExpanded, manualExpanding, initial-state, and controlled-state updater semantics

## Verification

- regression test fails on beta before the implementation change, retaining the expanded row ID
- table-core focused auto-reset suite: 22 tests passed
- full table-core suite: 52 files and 1,005 tests passed
- TypeScript source and declaration-emit checks passed
- ESLint passed for changed files
- Prettier check passed for changed files

### DIFF
--- a/packages/table-core/src/core/row-models/createCoreRowModel.ts
+++ b/packages/table-core/src/core/row-models/createCoreRowModel.ts
@@ -1,5 +1,6 @@
 import { constructRow } from '../rows/constructRow'
 import { makeObjectMap, tableMemo } from '../../utils'
+import { table_autoResetExpanded } from '../../features/row-expanding/rowExpandingFeature.utils'
 import { table_autoResetPageIndex } from '../../features/row-pagination/rowPaginationFeature.utils'
 import type { Table_Internal } from '../../types/Table'
 import type { RowModel } from './coreRowModelsFeature.types'
@@ -25,7 +26,10 @@ export function createCoreRowModel<
       fnName: 'table.getCoreRowModel',
       memoDeps: () => [table.options.data],
       fn: () => _createCoreRowModel(table, table.options.data),
-      onAfterUpdate: () => table_autoResetPageIndex(table),
+      onAfterUpdate: () => {
+        table_autoResetExpanded(table)
+        table_autoResetPageIndex(table)
+      },
     })
   }
 }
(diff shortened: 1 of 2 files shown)


packages/table-core/src/core/row-models/createCoreRowModel.ts
```diff
@@ -25,7 +26,10 @@ export function createCoreRowModel<
       fnName: 'table.getCoreRowModel',
       memoDeps: () => [table.options.data],
       fn: () => _createCoreRowModel(table, table.options.data),
-      onAfterUpdate: () => table_autoResetPageIndex(table),
+      onAfterUpdate: () => {
+        table_autoResetExpanded(table)
+        table_autoResetPageIndex(table)
```
riccardoperra, line 31: We could use `batch`from reactivity here I think, or move the batch invocation into onAfterUpdate 🤔

PR #6445, opened 2026-07-20
perf: compute NODE_ENV up-front & flip opt-in

43081j:
The code published to npm still contains the `NODE_ENV` guards. If you're running this code in an env where that isn't set, or where it is `production`, you pay the cost _per call_ of the string comparison.

We can avoid that by computing it up front since the env won't change.

Then we instead have `if (IS_DEV)` rather than
`if (process.env.NODE_ENV === ...)`.

Additionally, many of these guards were previously on `=== 'development'`. This meant only something like vite in dev mode would see useful errors (since most other tools won't set NODE_ENV at all). If we flip this around to `!== 'production'`, it means all dev runtimes would get the more useful error messages.

**DRAFT until we agree that its ok to flip the flag**

## ✅ Checklist

### DIFF
--- a/packages/table-core/src/core/columns/constructColumn.ts
+++ b/packages/table-core/src/core/columns/constructColumn.ts
@@ -10,6 +10,8 @@ import type {
 import type { Column } from '../../types/Column'
 import type { Column_CoreProperties } from './coreColumnsFeature.types'
 
+const IS_DEV = process.env.NODE_ENV !== 'production'
+
 /**
  * Creates or retrieves the column prototype for a table.
  * The prototype is cached on the table and shared by all column instances.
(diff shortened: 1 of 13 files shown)


KevinVandy:
@43081j Every time I've tried to add code like this in a centralized way in the past around `process.env?.NODE_ENV`, I thought that it ended up breaking tree-shaking. Is this really the correct way to do it?

43081j:
@KevinVandy im going to close this...

basically here's the situation today:

- a bundler like esbuild will tree-shake inline `NODE_ENV` conditions (like in `beta` today)
- anyone running table _without a bundler_ or _without tree-shaking_ pays a perf cost of evaluating these conditions in hot paths

given i assume most tanstack users are bundling, i think what's in `beta` is already fine 👍

PR #6446, opened 2026-07-20
perf(table-core): pre-compute instance init functions

43081j:
When we construct a row, we iterate over the table's features and initialise any which have an `initRowInstanceData`.

This is expensive since most don't have such a method.

We actually know when we construct the table which features have this method. This changes the table construction to store the initialisers up front so when we create a row, we can just iterate the already known initialisers.

## ✅ Checklist

## Summary by CodeRabbit

* **Performance**
  * Improved table instance creation by precomputing and reusing per-feature initialization callbacks across rows, cells, columns, headers, and header groups.
  * Instance setup now reuses cached initialization steps instead of re-walking feature hooks during each build.
* **New Features**
  * Added/extended per-instance lifecycle hooks so features can initialize mutable data for cells, columns, headers, and header groups.
* **Documentation**
  * Updated custom-feature guides and API references to describe the new initialization hooks and their lifecycle.


KevinVandy:
I like this change, and I will properly review this in a few days.

PR #6448, opened 2026-07-21
refactor(table-core): remove SyntheticEvent persist() calls

conao3:
## 🎯 Changes

Removes the `(e as any).persist?.()` calls from the sorting, expanding, resizing, and selection event handlers, plus the JSDoc / generated reference docs that described them.

`persist()` was a workaround for React's SyntheticEvent pooling, which was removed in React 17 (the method has been a deprecated no-op since then). v9 requires `react >= 18`, so these calls can't do anything anymore. Dropping them also removes 4 `as any` casts from table-core.

References:

- [React v17 RC announcement, "No Event Pooling"](https://legacy.reactjs.org/blog/2020/08/10/react-v17-rc.html#no-event-pooling): "Note that `e.persist()` is still available on the React event object, but now it doesn't do anything."
- [React event object in the react.dev reference](https://react.dev/reference/react-dom/components/common#react-event-object): "`persist()`: Not used with React DOM."

## ✅ Checklist

---
Assisted-by: Claude Code:claude-fable-5

### DIFF
--- a/docs/reference/static-functions/functions/column_getToggleSortingHandler.md
+++ b/docs/reference/static-functions/functions/column_getToggleSortingHandler.md
@@ -13,9 +13,8 @@ Defined in: [features/row-sorting/rowSortingFeature.utils.ts:494](https://github
 
 Creates a header event handler that toggles this column's sorting.
 
-The handler ignores events when the column cannot sort, persists React-style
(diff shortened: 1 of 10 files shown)


KevinVandy (approved):

conao3:
Thanks!

Issue #6450, opened 2026-07-21
"Cannot update a component while rendering a different component" while using controlled expanded state


volodymyrdn:
### TanStack Table version

9.0.0-beta.55

### Framework/Library version

React v19.2.6

### Describe the bug and the steps to reproduce it

Using the  official [Expanding example](https://tanstack.com/table/beta/docs/framework/react/examples/expanding?panel=code) and switching to controlled expanded state, I am receiving an error saying: **Cannot update a component (`App`) while rendering a different component (`App`)**. 

To switch to the controlled state, these steps were made: 
- `const [expanded, setExpanded] = useState<ExpandedState>({})` added to the App.
- `state: { expanded }` uncommented.
- `onExpandedChange: setExpanded` uncommented.

[Here](https://codesandbox.io/p/devbox/lingering-water-s9qsyv) is the link to the sandbox. 

Steps to reproduce the error: 
- try to expand any of rows. The error is seen in the console then.

Am I doing it wrong?

Thanks in advance.

### Your Minimal, Reproducible Example - (Sandbox Highly Recommended)

https://codesandbox.io/p/devbox/lingering-water-s9qsyv

### Do you intend to try to help solve this bug with your own PR?

No, because I do not know how

### Terms & Code of Conduct

- [x] I agree to follow this project's Code of Conduct
- [x] I understand that if my bug cannot be reliable reproduced in a debuggable environment, it will probably not be fixed and this issue may even be closed.

volodymyrdn:
**UPDATE**: this error occurs in all examples, where the controlled state is used. For example [this](https://tanstack.com/table/beta/docs/framework/react/examples/basic-external-state?panel=sandbox) and [that](https://tanstack.com/table/beta/docs/framework/react/examples/custom-plugin).

KevinVandy:
we think we finally fixed this in the latest v9 beta

PR #6458, opened 2026-07-27
fix: react adapter sync state

riccardoperra:
## Summary

This PR fixes React render-phase updates when using controlled table state, as reported in [#6450](https://github.com/TanStack/table/issues/6450).

The core idea is to separate two operations that previously happened together:

1. Reading the latest table options and controlled state during React render.
2. Publishing that state into the reactive atom graph after React commits.

Previously, `useTable` called `setOptions()` during render, and `setOptions()` immediately synchronized controlled state into writable base atoms. Updating those atoms synchronously notified `table.store`, which could call React’s external-store listener while another render was still in progress:

```text
React render
  -> table.setOptions()
  -> controlled state copied into base atom
  -> table.store synchronously notifies React
  -> React schedules an update during render
  -> "Cannot update a component while rendering a different component"
```

The new flow keeps render reads synchronous and current, but defers reactive publication until the React layout/commit phase.

## Goals

The implementation is intended to support all of the following cases:

- Controlled state passed through `options.state`.
- External atoms passed through `options.atoms`.
- Internal/uncontrolled table state.
- Per-slice ownership changes between controlled and internal state.
- External atoms without requiring an additional outer `useSelector`.
- Root-level state selectors without redundant React renders.
- Isolated `table.Subscribe` consumers.
- Stable `getSnapshot()` references.
- React Strict Mode.
- Rapid consecutive updates.
- Controlled objects recreated on every render.
- Suspended renders that must not publish uncommitted state.
- Backward compatibility for non-React adapters and existing static APIs.

## State ownership model

Each state slice now resolves using the documented precedence:

```text
options.atoms[key]
  > own options.state[key]
  > baseAtoms[key]
```

The check against `options.state` is an own-property check rather than a truthiness check. This distinguishes a missing controlled slice from a slice explicitly owned by the controlled state object.

```mermaid
flowchart TD
    A[Read table state slice] --> B{External atom configured?}

    B -- Yes --> C[Read and return external atom]
    B -- No --> D[Read base atom to retain reactive dependency]

    D --> E{options.state owns this key?}
    E -- Yes --> F[Return controlled render value]
    E -- No --> G[Return base atom value]

    C --> H[Stable live slice snapshot]
    F --> H
    G --> H

    H --> I[Shallow-stable aggregate table.store]
    I --> J[React selector gate]
```

The base or external atom is still read before resolving controlled state. This keeps the readonly atom connected to its reactive owner even while the controlled value currently wins.

### 1. Live readonly atoms in the React adapter

React table options are plain values updated during render. They are not themselves reactive dependencies, so a normal computed atom cannot know that `options.state` changed.

The React `createReadonlyAtom` implementation therefore returns a small live facade:

- `get()` evaluates the resolver against the latest render-time options.
- The result is cached using the configured comparator.
- Repeated reads return a stable reference when the result is semantically unchanged.
- A hidden computed atom owns the real subscription.
- The hidden computed tracks normal atom dependencies plus a React commit atom.

Conceptually:

```ts
const getSnapshot = () => {
  const next = computeFromCurrentOptions()

  if (!hasSnapshot || !compare(snapshot, next)) {
    snapshot = next
  }

  return snapshot
}

const reactiveAtom = createAtom(() => {
  commitAtom.get()
  return getSnapshot()
})
```

This gives us two complementary behaviors:

- Render-time getters always see the current controlled options.
- Subscribers are invalidated only by real reactive writes or by a React commit.

### 2. Render phase: update options without publishing atoms

`useTable` now updates the table options using:

```ts
table_setOptions(table, updater, {
  syncExternalState: false,
})
```

This keeps all options current during render but explicitly prevents `setOptions()` from synchronizing controlled state into base atoms at that point.

As a result, during the render that receives a new controlled value:

- `table.options.state` contains the new value.
- `table.atoms.<slice>.get()` returns the new value.
- `table.store.get()` returns the new value.
- `table.state` returns the selected new value.
- Row models can use the new value.
- No writable atom is updated.
- No store subscriber is notified during render.

### 3. Commit phase: publish the captured state

After React commits, an isomorphic layout effect publishes the controlled state captured by that render:

```ts
reactivity.batch(() => {
  table_syncExternalStateToBaseAtoms(
    table,
    capturedControlledState,
    shallow,
  )

  reactivity.commit()
})
```

Using the captured state is important. The synchronization must represent the render that actually committed, rather than rereading a potentially newer mutable `table.options` object.

The operation is batched so that:

- Controlled slices are mirrored into base atoms.
- Semantically equal slices do not cause redundant writes.
- The React commit atom invalidates readonly atoms.
- Subscribers observe one coherent snapshot.
- Isolated subscribers can update before paint.

The explicit commit invalidation also handles option-only changes where no base atom write occurs, such as releasing ownership of a controlled slice.

### 4. Root selector filtering

The public `table.store` must continue notifying its subscribers after a commit. This is required for consumers such as:

```tsx
<table.Subscribe selector={(state) => state.pagination.pageIndex}>
  {(pageIndex) => <span>{pageIndex}</span>}
</table.Subscribe>
```

However, the root `useTable` component has already rendered the new controlled value. Forwarding the same publication to React’s root external-store listener would be redundant.

`useTableSelector` wraps the existing TanStack Store `useSelector`; it does not reimplement `useSyncExternalStoreWithSelector`.

It records the selector and selected snapshot that React actually committed. When the public store later publishes:

```ts
if (!compare(committedSelection, nextSelection)) {
  onStoreChange(nextState)
}
```

Therefore:

- Public store subscribers still receive the update.
- Isolated `table.Subscribe` components can update.
- The root React listener is skipped when the committed selection already matches.
- Controlled updates produce only the render caused by the controlling React state.
- Unrelated store changes do not call the root listener when the selected result is unchanged.

Hook ordering is intentional:

1. `useTableSelector` records the committed root selection in its layout effect.
2. `useTable` publishes the atom graph in its following layout effect.
3. The filtered root subscription can compare against the correct committed selection.

### 5. Stable aggregate snapshots

The aggregate `table.store` readonly atom uses shallow comparison.

This is necessary because building the full table state creates a new object. Without snapshot caching, React could observe a different object reference on every `getSnapshot()` call and report:

```text
The result of getSnapshot should be cached to avoid an infinite loop
```

The store snapshot changes only when at least one resolved slice changes shallowly.

## Controlled update sequence

The following sequence uses:

- `C0`: previously committed controlled state.
- `C1`: next controlled state received by React.
- `B0`: previous base atom value.

```mermaid
sequenceDiagram
    autonumber

    participant E as Table event
    participant R as React owner
    participant U as useTable
    participant O as table.options
    participant S as live table.store
    participant X as root useTableSelector
    participant B as base atoms
    participant C as commit atom
    participant I as isolated Subscribe

    E->>R: onPaginationChange(updater)
    R->>U: render with controlled C1

    U->>O: table_setOptions(syncExternalState: false)
    Note over U,O: Options change during render<br/>No writable atom update

    U->>S: get current snapshot
    S->>B: read B0 to retain dependency
    Note over S: Resolve external atom<br/>then controlled C1<br/>then base B0
    S-->>X: stable snapshot containing C1
    X-->>U: selected C1

    U-->>R: table.state, atoms and row models use C1
    R->>R: commit UI for C1

    Note over X,U: Layout phase after commit

    X->>X: record committed selection C1

    U->>B: mirror captured C1, shallow-gated
    U->>C: increment commit version
    Note over U,C: Batched publication

    B-->>S: invalidate reactive dependencies
    C-->>S: invalidate option-dependent readonly atoms

    S->>I: publish committed snapshot C1
    S->>X: publish committed snapshot C1

    X->>X: compare next C1 with committed C1
    Note over X: Equal selection<br/>skip root onStoreChange

    I->>I: rerender only if its selector changed
```

## Supported scenarios

| Scenario                                              | State source         | Expected behavior                                                                                                                                                 |
| ----------------------------------------------------- | -------------------- | ----------------------------------------------------------------------------------------------------------------------------------------------------------------- |
| Internal/uncontrolled state                           | `baseAtoms[key]`     | Table APIs update the base atom and notify subscribers immediately. The root rerenders only if its selected value changes.                                        |
| Controlled state                                      | `options.state[key]` | The current render reads the controlled value immediately. Base synchronization and subscriber notification happen only after commit.                             |
| External atom                                         | `options.atoms[key]` | The table subscribes internally to the atom. No outer React `useSelector` is required.                                                                            |
| External atom and controlled state for the same slice | External atom        | The external atom wins. Controlled state does not override the slice resolver.                                                                                    |
| Controlled → internal ownership                       | Base atom            | After ownership is released, the current base value becomes visible. Commit invalidation wakes memoized or isolated subscribers even if no new base write occurs. |
| Internal → controlled ownership                       | Controlled state     | The controlled value is visible in the same React render without waiting for atom synchronization.                                                                |
| Unselected slice changes                              | Any source           | The public store may update, but the root selector does not rerender or forward `onStoreChange` when its selected result is unchanged.                            |
| Recreated controlled objects                          | Controlled state     | Shallow comparison suppresses equivalent base writes and prevents maximum-depth loops.                                                                            |
| Rapid controlled updates                              | Controlled state     | React resolves the final controlled value; committed publication brings isolated subscribers to the same final snapshot.                                          |
| Strict Mode                                           | Controlled state     | Repeated render/effect execution remains idempotent and comparator-gated.                                                                                         |
| Suspended render                                      | Controlled state     | Render-time reads can evaluate the WIP value, but base atoms are not mutated and store subscribers are not notified because no commit effect runs.                |

### `table_setOptions`

The function accepts an optional final configuration argument:

```ts
table_setOptions(table, updater, {
  syncExternalState: false,
})
```

The default remains unchanged:

```ts
table_setOptions(table, updater)
```

continues to synchronize external state immediately. This preserves existing behavior for core and other adapters.

### `table_syncExternalStateToBaseAtoms`

The function can now receive:

- An explicit state snapshot.
- An optional comparator.

```ts
table_syncExternalStateToBaseAtoms(
  table,
  capturedState,
  shallow,
)
```

Passing an explicit snapshot allows React to publish the state associated with a specific committed render.

The implementation distinguishes:

```ts
table_syncExternalStateToBaseAtoms(table)
```

from:

```ts
table_syncExternalStateToBaseAtoms(table, undefined)
```

The first form reads `table.options.state`. The second intentionally publishes no controlled state. This prevents an explicitly captured `undefined` value from accidentally falling back to a newer mutable options object.

### Controlled React state

```tsx
function App() {
  const [pagination, setPagination] = React.useState({
    pageIndex: 0,
    pageSize: 10,
  })

  const table = useTable(
    {
      data,
      columns,
      features,
      state: {
        pagination,
      },
      onPaginationChange: setPagination,
    },
    (state) => state.pagination,
  )

  return (
    <>
      <div>Page {table.state.pageIndex + 1}</div>
      <button onClick={() => table.nextPage()}>Next</button>
    </>
  )
}
```

The first and subsequent pagination updates now:

- Render the correct controlled state.
- Produce the correct paginated row model.
- Do not synchronously update the atom graph during render.
- Do not produce the React render-phase warning.
- Do not cause a redundant root render after commit.

### External atom without an outer subscription

```tsx
function App() {
  const paginationAtom = useCreateAtom({
    pageIndex: 0,
    pageSize: 10,
  })

  const table = useTable(
    {
      data,
      columns,
      features,
      atoms: {
        pagination: paginationAtom,
      },
    },
    (state) => state.pagination,
  )

  return (
    <>
      <div>Page {table.state.pageIndex + 1}</div>
      <button onClick={() => table.nextPage()}>Next</button>
    </>
  )
}
```

No additional subscription is required:

```tsx
// Not needed
useSelector(paginationAtom)
```

The hidden readonly atom tracks the external atom, and `useTable` reacts through `table.store`.

## Why not `queueMicrotask`, `setTimeout`, or debouncing?

A timer-based solution is not coupled to the React lifecycle.

Scheduling synchronization during render could cause a callback to run even if that render:

- Suspends.
- Is abandoned.
- Is replaced by a newer render.
- Never commits.

Timers also introduce ordering problems for rapid updates and can leave isolated subscribers stale for an arbitrary amount of time.

A passive effect would be tied to a real commit, but publication could happen after paint. A layout effect provides the required semantics:

- It runs only for a committed render.
- The committed root selection is recorded first.
- The atom graph is updated before paint.
- Root notifications can be filtered.
- Isolated subscribers can observe the committed state without relying on an arbitrary delay.

## Test coverage

The focused regression coverage includes:

| Area                  | Coverage                                                                                    |
| --------------------- | ------------------------------------------------------------------------------------------- |
| Controlled state      | Live store reads, stable root subscription and no redundant render                          |
| Ownership transitions | Controlled/internal transitions, including isolated-subscriber invalidation                 |
| External atoms        | Atom updates without an outer `useSelector`, including row-model changes                    |
| Precedence            | External atom > controlled state > internal base atom                                       |
| Warning regression    | No render-phase warning after the first or second pagination update                         |
| State consistency     | React state, `table.state`, slice atom, flat store, subscriber and row model remain aligned |
| Selector gating       | Unselected internal and external changes do not rerender the root                           |
| Snapshot stability    | Stable `store.get()`, fresh selector objects and recreated controlled slices                |
| Rapid updates         | Memoized isolated subscribers receive the final controlled value                            |
| Commit safety         | Suspended renders do not mutate base state or notify subscribers                            |
| Strict Mode           | Controlled updates remain stable under repeated rendering/effect execution                  |
| Browser integration   | Controlled-state and external-atoms examples exercise pagination in Chromium                |

Verified focused results:

- React adapter tests: **11/11**
- Core atom tests: **11/11**
- Controlled-state Chromium E2E: **3/3**
- External-atoms Chromium E2E: **3/3**
- Focused total: **28/28**

Additional validation performed:

- Full table-core suite: **1005/1005**
- Preact adapter suite: **1/1**
- React/core/preact typechecks: passed
- React/core/preact builds: passed
- Controlled-state and external-atoms example builds: passed
- ESLint, Prettier and diff checks: passed

## Known limitation / non-goal

This change makes publication into the reactive atom graph commit-only, but it does not make the entire table instance render-local.

`table.options` is still a shared mutable object updated during render. Therefore, while a concurrent render is suspended or later abandoned, imperative reads such as:

```ts
table.store.get()
table.atoms.pagination.get()
table.getRowModel()
```

may observe render-in-progress options.

What is guaranteed by this PR is narrower:

- A suspended render does not mutate the committed base atoms.
- A suspended render does not notify store subscribers.
- Only a committed render publishes controlled state into the reactive graph.

Providing full concurrent isolation for imperative table reads would require a larger architectural change, such as separating render-local options from committed options or returning a render-specific table facade.

External atom identities are also expected to remain construction-stable, consistent with the current options merging behavior.

---

This pull request introduces significant improvements to the React adapter for TanStack Table, focusing on more robust and predictable synchronization of controlled state, better integration with React's commit phase, and enhanced test coverage. The main changes ensure that externally controlled state is only published to the table's atom graph after React commits, preventing render-phase errors and improving fine-grained reactivity. Additionally, new tests are added to verify correct behavior, and dependency updates ensure compatibility with React 19.

**React Adapter and State Synchronization Enhancements:**

- `packages/react-table/src/reactivity.ts`, `packages/react-table/src/useTable.ts`, `packages/table-core/src/core/table/coreTablesFeature.utils.ts`, `packages/table-core/src/core/table/constructTable.ts`: Refactored the adapter to synchronize controlled state into the table's atom graph only after React commits, using a new `commit` method and a custom `useTableSelector` hook. This prevents render-phase errors and ensures that derived atoms and table APIs always read a consistent snapshot. The synchronization now supports an explicit comparator and is controlled via a new `syncExternalState` option. [\[1\]](diffhunk://#diff-4d91a001378663690e5682cf834160f04d8ab4b987438b4148fba8bb2727ae75R2-R29) [\[2\]](diffhunk://#diff-4d91a001378663690e5682cf834160f04d8ab4b987438b4148fba8bb2727ae75L31-R148) [\[3\]](diffhunk://#diff-519b4112fb6066ab95bc79dcf56d75db79545216e5deacd6dac8f55bd4a2e734L3-R10) [\[4\]](diffhunk://#diff-519b4112fb6066ab95bc79dcf56d75db79545216e5deacd6dac8f55bd4a2e734R25-R27) [\[5\]](diffhunk://#diff-519b4112fb6066ab95bc79dcf56d75db79545216e5deacd6dac8f55bd4a2e734L149-R165) [\[6\]](diffhunk://#diff-519b4112fb6066ab95bc79dcf56d75db79545216e5deacd6dac8f55bd4a2e734L172-R214) [\[7\]](diffhunk://#diff-cb2db581631155494b97cd54b2305037c8f42ebd5b129162e55d7a5d734d126eR1-R3) [\[8\]](diffhunk://#diff-cb2db581631155494b97cd54b2305037c8f42ebd5b129162e55d7a5d734d126eL136-R160) [\[9\]](diffhunk://#diff-cb2db581631155494b97cd54b2305037c8f42ebd5b129162e55d7a5d734d126eL165-R181) [\[10\]](diffhunk://#diff-05ae599c249ece6c138022a899cada7bc096ceab85008afa0e7820dd31375068R6-R15) [\[11\]](diffhunk://#diff-05ae599c249ece6c138022a899cada7bc096ceab85008afa0e7820dd31375068L21-R48) [\[12\]](diffhunk://#diff-05ae599c249ece6c138022a899cada7bc096ceab85008afa0e7820dd31375068L35-R62) [\[13\]](diffhunk://#diff-05ae599c249ece6c138022a899cada7bc096ceab85008afa0e7820dd31375068R177) [\[14\]](diffhunk://#diff-05ae599c249ece6c138022a899cada7bc096ceab85008afa0e7820dd31375068R186-R188)

**Test Coverage Improvements:**

- `examples/react/basic-external-atoms/tests/e2e/smoke.spec.ts`, `examples/react/basic-external-state/tests/e2e/smoke.spec.ts`: Added new end-to-end tests to verify that table updates correctly reflect changes from external atoms without requiring outer React subscriptions and that controlled pagination updates do not cause render-phase errors. [\[1\]](diffhunk://#diff-b1fbeaa3789cd623a221ee28c9c78cd5db8b1076a07d9545f5dfaea6e0d3ec25R87-R113) [\[2\]](diffhunk://#diff-3a91c4f6b1ba9e99d11c19edae9ba4064f733dce07b7c7bdc585ed3f8aed5114R87-R123)
- `packages/react-table/tests/test-setup.ts`, `packages/react-table/vite.config.ts`: Ensured the test environment is properly set up for React 19 by enabling the act environment and including the setup file in Vite config. [\[1\]](diffhunk://#diff-2f48dcfe3540f6655ae5c141f4177c33b53d0000f6e20477db6822f4c5aae1d1R1-R3) [\[2\]](diffhunk://#diff-6a2e294785403334ffe14e4f97ec3442c05f8837e93c36c7847294e45e843357R13)

**Dependency Updates:**

- [`packages/react-table/package.json`](diffhunk://#diff-0d204b354affc6f538e06ae4d76eb79672f32b03064a4468801e1313453c52dfR58-R63): Added `react-dom` and `@types/react-dom` as dev dependencies to support React 19 features and testing.

**Example and Usage Updates:**

- [`examples/react/basic-external-atoms/src/main.tsx`](diffhunk://#diff-9d3ebf2db439ecd6ca399fb6ab82e804669b6bc060812c0bf3a2b92dc9d6d90aL5-R5): Updated the example to remove unnecessary `useSelector` calls, relying on the improved React adapter for atom subscriptions and state access. [\[1\]](diffhunk://#diff-9d3ebf2db439ecd6ca399fb6ab82e804669b6bc060812c0bf3a2b92dc9d6d90aL5-R5) [\[2\]](diffhunk://#diff-9d3ebf2db439ecd6ca399fb6ab82e804669b6bc060812c0bf3a2b92dc9d6d90aL82-R83) [\[3\]](diffhunk://#diff-9d3ebf2db439ecd6ca399fb6ab82e804669b6bc060812c0bf3a2b92dc9d6d90aL188-R182) [\[4\]](diffhunk://#diff-9d3ebf2db439ecd6ca399fb6ab82e804669b6bc060812c0bf3a2b92dc9d6d90aL198-R192) [\[5\]](diffhunk://#diff-9d3ebf2db439ecd6ca399fb6ab82e804669b6bc060812c0bf3a2b92dc9d6d90aL207-R201)

## Summary by CodeRabbit

* **New Features**
  * Added selector-based subscription support for React state, plus an isomorphic layout effect helper.
  * Introduced render-phase reactivity to defer controlled-state publishing until commit, improving adapter consistency.
* **Bug Fixes**
  * Improved controlled vs external state synchronization timing to reduce render/commit update warnings.
  * Refined pagination and row-number rendering to stay in sync across pages and controls.
  * Prevented unnecessary auto-resets when grouping inputs or column-definition references change.
* **Tests**
  * Added Playwright smoke tests and expanded React/unit coverage for pagination, commit gating, and error-free updates.
* **Documentation**
  * Clarified `initTableInstanceData` feature initialization order across framework guides.


packages/react-table/src/reactivity.ts
```diff
@@ -28,14 +44,106 @@ export function reactReactivity(): TableReactivityBindings {
...
 }
+
+/**
+ * Selects the root table state while filtering notifications that merely
+ * publish a snapshot already read during the current React render.
+ *
+ * Isolated `table.Subscribe` consumers still receive the underlying store
+ * notification; only this hook's root subscription is filtered.
+ */
+export function useTableSelector<TState, TSelected>(
```
riccardoperra: fyi this is just a useSyncWithExternalStoreWithSelector revisited implementation

PR #6472, opened 2026-07-31
fix: align grouped sticky headers

### LINKED ISSUE #5397: there is a bug when you try sticky pin column group
### TanStack Table version

8.13.2

### Framework/Library version

18.2.0

open https://tanstack.com/table/v8/docs/framework/react/examples/column-pinning-sticky example and group firstname and lastname

```
const columnHelper = createColumnHelper<Person>();

const defaultColumns: ColumnDef<Person>[] = [
  columnHelper.group({
    id: 'info',
    header: () => <span>Info</span>,
    columns: [
      {
        accessorKey: 'firstName',
        id: 'firstName',
        header: 'First Name',
        cell: (info) => info.getValue(),
        footer: (props) => props.column.id,
        size: 180,
      },
      {
        accessorFn: (row) => row.lastName,
        id: 'lastName',
        cell: (info) => info.getValue(),
        header: () => <span>Last Name</span>,
        footer: (props) => props.column.id,
        size: 180,
      },
    ],
  }),
  {
    accessorKey: 'age',
    id: 'age',
    header: 'Age',
    footer: (props) => props.column.id,
    size: 180,
  },
  {
    accessorKey: 'visits',
    id: 'visits',
    header: 'Visits',
    footer: (props) => props.column.id,
    size: 180,
  },
  {
    accessorKey: 'status',
    id: 'status',
    header: 'Status',
    footer: (props) => props.column.id,
    size: 180,
  },
  {
    accessorKey: 'progress',
    id: 'progress',
    header: 'Profile Progress',
    footer: (props) => props.column.id,
    size: 180,
  },
];
```

then try pin "info" group info to the left (or right).

https://stackblitz.com/edit/tanstack-table-n7umup?file=src%2Fmain.tsx

image

None

### Terms & Code of Conduct

biubiukam:
## 🎯 Changes

Fix grouped sticky headers in the React column pinning example.

Grouped headers now derive their sticky side, offsets, width, and boundary shadow from the leaf columns rendered by the current header segment. This keeps mixed partial-pinning segments relative and aligns fully pinned groups when other pinned columns precede or follow them.

Add Playwright coverage for partial group pinning and non-zero start/end offsets.

Fixes #5397

## ✅ Checklist

## Summary by CodeRabbit

* **New Features**
  * Added support for grouped headers in the column-pinning example.
  * Grouped headers now correctly reflect pinning when all visible child columns are pinned to the same side.
  * Added sticky positioning and offset handling for grouped headers pinned to the start or end.

* **Tests**
  * Added end-to-end coverage for grouped-header pinning, alignment, dimensions, and sticky positioning.


MILLERMARRU (commented):
This example previously had no grouped columns at all, firstName and lastName were flat top-level accessors, so there was actually no way to exercise sticky positioning for a group header before this PR. Wrapping them in an `info` group is what exposes the bug in the first place, `getCommonPinningStyles(column)` was being called with the group's own `column`, whose `getIsPinned`/`getStart`/`getAfter` don't reflect what its leaf columns are actually doing.

I checked `header_getSize` in table-core to confirm `header.getSize()` on a group header already sums its leaf columns (`sumHeaderSize`), so using it directly as `width` for the group is correct and explains why the tests expect the Info header to be ~360px (180+180) rather than the single-column 180.

The "every leaf column must share the same pinned side" check in `getHeaderPinningStyles` is the right call for deciding sticky vs relative. The first test nails exactly why: pin only First Name to start while Last Name stays unpinned, and since pinning to `start` doesn't reorder columns when the pinned one is already first, TanStack still renders Info as one merged header cell, just now with mixed-pinned leaves underneath. Without the "all leaves agree" check, that merged cell would likely have inherited a stale or partial sticky style. The test asserting `position: relative` in that exact intermediate state is a good regression guard, and the second test showing Info splits into two separate header cells once First Name is pinned to `end` (because pinning to end physically reorders it away from Last Name) is a nice demonstration that the split itself is TanStack's own header-group logic doing its job, not something this PR has to handle manually.

Using `firstLeafColumn.getStart('start')` and `lastLeafColumn.getAfter('end')` instead of the group column's own offsets is correct since a group column has no size/position of its own outside of what its leaves contribute, and the boundingBox assertions comparing the sticky header's `x` against the actual leaf header's `x` are a good way to catch an off-by-one-column offset that a snapshot-only test would miss.

Only thing I'd note is this all lives in an example, so it's more "here's the pattern to copy" than covered by table-core's own test suite, but for a docs example that's the right place for this logic to live since it's inherently presentational.

biubiukam:
> This example previously had no grouped columns at all, firstName and lastName were flat top-level accessors, so there was actually no way to exercise sticky positioning for a group header before this PR. Wrapping them in an `info` group is what exposes the bug in the first place, `getCommonPinningStyles(column)` was being called with the group's own `column`, whose `getIsPinned`/`getStart`/`getAfter` don't reflect what its leaf columns are actually doing.
> 
> I checked `header_getSize` in table-core to confirm `header.getSize()` on a group header already sums its leaf columns (`sumHeaderSize`), so using it directly as `width` for the group is correct and explains why the tests expect the Info header to be ~360px (180+180) rather than the single-column 180.
> 
> The "every leaf column must share the same pinned side" check in `getHeaderPinningStyles` is the right call for deciding sticky vs relative. The first test nails exactly why: pin only First Name to start while Last Name stays unpinned, and since pinning to `start` doesn't reorder columns when the pinned one is already first, TanStack still renders Info as one merged header cell, just now with mixed-pinned leaves underneath. Without the "all leaves agree" check, that merged cell would likely have inherited a stale or partial sticky style. The test asserting `position: relative` in that exact intermediate state is a good regression guard, and the second test showing Info splits into two separate header cells once First Name is pinned to `end` (because pinning to end physically reorders it away from Last Name) is a nice demonstration that the split itself is TanStack's own header-group logic doing its job, not something this PR has to handle manually.
> 
> Using `firstLeafColumn.getStart('start')` and `lastLeafColumn.getAfter('end')` instead of the group column's own offsets is correct since a group column has no size/position of its own outside of what its leaves contribute, and the boundingBox assertions comparing the sticky header's `x` against the actual leaf header's `x` are a good way to catch an off-by-one-column offset that a snapshot-only test would miss.
> 
> Only thing I'd note is this all lives in an example, so it's more "here's the pattern to copy" than covered by table-core's own test suite, but for a docs example that's the right place for this logic to live since it's inherently presentational.

The grouped-header pinning logic intentionally stays in the React example because it is presentation-specific. No code changes are required from this review.

PR #6506, opened 2026-08-03
feat: implement and export `getAriaSort`

mstruebing:
This function was placed in a lot of examples and we just introduced it into our codebase. I think this would be value to be placed in the library as this needs to be used anyway all the time if you want to have an accesible table.

## 🎯 Changes

The commit message already tells the story, we needed to add this function to our codebase and while reviewing the PR I've thought this function should be exposed by the library in order to create easily accessible tables.

## ✅ Checklist

## Summary by CodeRabbit

* **Accessibility**
  * Improved table sorting announcements for assistive technologies by providing valid `aria-sort` values for ascending, descending, and unsorted columns.
* **Consistency**
  * Standardized sorting accessibility behavior across table examples and UI integrations.
* **Tests**
  * Added coverage for all supported sorting states.

### DIFF
--- a/examples/react/kitchen-sink-hero-ui/src/main.tsx
+++ b/examples/react/kitchen-sink-hero-ui/src/main.tsx
@@ -18,6 +18,7 @@ import {
   tableDevtoolsPlugin,
   useTanStackTableDevtools,
 } from '@tanstack/react-table-devtools'
+import { getAriaSort } from '@tanstack/react-table'
 import type { Column } from '@tanstack/react-table'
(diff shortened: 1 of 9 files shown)


mstruebing:
hey @KevinVandy can you share why you just closed this?
Like, why shouldn't this useful function in this project?

KevinVandy:
If we were to add something like this, it would be under a more comprehensive accessibility plugin. Not in core. At this point, it's best to leave this to the component libraries and design system, I think.

PR #6518, opened 2026-08-05
docs: correct the pageCount -1 note in the pagination guides

dfedoryshchev:
## Changes

In the manual-pagination guide, the Page Count and Row Count section says that with `pageCount: -1` both `getCanNextPage` and `getCanPreviousPage` "will always return `true`". That holds for `getCanNextPage` only. `table_getCanPreviousPage` never reads `pageCount`: in `rowPaginationFeature.utils.ts` it is `return (table.atoms.pagination?.get()?.pageIndex ?? 0) > 0`, so it is still `false` on page 0, and a reader following the guide would leave a "Previous" button enabled on the first page.

Both JSDoc blocks in that file already describe the real behaviour, and the unit tests cover the `-1` case for `getCanNextPage` only, so this is the guide prose having drifted rather than a code bug.

Corrected in all 10 framework copies of the guide, which carry the sentence verbatim. Prose only.

## Release Impact

- [ ] This change affects published code, and I have generated a [changeset](https://github.com/changesets/changesets/blob/main/docs/adding-a-changeset.md).
- [x] This change is docs/CI/dev-only (no release).

## Summary by CodeRabbit

* **Documentation**
  * Clarified pagination behavior across framework guides.
  * Documented that previous-page availability depends only on whether the current page index is greater than zero.
  * Clarified that when the total page count is unknown (`pageCount: -1`), next-page navigation remains enabled.
  * Updated server-side and manual pagination guidance for consistent behavior across supported frameworks.

### DIFF
--- a/docs/framework/alpine/guide/pagination.md
+++ b/docs/framework/alpine/guide/pagination.md
@@ -84,7 +84,7 @@ No pagination row model is needed for server-side pagination, but if you have pr
 
 #### Page Count and Row Count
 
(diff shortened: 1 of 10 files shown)


KevinVandy:
This ended up getting addressed in #6526 separately

Issue #6530, opened 2026-08-08
injectTable eagerly initializes options on a microtask, throwing NG0950 when the initializer reads a required signal input


michael-dg:
## Summary

`injectTable()` force-initializes its options via `queueMicrotask`, independently of whether the returned proxy is ever accessed. When the options initializer reads an Angular `input.required()`, that microtask can run before Angular has applied the binding, throwing `NG0950: Input "data" is required but no value is available yet` as an **uncaught exception**. A secondary issue compounds it: the `isMount` guard in the options-update effect swallows the first update, so a table that *does* construct before its inputs are bound keeps those pre-binding options.

## Description

`injectTable` wraps its initializer in `lazyInit`, defined in `packages/angular-table/src/lazySignalInitializer.ts`:

```ts
export function lazyInit<T extends object>(initializer: () => T): T {
  let object: T | null = null

  const initializeObject = () => {
    if (!object) {
      object = untracked(() => initializer())
    }
  }

  queueMicrotask(() => initializeObject())   // <-- eager, not lazy

  const table = () => {}

  return new Proxy<T>(table as T, {
    apply(...) { initializeObject(); /* ... */ },
    get(_, prop, receiver) { initializeObject(); return Reflect.get(object as T, prop, receiver) },
    has(_, prop) { initializeObject(); return Reflect.has(object as T, prop) },
    ownKeys() { initializeObject(); return Reflect.ownKeys(object as T) },
    // ...
  })
}
```

The `Proxy` traps make access-triggered initialization lazy, which is correct. The `queueMicrotask` line defeats that: initialization is attempted unconditionally on the next microtask, whether or not anything touched the table.

That is fine when the component's inputs are already bound. It is not fine for a component whose options initializer reads a required input:

```ts
export class DataTableComponent<T extends RowData = any> {
  data = input.required<T[]>();

  table = injectTable<typeof features, T>(() => ({
    features,
    data: this.data(),          // throws NG0950 if the microtask wins the race
    columns: this.tanstackColumns(),
  }));
}
```

Angular splits this across two phases: a template's **creation** pass instantiates child components (leaving their inputs unset), and the following **update** pass applies the bindings. In an application both phases run inside the same synchronous change-detection cycle, so the microtask drains after the inputs are bound and nothing goes wrong. It breaks whenever an `await` separates the two — which is the default shape of an Angular `TestBed` suite, where `TestBed.createComponent` performs the creation pass and the first `detectChanges()` performs the update pass:

```ts
[...]

PR #6534, opened 2026-08-09
fix(angular): ensure table options updates are not missed during first mount window

riccardoperra:
This pull request addresses a key issue in the `@tanstack/angular-table` package where options updates could be missed during the initial mount of a table. The changes ensure that option updates are reliably captured from the very first render, and also refactor the lazy initialization logic to provide more robust state tracking and improved test coverage.

**Core improvements to options reactivity and initialization:**

* The `injectTable` function now uses a computed signal for options and ensures updates are not missed during the initial mount, fixing a bug where the first update could be dropped. This involves tracking previous options and applying updates only when they change.
* The `lazyInit` utility is refactored to return an object with `value`, `rawValue`, and `initialized` properties, allowing for more precise control and introspection of the lazy initialization state. [[1]](diffhunk://#diff-024c4114c19c176041be2a5e2b978534bd3890d278ccd74f7e548a7ede2bd8d7L1-R10) [[2]](diffhunk://#diff-024c4114c19c176041be2a5e2b978534bd3890d278ccd74f7e548a7ede2bd8d7L16-R25) [[3]](diffhunk://#diff-024c4114c19c176041be2a5e2b978534bd3890d278ccd74f7e548a7ede2bd8d7R52-R61)

**Testing and reliability improvements:**

* Additional and updated tests ensure that options updates are not dropped before the effect first runs, and that the new lazy initialization logic behaves correctly and is testable. [[1]](diffhunk://#diff-f0e5b4a46fb3cfb1612f5ce75e3f2d55b86c75d1e64fb79837d1f80f29ba159bR150-R174) [[2]](diffhunk://#diff-34ed57b020d139a2b9addc1cbb68a15f3ef111c4873c779c2f0f8e6199366088L16-R37) [[3]](diffhunk://#diff-34ed57b020d139a2b9addc1cbb68a15f3ef111c4873c779c2f0f8e6199366088L46-R50) [[4]](diffhunk://#diff-34ed57b020d139a2b9addc1cbb68a15f3ef111c4873c779c2f0f8e6199366088L66-R77) [[5]](diffhunk://#diff-34ed57b020d139a2b9addc1cbb68a15f3ef111c4873c779c2f0f8e6199366088L105-R114)
* Minor test and type fixes, such as correcting feature spreading and updating test assertions to match the new initialization API.

**Documentation:**

* A changeset is added to document the patch release and the fix for missed options updates on first mount.

## Summary by CodeRabbit

- **Bug Fixes**
  - Fixed Angular Table options updates during initial rendering so the latest data and configuration are preserved.
  - Improved reactive updates to apply only when table options change.
  - Ensured table cleanup is safe when initialization has not completed.

- **Behavior Updates**
  - Lazy table initialization now reports its status and provides access to the initialized value.
  - Required signal inputs can trigger initialization during rendering.

- **Release**
  - Prepared a patch release for `@tanstack/angular-table`.

### DIFF
--- a/packages/angular-table/src/injectTable.ts
+++ b/packages/angular-table/src/injectTable.ts
@@ -97,44 +97,46 @@ export function injectTable<
   assertInInjectionContext(injectTable)
   const injector = inject(Injector)
   const ngZone = inject(NgZone)
+  const destroyRef = inject(DestroyRef)
   const options = computed(() => _options())
+  const coreReactivityFeature = angularReactivity(injector)
 
-  return ngZone.runOutsideAngular(() =>
+  const lazyTable = ngZone.runOutsideAngular(() =>
     lazyInit(() => {
-      // Explicit type arguments skip generic inference from the spread object
-      // (a type-check hot spot); the spread only adds the angular reactivity
-      // binding to `features`.
-      const table = constructTable<TFeatures, TData>({
-        ...options(),
-        features: {
-          coreReactivityFeature: angularReactivity(injector),
-          ...options().features,
-        },
+      const currentOptions = options()
+      const features = {
+        coreReactivityFeature,
+        ...currentOptions.features,
+      } satisfies TableFeatures
+      return constructTable<TFeatures, TData>({
+        ...currentOptions,
+        features,
       })
-
-      injector.get(DestroyRef).onDestroy(() => {
-        table._reactivity.unmount?.()
-      })
-
-      let previousOptions = options()
-      effect(
-        () => {
-          const currentOptions = options()
-          if (previousOptions === currentOptions) {
-            return
-          }
-          untracked(() =>
-            table.setOptions((previous) => ({
-              ...previous,
-              ...currentOptions,
-            })),
-          )
-          previousOptions = currentOptions
-        },
-        { injector, debugName: 'tableOptionsUpdate' },
-      )
-
-      return table
     }),
   )
+
+  destroyRef.onDestroy(() => {
+    if (lazyTable.initialized) {
+      lazyTable.value._reactivity.unmount?.()
+    }
+  })
+
+  let previousOptions: TableOptions<TFeatures, TData> | undefined = undefined
+  effect(
+    () => {
+      const currentOptions = options()
+      const tableInstance = lazyTable.rawValue
+      if (previousOptions === currentOptions) return
+      untracked(() =>
+        tableInstance.setOptions((previous) => ({
+          ...previous,
+          ...currentOptions,
+        })),
+      )
+      previousOptions = currentOptions
+    },
+    { injector, debugName: 'tableOptionsUpdate' },
+  )
+
+  return lazyTable.value
 }
--- a/packages/angular-table/src/injectTable.ts
+++ b/packages/angular-table/src/injectTable.ts
@@ -3,6 +3,7 @@ import {
   Injector,
   NgZone,
   assertInInjectionContext,
+  computed,
   effect,
   inject,
   untracked,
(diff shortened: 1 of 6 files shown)


MILLERMARRU (commented):
The mechanism here is solid. Wrapping `_options` in `computed()` is what makes the `previousOptions === currentOptions` reference check meaningful in the first place, calling the raw factory twice always returns two different object literals, so without the memoization the comparison would never be equal and every effect run would apply. With `computed()`, a real signal change is what produces a new reference, so the effect's very first run can now tell "options actually changed since I captured the initial snapshot" from "nothing changed, this is just the effect settling," instead of unconditionally skipping the first run like `isMount` did. Traced the timeline in #6530 and this correctly handles the specific race described there, an update that lands in the gap between construction and the effect's first flush is no longer silently dropped.

Worth being precise about scope though, since the changeset and the linked issue don't quite line up. #6530 describes two problems: the primary one is `lazySignalInitializer.ts`'s `queueMicrotask` eager-init firing before Angular's input bindings are applied, throwing an uncaught `NG0950` in test environments with an async `beforeEach`. The `isMount` swallow is called out as a secondary issue, one that only surfaces once you work around the first one. This PR only touches `injectTable.ts`, `lazySignalInitializer.ts` isn't part of the diff, so it fixes proposal (3) from the issue but not proposal (1). The reported symptom, the uncaught NG0950 in TestBed suites, is still there after this merges. Worth linking as "part of #6530" rather than "closes #6530", or the eager-init half is going to look resolved when it isn't.

MILLERMARRU:
Just went through the new commit, this closes the gap I mentioned earlier. `queueMicrotask` is gone from `lazySignalInitializer.ts`, so initialization now only happens through the Proxy traps, there's no more scheduled work that could race ahead of Angular applying the input bindings. Between that and the `previousOptions` comparison from the first commit, this now covers both halves of #6530, not just the swallowed-update part.

Traced through a few sequences by hand (options changing before the table is ever constructed, options changing right after) and the effect's `if (!table) return` guard combined with the reference comparison handles them correctly in all of them.

riccardoperra:
@KevinVandy pr ready

packages/angular-table/src/injectTable.ts
```diff
@@ -91,48 +92,51 @@ export function injectTable<
...
 >(
-  options: () => TableOptions<TFeatures, TData>,
+  _options: () => TableOptions<TFeatures, TData>,
 ): AngularTable<TFeatures, TData> {
   assertInInjectionContext(injectTable)
   const injector = inject(Injector)
   const ngZone = inject(NgZone)
+  const destroyRef = inject(DestroyRef)
+  const options = computed(() => _options())
+  const coreReactivityFeature = angularReactivity(injector)
```
riccardoperra:
```suggestion
  optionsFactory: () => TableOptions<TFeatures, TData>,
): AngularTable<TFeatures, TData> {
assertInInjectionContext(injectTable)
const injector = inject(Injector)
const ngZone = inject(NgZone)
  const destroyRef = inject(DestroyRef)
  const options = computed(() => optionsFactory())
  const coreReactivityFeature = angularReactivity(injector)
```

packages/angular-table/src/injectTable.ts
```diff
@@ -91,48 +92,51 @@ export function injectTable<
...
-            })),
-          )
-        },
-        { injector, debugName: 'tableOptionsUpdate' },
+  let previousOptions: TableOptions<TFeatures, TData> | undefined = undefined
+  effect(
+    () => {
+      const currentOptions = options()
+      const tableInstance = lazyTable.rawValue
+      if (previousOptions === currentOptions) return
```
riccardoperra:
```suggestion
      // rawValue will be always valued here due to internal lazyInit effect
      const tableInstance = lazyTable.rawValue
      if (previousOptions === currentOptions) return
```

imaksp:
This has resolved an issue with required input usage and also one setOptions workaround, thanks!

PR #6540, opened 2026-08-11
fix(table-core): skip onColumnVisibilityChange for a redundant toggle-all

### LINKED ISSUE #6539: toggleAllColumnsVisible fires onColumnVisibilityChange on a no-op, same gap as #6538 for row selection
Same gap as #6538, one feature over: `table_toggleAllColumnsVisible` has no guard against firing `onColumnVisibilityChange` when the requested visibility state already matches the current one, unlike `table_toggleAllRowsExpanded` after #6501 (closing #6136).

`columnVisibilityFeature.utils.ts`, current `main`:

```ts
export function table_toggleAllColumnsVisible(table, value?) {
  value = value ?? !table_getIsAllColumnsVisible(table)
  const visibility = makeObjectMap<boolean>()
  const leafColumns = table.getAllLeafColumns()
  for (let i = 0; i < leafColumns.length; i++) {
    const column = leafColumns[i]!
    visibility[column.id] = !value ? !column_getCanHide(column) : value
  }
  table_setColumnVisibility(table, visibility)
}
```

It rebuilds the full visibility map and calls `table_setColumnVisibility` unconditionally, there's no early return comparable to `table_toggleAllRowsExpanded`'s `if (currentExpanded === true) return` for the already-all-shown/already-all-hidden case.

### Repro

```ts
table.toggleAllColumnsVisible(true)  // shows every column, fires onColumnVisibilityChange
table.toggleAllColumnsVisible(true)  // every column already visible...
// onColumnVisibilityChange still fires, with a map equal to the previous one
```

`table_getIsAllColumnsVisible(table)` is already available and cheap to check (it's what the function itself falls back to when `value` isn't passed), so guarding the explicit-`value` branch the same way wouldn't need a new primitive, just the same `if (value === table_getIsAllColumnsVisible(table)) return` shape #6501 used for expansion.

Filing this and #6538 as separate reports since they're two independent call sites in two different features, even though the underlying pattern (and likely the fix) is the same.

lazerg:
## 🎯 Changes

`table_toggleAllColumnsVisible` rebuilds the whole visibility map and hands it to `table_setColumnVisibility` unconditionally, so asking for a state the table is already in still runs `onColumnVisibilityChange`:

```ts
table.toggleAllColumnsVisible(true) // shows every column
table.toggleAllColumnsVisible(true) // nothing left to change, the handler fires anyway
```

`setStateSlice` can't catch this one. The map it receives (`{ id: true, firstName: true, ... }`) is structurally different from the `{}` the state usually holds while every column is visible, so the two never compare equal. The fix compares the map against each leaf column's current visibility instead and returns early when they already agree, the same kind of explicit no-op check `table_toggleAllRowsExpanded` keeps.

Two existing tests toggled all columns on from a table where everything was already visible, which is now a no-op. Both start from a hidden `firstName` instead, so they still exercise the show-all path.

Fixes #6539

## ✅ Checklist

Two tests added to `columnVisibilityFeature.utils.test.ts`, both verified to fail on `main`: one for the already-all-visible case, one for the already-all-hidden case.

## 🚀 Release Impact

## Summary by CodeRabbit

* **Bug Fixes**
  * Toggling all columns now does nothing when they already match the requested visibility state.
  * Prevented unnecessary visibility updates and change notifications in no-op scenarios.
  * Improved behavior when showing or hiding all columns from partially or fully hidden states.

* **Documentation**
  * Updated column visibility function references and documented the no-op behavior.

### DIFF
--- a/.changeset/toggle-all-columns-visible-noop.md
+++ b/.changeset/toggle-all-columns-visible-noop.md
@@ -1,0 +1,5 @@
+---
+'@tanstack/table-core': patch
+---
+
+Skip `onColumnVisibilityChange` when `table.toggleAllColumnsVisible()` is called with the visibility state every leaf column is already in. This matches the no-op guards `table.toggleAllRowsExpanded()` already has.
--- a/docs/reference/static-functions/functions/table_getIsAllColumnsVisible.md
+++ b/docs/reference/static-functions/functions/table_getIsAllColumnsVisible.md
@@ -9,7 +9,7 @@ title: table_getIsAllColumnsVisible
 function table_getIsAllColumnsVisible<TFeatures, TData>(table): boolean;
 ```
 
-Defined in: [features/column-visibility/columnVisibilityFeature.utils.ts:360](https://github.com/TanStack/table/blob/main/packages/table-core/src/features/column-visibility/columnVisibilityFeature.utils.ts#L360)
+Defined in: [features/column-visibility/columnVisibilityFeature.utils.ts:373](https://github.com/TanStack/table/blob/main/packages/table-core/src/features/column-visibility/columnVisibilityFeature.utils.ts#L373)
 
 Checks whether every leaf column is currently visible.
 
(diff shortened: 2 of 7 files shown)


lazerg:
Closing this since the linked issue (#6539) has been closed, so this fix isn't needed anymore. Thanks anyway for taking a look at it.

PR #6541, opened 2026-08-11
fix(table-core): keep depth-truncated sub-rows in the leaf-up filter path

### LINKED ISSUE #6537: maxLeafRowFilterDepth: leaf-up path drops truncated subRows entirely, unlike root-down after #6503
#6503 fixed `maxLeafRowFilterDepth` under `filterRowModelFromRoot` so a truncated subtree's descendants still land in `flatRows`/`rowsById` even though recursion stopped. The PR says this was deliberately scoped to that one path:

> `maxLeafRowFilterDepth` keeps `flatRows`/`rowsById` complete. When the root-down recursion stops at the max depth, the kept rows' unfiltered descendants stay visible through `row.subRows` but never entered the flat arrays, under-counting facets after filtering. Truncated subtrees now join both. Scoped to the root-down path like the original PR; the leaf-up path drops truncated subRows entirely (pre-existing behavior) and is left as a possible follow-up.

Checked `filterRowsUtils.ts` on `main`: `filterRowModelFromRoot` now calls `addSubRowsToFlatArrays(row.subRows, newFilteredFlatRows, newFilteredRowsById)` when `depth >= maxDepth`, but `filterRowModelFromLeafs`'s `recurseFilterRows` has no equivalent. Its recursion is `newRow.subRows = recurseFilterRows(row.subRows, depth + 1)`, gated by `if (row.subRows.length && depth < maxDepth)`, with no branch for `depth >= maxDepth`, so once the max depth is hit the subtree just isn't included in `subRows`, `flatRows`, or `rowsById` at all, not "still visible but not flattened," actually gone.

That's a different failure mode than the one #6503 fixed (silent under-count in flat-derived data like facets) since here the descendants disappear from `row.subRows` too, so they're not even rendered as nested rows in the tree, not just missing from the flat arrays.

Repro (with `filterFromLeafRows: true`):

```ts
const table = useReactTable({
  data, columns,
  getSubRows: row => row.subRows,
  getFilteredRowModel: getFilteredRowModel(),
  filterFns: { fuzzy: fuzzyFilter },
  state: { columnFilters },
  filterFromLeafRows: true,
  maxLeafRowFilterDepth: 1,
})

// row 'a' has a matching child 'a1' at depth 1, and 'a1' has its own
// child 'a1x' at depth 2, beyond maxLeafRowFilterDepth

table.getFilteredRowModel().rowsById['a1']?.subRows
// expected (matches root-down behavior after #6503): still contains a1x
// actual: [] — a1x is gone from subRows, flatRows, and rowsById alike
```

Since #6503 already names this exact gap as a known follow-up, opening this mainly so there's something to reference from a PR rather than leaving the note buried in an unrelated changelog entry.

lazerg:
## 🎯 Changes

With `filterFromLeafRows: true`, a row kept at `maxLeafRowFilterDepth` loses its whole subtree. `filterRowModelFromLeafs` rebuilds every row through `constructRow`, and once the depth cap stops the recursion the rebuilt row never gets its sub-rows back, so the descendants vanish from `subRows`, `flatRows` and `rowsById` at once.

```ts
// maxLeafRowFilterDepth: 1, 'keep-a1' matches at depth 1 and owns 'drop-a1a' at depth 2
table.getFilteredRowModel().rows[0].subRows[0].subRows
// before: []
// after:  [drop-a1a]
```

The root-down path keeps the original row instead of a rebuilt one, so its unfiltered subtree stays visible, and #6503 made those descendants join the flat arrays through `addSubRowsToFlatArrays`. The leaf-up path now carries the original sub-rows onto the rebuilt row and flattens them with that same helper, which #6503 listed as the remaining follow-up.

Fixes #6537

## ✅ Checklist

Two tests added to `createFilteredRowModel.test.ts`, mirroring the root-down ones already there for depth 0 and depth 1. Both fail on `main` and pass with the fix. `@tanstack/table-core` is at 1318/1318 tests, with eslint, types and build clean. The only red target is `@tanstack/ember-table:test:lib`, where testem cannot reach a local headless Chrome, and it fails the same way without this change.

## 🚀 Release Impact

## Summary by CodeRabbit

- **Bug Fixes**
  - Fixed hierarchical filtering when the maximum leaf-row filter depth is reached.
  - Retained rows now preserve their unfiltered descendants, including nested rows and flattened row results.
  - Row references and ordering remain consistent across filtered table results.

- **Tests**
  - Added coverage for filtering at multiple depth limits, including validation of nested rows and flattened results.

### DIFF
--- a/.changeset/leaf-up-truncated-sub-rows.md
+++ b/.changeset/leaf-up-truncated-sub-rows.md
@@ -1,0 +1,5 @@
+---
+'@tanstack/table-core': patch
+---
+
+Fix `filterFromLeafRows` discarding the sub-rows of rows kept past `maxLeafRowFilterDepth`. The leaf-up filter path rebuilt those rows without their unfiltered subtree, so the truncated descendants vanished from `subRows`, `flatRows` and `rowsById` alike. They are now carried over and flattened, the way the root-down path already keeps them.
--- a/packages/table-core/src/features/column-filtering/filterRowsUtils.ts
+++ b/packages/table-core/src/features/column-filtering/filterRowsUtils.ts
@@ -82,11 +82,20 @@ function filterRowModelFromLeafs<
           continue
         }
       } else {
+        // Past maxLeafRowFilterDepth the subtree is never filtered, so it stays
+        // visible through row.subRows and its rows must enter flatRows and
+        // rowsById as well, like the root-down path already does
+        newRow.subRows = row.subRows
         row = newRow
         if (filterRow(row)) {
           filteredRows.push(row)
           newFilteredRowsById[row.id] = row
           newFilteredFlatRows.push(row)
+          addSubRowsToFlatArrays(
+            row.subRows,
+            newFilteredFlatRows,
+            newFilteredRowsById,
+          )
         }
       }
     }
--- a/packages/table-core/tests/implementation/features/column-filtering/createFilteredRowModel.test.ts
+++ b/packages/table-core/tests/implementation/features/column-filtering/createFilteredRowModel.test.ts
@@ -319,6 +319,55 @@ describe('createFilteredRowModel', () => {
       expect(model.rowsById[keepA1.id]).toBe(keepA1)
     })
 
+    it('should include unfiltered descendants of kept rows in flatRows and rowsById (from leaf, depth 0)', () => {
+      const table = makeNestedTable({
+        filterFromLeafRows: true,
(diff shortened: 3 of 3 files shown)


lazerg:
Closing this since the linked issue (#6537) was closed by a maintainer without a fix landing anywhere, so this PR isn't needed anymore. Happy to reopen if the underlying problem comes back up.

PR #6542, opened 2026-08-11
fix(table-core): skip onRowSelectionChange on a no-op toggleAllRowsSelected

### LINKED ISSUE #6538: toggleAllRowsSelected fires onRowSelectionChange on a no-op, unlike toggleAllRowsExpanded after #6501
#6501 (closing #6136) added a no-op guard to `table_toggleAllRowsExpanded` so it stops calling `onExpandedChange` when the requested state already matches the current one. `table_toggleAllRowsSelected` has the identical shape of API (imperative toggle-all with an optional explicit boolean) and never got the same guard.

`rowExpandingFeature.utils.ts`, post-#6501:

```ts
export function table_toggleAllRowsExpanded(table, expanded?) {
  const currentExpanded = table.atoms.expanded?.get() ?? {}
  if (expanded ?? !table_getIsAllRowsExpanded(table)) {
    if (currentExpanded === true) return
    if (!table_getCanSomeRowsExpand(table)) return
    table_setExpanded(table, true)
  } else {
    if (currentExpanded !== true && !Object.keys(currentExpanded).length) return
    table_setExpanded(table, makeObjectMap())
  }
}
```

`rowSelectionFeature.utils.ts`, current `main`:

```ts
export function table_toggleAllRowsSelected(table, value?, opts?) {
  table._lastSelectedRowId = null
  table_setRowSelection(table, (old) => {
    value = typeof value !== 'undefined' ? value : !callMemoOrStaticFn(table, 'getIsAllRowsSelected', table_getIsAllRowsSelected)
    if (opts?.deselectAll && !value) return makeObjectMap<true>()
    const rowSelection = Object.assign(makeObjectMap<true>(), old)
    // ...builds the full selection map unconditionally...
  })
}
```

No early return for the case where the resulting selection already matches `old`. `table_setRowSelection` itself is intentionally unguarded too (there's a comment explaining why: a full deep-equal over an arbitrary-sized selection map would be real overhead for genuine changes), but that rationale doesn't apply to `toggleAllRowsSelected` specifically, the O(1)-ish checks `table_toggleAllRowsExpanded` added (`getIsAllRowsSelected()` is already memoized, an empty-map check) are the same shape of cheap guard, not a full map comparison.

### Repro

```ts
table.toggleAllRowsSelected(true) // selects everything, fires onRowSelectionChange
table.toggleAllRowsSelected(true) // everything is already selected...
// onRowSelectionChange still fires, with a functionally identical resulting map
```

Same symptom #6136 described for expansion: a consumer syncing `rowSelection` to external state (URL, server) sees redundant writes on every no-op toggle, and any `useEffect`/watcher keyed on `onRowSelectionChange` re-runs for nothing.

Might be intentional given the differing rationale for the general setter, flagging with that caveat rather than assuming it's simply missed, but the parity with the just-fixed expansion API made it worth raising explicitly.

lazerg:
## 🎯 Changes

`table.toggleAllRowsSelected(true)` rebuilds the whole selection map and calls `onRowSelectionChange` even when every selectable row is already selected, and `toggleAllRowsSelected(false)` does the same when nothing is selected. Consumers that mirror `rowSelection` into a URL or onto a server get a redundant write on every no-op toggle.

`toggleAllRowsExpanded` grew the equivalent guards in #6501, and #6532 kept those O(1) checks on purpose while leaving the general `setRowSelection` unguarded. This adds the matching pair: return early when selection is requested and `getIsAllRowsSelected()` is already true, and when deselection is requested and nothing is selected. Both read memoized getters, so the row-scaled map comparison that `setRowSelection` deliberately avoids stays out of the picture.

Three adapter tests drove their second `onRowSelectionChange` through `toggleAllRowsSelected(false)` with a mock handler that never writes state back, leaving the selection empty, so that call is now a no-op. They toggle selection on instead.

Fixes #6538

## ✅ Checklist

Green across all 409 projects except `@tanstack/ember-table:test:lib`, whose testem browser will not start on my machine. It fails the same way on an unmodified `main`.

## 🚀 Release Impact

## Summary by CodeRabbit

* **Bug Fixes**
  * Prevented redundant row-selection updates when all selectable rows are already selected or deselected.
  * Avoided triggering selection change callbacks for no-op toggle requests.
  * Preserved selection updates when the table contains a mix of selected and unselected rows.

* **Tests**
  * Added coverage for redundant selection operations and partial-selection scenarios.

### DIFF
--- a/.changeset/toggle-all-rows-selected-noop.md
+++ b/.changeset/toggle-all-rows-selected-noop.md
@@ -1,0 +1,5 @@
+---
+'@tanstack/table-core': patch
+---
+
+Stop `table.toggleAllRowsSelected()` from calling `onRowSelectionChange` when the requested state already matches the current one. Selecting when every selectable row is already selected, and deselecting when nothing is selected, are now no-ops, matching the O(1) guards `table.toggleAllRowsExpanded()` already has.
--- a/packages/alpine-table/tests/unit/adapterCoverage.test.ts
+++ b/packages/alpine-table/tests/unit/adapterCoverage.test.ts
@@ -230,7 +230,7 @@ describe('Alpine adapter reactivity', () => {
 
     options.onRowSelectionChange = secondHandler
     await flushEffects()
-    table.toggleAllRowsSelected(false)
+    table.toggleAllRowsSelected(true)
 
     expect(firstHandler).toHaveBeenCalledTimes(1)
     expect(secondHandler).toHaveBeenCalledTimes(1)
--- a/packages/lit-table/tests/unit/adapterLifecycle.test.ts
+++ b/packages/lit-table/tests/unit/adapterLifecycle.test.ts
@@ -369,7 +369,7 @@ describe('TableController lifecycle and option ownership', () => {
       '{"canSelect":false,"columnIds":["title"],"values":["Final"]}',
     )
 
-    element.table!.toggleAllRowsSelected(false)
+    element.table!.toggleAllRowsSelected(true)
 
     expect(firstSelectionHandler).toHaveBeenCalledOnce()
     expect(secondSelectionHandler).toHaveBeenCalledOnce()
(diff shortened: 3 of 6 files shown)


lazerg:
Closing this since the linked issue (#6538) was closed by the maintainers without the guard being adopted, so this fix isn't needed anymore. Happy to reopen if the behavior gets revisited.

PR #6545, opened 2026-08-12
fix(table-core): flatten filtered parent rows ahead of their sub-rows

### LINKED ISSUE #6536: getFilteredRowModel().flatRows is post-order, same bug #6529 just fixed for getSortedRowModel
`getFilteredRowModel().flatRows` has the exact same post-order bug that #6529 just fixed for `getSortedRowModel().flatRows`, and the PR itself flags it as a known follow-up that wasn't in scope there.

From #6529's description:

> `filterRowModelFromLeafs` and `filterRowModelFromRoot` flatten the same way, so `getFilteredRowModel().flatRows` is post-order too. I left those alone... That felt like a bigger behavioral call than this fix, happy to do it in a follow-up if you'd like it changed.

I checked `filterRowsUtils.ts` on current `main` and both paths still push children before their own parent:

`filterRowModelFromLeafs` recurses into `row.subRows` first (`newRow.subRows = recurseFilterRows(...)`), and only pushes the row to `newFilteredFlatRows` after that call returns, so descendants land in the array first.

`filterRowModelFromRoot` (the default, `filterFromLeafRows: false`) looks parent-first at a glance, but the recursive call `newRow.subRows = recurseFilterRows(row.subRows, depth + 1)` runs and pushes the children to the shared `newFilteredFlatRows` closure *before* the parent's own `newFilteredFlatRows.push(row)` a few lines later executes. Same net effect, children first.

Repro, mirroring the one in #6529:

```ts
const table = useReactTable({
  data,
  columns,
  getSubRows: row => row.subRows,
  getFilteredRowModel: getFilteredRowModel(),
  state: { columnFilters },
})

// two parent rows with sub-rows, both passing the filter
table.getCoreRowModel().flatRows.map(r => r.id)
// ['a', 'a1', 'a2', 'b', 'b1']

table.getFilteredRowModel().flatRows.map(r => r.id)
// ['a1', 'a2', 'a', 'b1', 'b']   <- descendants before their own parent
```

This disagrees with `rows` (already in render order) and with `getCoreRowModel`/`getPaginatedRowModel`, which both push parent-first. It matters concretely for `column_getAutoSortFn`/`column_getAutoFilterFn`, which sample `getFilteredRowModel().flatRows.slice(0, 10)` to auto-detect a column's data type, same concern #6529 raised for why it didn't touch this itself.

Since #6529 already explains the mechanism and offers to follow up, this issue is just to have something to track and link a PR against, rather than leaving it as a comment buried in an unrelated PR.

waterWang:
## 🎯 Changes

`getFilteredRowModel().flatRows` has the exact same post-order problem that
[#6529](https://github.com/TanStack/table/pull/6529) fixed for
`getSortedRowModel().flatRows`.

### Root cause

Both `filterRowModelFromLeafs` (used when `filterFromLeafRows: true`) and
`filterRowModelFromRoot` (default) recurse into a row's sub-rows *before*
pushing the row itself to `flatRows`, so every descendant precedes its own
parent in the flattened list.

### Fix

- **`filterRowModelFromRoot`**: reserves the parent's flat-array slot before
  descending (same pattern as #6529's `createSortedRowModel`), then replaces
  it with the cloned row when recursion returns.
- **`filterRowModelFromLeafs`**: records the insertion point before recursion
  and splices the parent ahead of its surviving children when the parent is
  retained.

### Before

```ts
table.getCoreRowModel().flatRows.map((r) => r.original.name)
// ['keep-a', 'keep-a1', 'drop-a1a', 'drop-a2', 'drop-b', 'keep-b1',
//  'keep-c', 'keep-d', 'drop-d1']

table.getFilteredRowModel().flatRows.map((r) => r.original.name)
// before: ['keep-a1', 'keep-a', 'keep-c', 'keep-d']
// after:  ['keep-a', 'keep-a1', 'keep-c', 'keep-d']
```

### Checklist

- [x] `filterRowModelFromLeafs` fix
- [x] `filterRowModelFromRoot` fix
- [x] Update existing test expectation (the `maxLeafRowFilterDepth: 1` test
  that explicitly documented the old post-order behaviour)
- [x] Add new tests for both modes

Closes #6536

## Summary by CodeRabbit

* **Bug Fixes**
  * Corrected filtered row ordering so parent rows consistently appear before their surviving sub-rows.
  * Aligned flat row results with the displayed filtered row hierarchy across all filtering modes.

* **Tests**
  * Added coverage for hierarchical filtering and updated depth-limited filtering expectations.

### DIFF
--- a/.changeset/great-pugs-sniff.md
+++ b/.changeset/great-pugs-sniff.md
@@ -1,0 +1,5 @@
+---
+"@tanstack/table-core": patch
+---
+
+Fix `getFilteredRowModel().flatRows` listing sub-rows before their parent in both `filterFromLeafRows` and `filterFromRoot` modes. This restores the parent-first order used by the core, sorted, and paginated row models, and aligns the flattened result with the filtered `rows` tree.
--- a/packages/table-core/src/features/column-filtering/filterRowsUtils.ts
+++ b/packages/table-core/src/features/column-filtering/filterRowsUtils.ts
@@ -65,20 +65,25 @@ function filterRowModelFromLeafs<
       newRow.columnFilters = row.columnFilters
 
       if (row.subRows.length && depth < maxDepth) {
+        // Descendants are filtered first, so in the flat list a parent must
+        // be inserted before its surviving children (pre-order) to match the
+        // readable `rows` tree and the core/sorted/paginated row models.
+        const flatIndex = newFilteredFlatRows.length
+
         newRow.subRows = recurseFilterRows(row.subRows, depth + 1)
         row = newRow
 
         if (filterRow(row) && !newRow.subRows.length) {
+          newFilteredFlatRows.splice(flatIndex, 0, row)
           filteredRows.push(row)
           newFilteredRowsById[row.id] = row
-          newFilteredFlatRows.push(row)
           continue
         }
 
         if (filterRow(row) || newRow.subRows.length) {
+          newFilteredFlatRows.splice(flatIndex, 0, row)
           filteredRows.push(row)
           newFilteredRowsById[row.id] = row
-          newFilteredFlatRows.push(row)
           continue
         }
       } else {
@@ -127,6 +132,11 @@ function filterRowModelFromRoot<
       const pass = filterRow(row)
 
       if (pass) {
+        // Take this row's slot before descending, so a parent stays ahead of
+        // its own sub-rows in flatRows (pre-order).
+        const flatIndex = newFilteredFlatRows.length
+        newFilteredFlatRows.push(row)
+
         if (row.subRows.length && depth < maxDepth) {
           const newRow = constructRow(
             table,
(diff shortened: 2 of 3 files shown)


MILLERMARRU (commented):
This looks like it has already been fixed by the merged #6568. I pulled the current `filterRowsUtils.ts` from `main`:

- `filterRowModelFromRoot` now does `newFilteredFlatRows.push(newRow)` before `newRow.subRows = recurseFilterRows(...)`, so the parent lands ahead of its descendants (the reserve-slot dance this PR adds is not needed there any more).
- `filterRowModelFromLeafs` was reworked to build `flatRows` from the finished tree via `addSubRowsToFlatArrays`, which is parent-first by construction.
- #6568 also folded in `newRow.columnFiltersMeta = filterData.columnFiltersMeta`, which is what the now-closed #6557 was for.

So `getFilteredRowModel().flatRows` on `main` is already pre-order in both `filterFromLeafRows` modes, which is why this PR now shows as conflicting. Unless there is a case #6568 missed, this can probably be closed. If there is a gap, it would help to rebase onto `main` and show the failing assertion against the current implementation rather than the old one.

(Context: I opened #6536 for the root-mode half of this and it is closed by #6568.)

PR #6548, opened 2026-08-13
fix(mantine-react-table): expose loading state slices on table.state

naughton:
The post-useTable patch that re-injects MRT-only state slices onto table.state omits the controlled-only loading slices (isLoading, isSaving, showLoadingOverlay, showProgressBars, showSkeletons). Those slices have no internal useState — they only arrive via options.state — so every component that destructures them from table.state reads undefined.

Visible symptom: with state.isLoading true, the blank skeleton rows are generated, but MRT_TableBodyCell's skeleton branch never triggers, so column Cell renderers run against the null placeholder data (epoch dates, fallback strings) instead of rendering <Skeleton>s. The loading overlay and progress bars are similarly dead.

showSkeletons is intentionally not defaulted to false: MRT_TableBodyCell checks `showSkeletons !== false` — false means 'suppress skeletons even while loading', undefined means 'auto'.

## Summary by CodeRabbit

* **Bug Fixes**
  * Improved loading-state handling for controlled tables.
  * Loading indicators now display consistently while preserving explicitly configured skeleton visibility.

### DIFF
--- a/examples/react/mantine-react-table/src/mantine-react-table/hooks/useMRT_TableInstance.ts
+++ b/examples/react/mantine-react-table/src/mantine-react-table/hooks/useMRT_TableInstance.ts
@@ -391,6 +391,17 @@ export const useMRT_TableInstance = <TData extends MRT_RowData>(
     showColumnFilters,
     showGlobalFilter,
     showToolbarDropZone,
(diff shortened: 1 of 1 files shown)


hillac:
This is also affecting me, I'm getting runtime errors due to the placeholder data attempting to be rendered.

PR #6552, opened 2026-08-14
fix(table-core): flatten grouped parent rows ahead of their sub-rows

### LINKED ISSUE #6551: getGroupedRowModel().flatRows puts every group row after all its descendants, not just tree data below terminal depth
`getSortedRowModel().flatRows` was just fixed by #6529 to put each parent ahead of its own sub-rows instead of post-order, and #6529's own description notes `createCoreRowModel`/`createPaginatedRowModel` already push parent-before-children, and that `getFilteredRowModel().flatRows` was deliberately left post-order for now (I filed #6536 for that one, still open).

`getGroupedRowModel().flatRows` has the same class of bug, and it's not scoped to "tree data below the terminal grouping depth" the way `perf-new.md`'s N1 write-up frames it ("(a) tree-data rows below terminal depth move from preorder to postorder in grouped flatRows... the sorted model already emits postorder"). That justification is stale now that #6529 fixed sorted's ordering, and reading the current `createGroupedRowModel.ts` on `main`, the issue is broader than tree-data: every group-level row, at every depth, is pushed to `flatRows` by its *parent's* loop, which only runs after that parent's own recursive call has fully returned:

```ts
const subRows = groupUpRecursively(groupedRows, depth + 1, id)   // descend first
...
const row = constructRow(table, id, ...)                          // build this group's row
...
subRows.forEach((subRow) => {                                     // push the CHILDREN
  groupedFlatRows.push(subRow)
  groupedRowsById[subRow.id] = subRow
})
return row                                                         // return row WITHOUT pushing it
```

and the group row itself is only pushed by whichever loop called `groupUpRecursively` for it — for a top-level group, that's the very last statement in the whole function (`groupedRows.forEach((subRow) => { groupedFlatRows.push(subRow); ... })`), which can't run until recursion has finished for *every* top-level group.

Concrete repro, single-level grouping (`grouping: ['a']`), no tree data at all, so N1's terminal-depth explanation doesn't apply:

```ts
// rows: { id: 'r1', a: 1 }, { id: 'r2', a: 1 }, { id: 'r3', a: 2 }
table.getCoreRowModel().flatRows.map(r => r.id)
// ['r1', 'r2', 'r3']

table.getGroupedRowModel().rows.map(r => r.id)
// ['a:1', 'a:2']  -- correctly parent-first as a tree

table.getGroupedRowModel().flatRows.map(r => r.id)
// actual:   ['r1', 'r2', 'r3', 'a:1', 'a:2']
// expected: ['a:1', 'r1', 'r2', 'a:2', 'r3']  (matches `rows`, matches core/paginated/now-sorted convention)
```

Every leaf row from every group lands ahead of every group row, not just "descendants below the terminal depth" — the two `getGroupedRowModel` tests added for N1 (`single-level flat 12→7`, `two-level flat 13→9`, etc., in `tests/implementation/features/column-grouping/createGroupedRowModel.test.ts`) check the flatRows *count* dropped after de-duplication but don't assert this ordering, so it wasn't caught there.

This also means `getExpandedRowModel().flatRows` and the paginated model inherit the same ordering, same as the N1 write-up already notes for the duplicate-push case ("expandRows forwards grouped flatRows by reference... the duplication propagates down the pipeline") — here it'd be "the ordering propagates down the pipeline" instead.

Given #6529 already established parent-before-children as the intended contract for `flatRows` and specifically invalidated the "sorted already does this" reasoning `perf-new.md` used to accept the grouped-model gap, it seems worth re-opening/widening N1's accepted-delta scope rather than leaving it as accepted. Happy to help verify a fix mirroring #6529's approach (take the row's flatRows slot before descending, overwrite it when the row's subRows get reassigned) if useful.

lazerg:
## 🎯 Changes

In `_createGroupedRowModel` every frame pushes the rows its *caller* asked for, so a group row only reaches `flatRows` once the parent's loop runs, which is after that group's whole subtree is already in. The result is that every group row sits behind its own descendants, at every depth, not just below the terminal grouping depth.

With `grouping: ['a']` over three flat rows and no tree data:

```ts
table.getGroupedRowModel().rows.map((r) => r.id)
// ['a:1', 'a:2']

table.getGroupedRowModel().flatRows.map((r) => r.id)
// before: ['0', '1', '2', 'a:1', 'a:2']
// after:  ['a:1', '0', '1', 'a:2', '2']
```

That disagrees with the grouped `rows` tree, and with how `createCoreRowModel`, `createPaginatedRowModel` and (since #6529) `createSortedRowModel` flatten. Each frame now pushes the rows it returns, and a group takes its `flatRows` slot before it descends, the same reservation #6529 used for the sorted parent. The worker's grouped tree rebuild follows, so `getGroupedRowModel().flatRows` matches between the sync and worker paths. Rows are still pushed exactly once, which was the point of the N1 fix; I updated that entry in `perf-new.md` since its accepted delta (a) no longer holds.

Fixes #6551

## ✅ Checklist

Three tests added to `createGroupedRowModel.test.ts` (single-level, multi-level, and tree data below the terminal depth) and one `flatRows` assertion added to the nested-grouping worker round trip. All four fail on `main` and pass with the fix. `@tanstack/table-core` is at 63 files / 1319 tests. The affected run over 407 projects is green apart from `@tanstack/ember-table:test:lib`, which fails the same way on an untouched `main` here ("Browser failed to connect within 120s").

## 🚀 Release Impact

## Summary by CodeRabbit

- **Bug Fixes**
  - Corrected grouped row ordering so flattened results consistently list each parent group before its descendants at every nesting level.
  - Aligned worker-rebuilt row models with the same parent-first ordering across applicable processing stages.
  - Preserved consistent flattened row results for grouped, sorted, paginated, and core row models.
- **Tests**
  - Added coverage for nested grouped data and verified flattened row order during model reconstruction.

### DIFF
--- a/.changeset/grouped-flat-rows-parent-first.md
+++ b/.changeset/grouped-flat-rows-parent-first.md
@@ -1,0 +1,5 @@
+---
+'@tanstack/table-core': patch
+---
+
+Fix `getGroupedRowModel().flatRows` listing every group row after its own descendants. Group rows now come first at every depth, which aligns the flattened result with the grouped `rows` tree and with the core, paginated and sorted row models.
--- a/packages/table-core/src/features/column-grouping/createGroupedRowModel.ts
+++ b/packages/table-core/src/features/column-grouping/createGroupedRowModel.ts
@@ -107,17 +107,14 @@ function _createGroupedRowModel<
       return rows.map((row) => {
         row.depth = depth
 
-        // Every row is pushed into flatRows/rowsById exactly once, by its
-        // parent frame: rows returned here are pushed by the caller (the
-        // parent group's loop or the root loop), so only descendants below
-        // the terminal depth are pushed here.
+        // Every row is pushed into flatRows/rowsById exactly once, by its own
+        // frame and before it descends, so a parent stays ahead of its own
+        // sub-rows.
+        groupedFlatRows.push(row)
+        groupedRowsById[row.id] = row
+
         if (row.subRows.length) {
           row.subRows = groupUpRecursively(row.subRows, depth + 1, row.id)
-          for (let i = 0; i < row.subRows.length; i++) {
-            const subRow = row.subRows[i]!
-            groupedFlatRows.push(subRow)
-            groupedRowsById[subRow.id] = subRow
-          }
         }
 
         return row
@@ -134,6 +131,11 @@ function _createGroupedRowModel<
       ([groupingValue, groupedRows], index) => {
         let id = `${columnId}:${groupingValue}`
         id = parentId ? `${parentId}>${id}` : id
+
+        // Take this group's slot before descending, so it stays ahead of its
+        // own sub-rows.
+        const flatIndex = groupedFlatRows.length
+        groupedFlatRows.push(undefined as unknown as Row<TFeatures, TData>)
 
         // First, Recurse to group sub rows before aggregation
         const subRows = groupUpRecursively(groupedRows, depth + 1, id)
@@ -206,10 +208,8 @@ function _createGroupedRowModel<
           },
         })
 
-        subRows.forEach((subRow) => {
-          groupedFlatRows.push(subRow)
-          groupedRowsById[subRow.id] = subRow
-        })
+        groupedFlatRows[flatIndex] = row
+        groupedRowsById[id] = row
 
         return row
       },
@@ -219,11 +219,6 @@ function _createGroupedRowModel<
   }
 
   const groupedRows = groupUpRecursively(rowModel.rows, 0)
-
-  groupedRows.forEach((subRow) => {
-    groupedFlatRows.push(subRow)
-    groupedRowsById[subRow.id] = subRow
-  })
 
   return {
     rows: groupedRows,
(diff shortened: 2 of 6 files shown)


HarperZ9:
The grouped fix in `createGroupedRowModel.ts` looks right — reserving each group's slot before descending so the parent lands ahead of its descendants, with the multi-level test pinning it. That part's self-contained.

The snag is `rebuildRowModel.ts`. #6568 merged today and set `flattenParentsFirst = stage === 'filtered' || stage === 'sorted'` on main, so `filtered` is now parent-first — the opposite of this PR's premise ("Only the filtered model still flattens post-order"). That's the conflict. If it rebases to `stage !== 'filtered'`, it'd put filtered back to post-order and quietly undo #6568.

Cleaner to extend #6568's condition (keep `filtered`/`sorted`, add the grouped stage) rather than replace it, and update the comment + perf-new.md since filtered isn't post-order anymore. The grouped change and its tests can stay as they are.

KevinVandy:
superceeded by my own PR. Conflicts and some more testing were added. Co-author credit given

PR #6555, opened 2026-08-15
fix(table-core): guard process.env access with isDev() for non-Node environments

### LINKED ISSUE #6078: `process is not defined` when used in Vanilla JS (without Node.js)
### TanStack Table version

v8.21.3

### Framework/Library version

vanilla (through Rails importmap)

I'm using @tan-stack/table-core in vanilla JS without any bundler or Node.js runtime. It seems that `createTable()` function checks runtime environment for debugging, which throw an exception for my setup.
```
const M = [t, c, s, a, n, l, p, m, F, r, f, C, h, S, g];
function createTable(e) {
    var t, n;
    process.env.NODE_ENV !== "production" && (e.debugAll || e.debugTable) && console.info("Creating Table Instance...");
```

I temporarily use a mock process object to workaround. Would it be nicer to check `process` at first?

(no need)

_No response_

### Terms & Code of Conduct

ErfanBagheri404:
Fixes #6078

Raw `process.env.NODE_ENV` references throw `process is not defined` when @tanstack/table-core is consumed in vanilla JS without a bundler (e.g. Rails importmap).

Add a safe `isDev()` utility that checks for the existence of `process` before accessing it, and replace all dev-only guards across table-core.

```ts
function isDev(): boolean {
  return typeof process !== "undefined" && process.env?.NODE_ENV !== "production"
}
```

## Summary by CodeRabbit

* **Bug Fixes**
  * Improved development-mode compatibility in browser and worker environments by avoiding reliance on Node.js-specific environment globals.
  * Standardized development warnings and diagnostic logging across table features, including columns, rows, filtering, sorting, and aggregation.

* **Refactor**
  * Centralized development-environment detection for more consistent debug behavior and diagnostics.

### DIFF
--- a/packages/table-core/src/core/columns/constructColumn.ts
+++ b/packages/table-core/src/core/columns/constructColumn.ts
@@ -1,4 +1,5 @@
 import type { Table_Internal } from '../../types/Table'
+import { isDev } from '../../utils'
 import type { CellData, RowData } from '../../types/type-utils'
 import type { TableFeatures } from '../../types/TableFeatures'
 import type {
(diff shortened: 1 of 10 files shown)


KevinVandy:
Every time I've tried to add code like this in a centralized way in the past around `process.env?.NODE_ENV`, I thought that it ended up breaking tree-shaking.

PR #6557, opened 2026-08-15
fix(table-core): preserve columnFiltersMeta in filterRowModelFromLeafs

### LINKED ISSUE #6074: filterFromLeafRows causes columnFiltersMeta to get wiped
### TanStack Table version

v8.21.3

### Framework/Library version

React v18.3.1

# Overview
If the `filterFromLeafRows` config option is `true`, `columnFiltersMeta` gets wiped from each row when the row model gets built. This happens even when there are no leaf rows (AKA `subRows`). A helper used to build the row model replaces each row with a copy, but neglects to copy this particular attribute over from the old one.

My use case for `columnFiltersMeta` is essentially identical to the one demonstrated in the [Filters Fuzzy](https://tanstack.com/table/latest/docs/framework/react/examples/filters-fuzzy?panel=sandbox) example (sorting based on a rank score which is calculated during filtering and stored in `columnFiltersMeta`).

# Expected behavior
My opinion is that each row's `columnFiltersMeta` should remain as-is amidst whatever else gets done during leaf-row filtering logic. The convenient behavior in my particular case would be for sub rows' search rankings to somehow get aggregated into the parent's value (e.g. take the max rank among parent and all sub rows). That would be analogous to the main point of `filterFromLeafRows`, which is that sub rows passing the filter cause parent rows to also pass. However, the thing that consumes `columnFiltersMeta` in my case (column `sortingFn`) has access to `row.subRows` directly anyway. I would support a new convenience feature for this specific filter/rank use case, but I don't see a need to bake any part of it into the generic `columnFiltersMeta` feature.

My hoped-for fix would be that `columnFiltersMeta` simply gets copied into the new row alongside other things that are already correctly copied (e.g. `columnFilters`).

# Repro
Forked from [React Example: Filters Fuzzy](https://tanstack.com/table/latest/docs/framework/react/examples/filters-fuzzy?panel=sandbox), with one line added: `filterFromLeafRows: true` in the table config.

https://stackblitz.com/edit/tanstack-table-uy4nphk2?file=src%2Fmain.tsx

_No response_

Maybe, I'll investigate and start debugging

### Terms & Code of Conduct

ErfanBagheri404:
Fixes #6074

When `filterFromLeafRows` is true, `filterRowModelFromLeafs` recreates each row via `constructRow` and copies `columnFilters`, but neglects to copy `columnFiltersMeta`. Users who store filter-rank metadata (e.g. for fuzzy-sort use cases) find their data wiped on every filter recalculation.

Copy `columnFiltersMeta` alongside `columnFilters` when building the filtered row model from leaf rows.

## Summary by CodeRabbit

* **Bug Fixes**
  * Preserved column filter metadata when filtering nested or hierarchical table rows.
  * Ensured filtered rows retain the metadata needed for consistent filtering behavior.

### DIFF
--- a/packages/table-core/src/features/column-filtering/filterRowsUtils.ts
+++ b/packages/table-core/src/features/column-filtering/filterRowsUtils.ts
@@ -63,6 +63,7 @@ function filterRowModelFromLeafs<
       ) as Row<TFeatures, TData> &
         Partial<Row_ColumnFiltering<TFeatures, TData>>
       newRow.columnFilters = row.columnFilters
+      newRow.columnFiltersMeta = row.columnFiltersMeta
 
       if (row.subRows.length && depth < maxDepth) {
         newRow.subRows = recurseFilterRows(row.subRows, depth + 1)


MILLERMARRU (commented):
I pulled the current filterRowsUtils.ts and confirmed the gap directly: `filterRowModelFromRoot` and `filterRowModelFromLeafs` both reconstruct rows via `constructRow`, but only the leaf-up path drops `columnFiltersMeta` on the copy while keeping `columnFilters`. Anything that reads per-row filter match metadata (e.g. highlighting the matched substring for a fuzzy filter) would silently lose it whenever `filterFromLeafRows` is enabled, with no equivalent loss in the root-down path. The fix is a one-line, symmetric addition that matches how `columnFilters` is already handled two lines above, so it's low risk.

PR #6559, opened 2026-08-15
fix(table-core): count array values correctly in facetedUniqueValues

### LINKED ISSUE #5405: Unable to access Map entries with array keys by value in getFacetedUniqueValues function
### TanStack Table version

8.12.0

### Framework/Library version

NextJS v14.0.4

I've encountered an issue when using the getFacetedUniqueValues function to retrieve the faceted unique values for a column in a table. The problem arises when the column value is an array and I attempt to access the corresponding facet using the array value as the key.

https://stackblitz.com/edit/tanstack-table-stfatx?file=src%2Fmain.tsx

Status: Shows the number of entries for each option
CleanShot 2024-03-13 at 00 33 21@2x

Labels: Nothing due to it being an array
CleanShot 2024-03-13 at 00 34 56@2x

None

### Terms & Code of Conduct

ErfanBagheri404:
Fixes #5405

When a cell value is an array (e.g. multi-select columns), `getUniqueValues` wraps it producing a nested array. `Map` lookup then compares arrays by reference, so equal arrays count as distinct facets and the facet list never shows the right frequency.

Normalize array values to a stable JSON string key before counting.

## Summary by CodeRabbit

* **Bug Fixes**
  * Fixed faceted filtering so identical array-valued entries are counted as a single unique value.
  * Preserved existing handling for scalar facet values.

### DIFF
--- a/packages/table-core/src/features/column-faceting/createFacetedUniqueValues.ts
+++ b/packages/table-core/src/features/column-faceting/createFacetedUniqueValues.ts
@@ -82,10 +82,16 @@ function _createFacetedUniqueValues<
       if (!values) continue
 
       for (let j = 0; j < values.length; j++) {
-        const value = values[j]
-        const previousValue = facetedUniqueValues.get(value)
+        const value = values[j]!
+        // Array-valued cells (e.g. multi-select columns) are wrapped in a
+        // single-item array by getUniqueValues, so `value` is an array. Arrays
+        // compare by reference in a Map, so two equal arrays would count as
+        // distinct facets. Normalize array values to a stable string key so
+        // the facet count is correct.
+        const key = Array.isArray(value) ? JSON.stringify(value) : value
+        const previousValue = facetedUniqueValues.get(key)
         facetedUniqueValues.set(
-          value,
+          key,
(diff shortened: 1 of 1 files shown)


KevinVandy:
I'm hesitant to review any of your PRs @ErfanBagheri404 , because it looks like you're using an LLM to mass generate PRs or even regenerate other pre-existing PRs as your own. We usually block users who look like they are doing something like that.

ErfanBagheri404:
Thanks for the flag. These were meant as genuine fixes for open issues, but I hear you and don't want to add review load. Closing all of my open PRs. Feel free to take any of the approaches from my branches and implement them yourselves — no attribution needed. I just wanted the bugs fixed.

KevinVandy:
> Thanks for the flag. These were meant as genuine fixes for open issues, but I hear you and don't want to add review load. Closing all of my open PRs. Feel free to take any of the approaches from my branches and implement them yourselves — no attribution needed. I just wanted the bugs fixed.

Ok that's fine. Some of the ai prs are just getting a bit overwhelming. I'll co author you if the fix is similar.

PR #6560, opened 2026-08-16
refactor(angular): simplify lazy table initialization

benjavicente:
## 🎯 Changes

This can be considered a continuation of https://github.com/TanStack/table/pull/6534.

I saw 2 simplification opportunities in that PR:
- `previousOptions` isn't needed because the [signal will be considered the same with `Object.is`](https://angular.dev/guide/signals#signal-equality-functions)
- The effect to initialize the lazy object isn't needed. The `tableOptionsUpdate` `effect` will initialize the object  so the inner effect is redundant. The timing of when it is initialized is the same since those 2 effects were run together. The object will always be initialized after effect runs. Considering all of that, returning an object holding the initialized state and the raw value isn't needed. Calling `untracked(object)` every time could be slower but it isn't in a hot path.
- [Added due to coderabbit comment]: Removed `apply` trap for the proxy since that will never be reached, and shouldn't be necessary for the input signal workaround. 

## ✅ Checklist

- [X] I have followed the steps in the [Contributing guide](https://github.com/TanStack/table/blob/main/CONTRIBUTING.md).
- [X] I have tested this code locally with `pnpm run test:pr`.

## 🚀 Release Impact

- [X] This change affects published code, and I have generated a [changeset](https://github.com/changesets/changesets/blob/main/docs/adding-a-changeset.md).

## Summary by CodeRabbit

* **New Features**
  * Simplified lazy initialization for Angular table instances.
  * Table instances are available immediately, while setup is deferred until needed.
* **Bug Fixes**
  * Improved updates when table options or data change.
  * Added reliable cleanup when components are destroyed, including before initialization.
  * Improved compatibility across supported Angular versions.
* **Tests**
  * Expanded coverage for deferred initialization, proxy behavior, row updates, required inputs, and destruction cleanup.


packages/angular-table/src/lazySignalInitializer.ts
```diff
@@ -1,46 +1,24 @@
...
     apply(target: T, thisArg: any, argArray: Array<any>): any {
-      initializeObject()
-      if (typeof object === 'function') {
-        return Reflect.apply(object, thisArg, argArray)
+      const currentObject = untracked(object)
+      if (typeof currentObject === 'function') {
+        return Reflect.apply(currentObject, thisArg, argArray)
       }
       return Reflect.apply(target as any, thisArg, argArray)
     },
```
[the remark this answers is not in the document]
benjavicente: 3rd find: there is no need to add `apply` if we are not using `lazyInit` for functions, and we shouldn't anyway in Angular, since in that case we can just `() => untracked(lazyFnCreation)()`.

riccardoperra:
I am not sure if removing the ‘initialized’ boolean is fine. I mean, everything should work as is but I’ve introduced it to avoid a forced table initialization by accessing the property during the component destruction.

I guess something like this happen
- create component A
- register onDestroy callback
- running table init
- destroy component A before table init (here onDestroy callback could be invoked)

benjavicente:
True. I got a `NG0950: Input is required but no value is available yet` error by not waiting for stability or not calling detectChanges researching that case 😵 

I changed it such that the lazy function registers a cleanup callback only when the object has been initialized. 

This will get the same error when the table is read earlier that expected (like `table.options` in the instance construction), but at least it will not get the issue when the component is half-initialized, the table hasn't been read yet and the component gets destroyed.

packages/angular-table/src/injectLazyInit.ts
```diff
@@ -0,0 +1,43 @@
...
+): T {
+  assertInInjectionContext(injectLazyInit)
+  const destroyRef = inject(DestroyRef)
+  let object: T | null = null
+
+  const getObject = () => {
+    if (object === null) {
+      const initializedObject = untracked(initializer)
+      object = initializedObject
+      destroyRef.onDestroy(() => cleanup(initializedObject))
```
riccardoperra:
```suggestion
      if (!destroyRef.destroyed) {
        destroyRef.onDestroy(() => cleanup(initializedObject))
      }
```

I've added an extra check with the [destroyed](https://angular.dev/api/core/DestroyRef#destroyed) property

riccardoperra:
@benjavicente So I've thinked about it and I'm a bit doubtful about the destroy behavior.

My latest suggestion added `destroyRef.destroyed` check to prevent the destroy callback to be created if the view has already been destroyed. This is because if u do something like this will produce the `NG0911: View has already been destroyed` error.

```ts
destroyRef.onDestroy(() => {
  setTimeout(() => {
    lazy.something
  })
})
```

In my opinion that's not an impossible use case since someone may run an http call or something else async during destroy, and access to the table (while not already constructed)

My doubt is that we prevent the callback registration, but still the table has been created. I've pushed a new [commit](https://github.com/TanStack/table/pull/6560/changes/9b7f65e9fd12b2f4a4255e036b4f7a67d735a373) in your branch with those edits:

- destroyRef callback is now created eagerly, and checks the object is valued before calling the cleanup
- i'm using now a unique symbol value to detect the object is not initialized yet, so actually if your lazyInit returns undefined/null this still works (I know thats not a real use-case but actually that helper is like "abstract implementation")
- if the view has already been destroyed, and you try to access a non-initialized object, you'll now get an error

I’m not totally convinced, so I’d like to hear what you think

packages/angular-table/src/injectLazyInit.ts
```diff
@@ -5,23 +5,30 @@ import {
...
+  })
 
+  const getObject = () => {
+    if (destroyRef.destroyed && object === notInitializedObject) {
+      throw new Error(
+        '[@tanstack/angular-table] Cannot initialize object after view is destroyed',
+      )
+    }
+    if (object === notInitializedObject) {
+      object = untracked(initializer)
```
benjavicente, line 31: Why don't we attach the `destroyRef.onDestroy` here?
We know that the object has been initialized and that the components hasn't been destroyed.

benjavicente: Ignore this

benjavicente:
I think the timing issues of input signals, destroy ref and effects is really annoying 😅

It looks ok, not as simple as I expected but seems to cover all of the annoying edge cases. The only change I see that we should also take is changing `destroyRef.destroyed` to instead have a separate variable to track destroyment status, since older versions of Angular this library supports do not seem to have that field. I will take a look into running tests and examples with the lowest version, Angular 19.

riccardoperra:
@benjavicente As I remember `.destroyed` has been introduced in Angular 20. Angular 19 is not LTS anymore, so probably I wouldn't bother..I see no reason not to update honestly since 19->20 has not relevant breaking changes

btw: https://github.com/TanStack/table/pull/6560/changes/e290a3b0a138678283a84c5086e2bcd78b6e7e26

PR #6567, opened 2026-08-21
fix(table-core): guard process.env.NODE_ENV checks for bundler-less environments

### LINKED ISSUE #6078: `process is not defined` when used in Vanilla JS (without Node.js)

koreahghg:
## 🎯 Changes

Fixes #6078.

`table-core`'s ~14 dev-only debug/validation checks read `process.env.NODE_ENV` directly. That raw read survives unguarded into the published ESM build (`unbundle: true`, no `define` in `tsdown.config.ts`), so any environment with no `process` global — vanilla JS loaded via an import map, or any other bundler-less/non-Node setup — throws `ReferenceError: process is not defined` the moment one of these checks runs (e.g. simply constructing a table with `debugTable: true`).

Added a single `isDevelopmentEnv()` helper in `utils.ts` that checks `typeof process !== 'undefined'` before reading `process.env.NODE_ENV`, and routed all 14 call sites through it. Behavior is otherwise unchanged — still `=== 'development'`, not flipped to `!== 'production'` (that was the debated part of the related draft PR #6445, which is why it's stuck; this PR intentionally only fixes the crash and leaves that semantic question alone).

Verified the fix actually reaches the published artifact: after building, the only `process.env` reference left in `dist/utils.js` is the guarded one inside `isDevelopmentEnv()`.

## ✅ Checklist

- [x] I have tested this code locally with `pnpm run test:pr` (ran the equivalent steps directly: `vitest run` for `@tanstack/table-core` — 1275/1275 passing, including new regression tests — plus `tsc --noEmit`, `eslint`, and a `tsdown` build; the full `nx affected` pipeline timed out locally on an unrelated `test:sherif`/`test:knip` pass over the whole workspace).

## 🚀 Release Impact

### DIFF
--- a/.changeset/guard-process-env-checks.md
+++ b/.changeset/guard-process-env-checks.md
@@ -1,0 +1,5 @@
+---
+'@tanstack/table-core': patch
+---
+
+Guard the `process` global before reading `process.env.NODE_ENV` in development-only debug and validation checks. Raw `process.env.NODE_ENV` reads survived unguarded into the published ESM build, so any environment without a `process` global (e.g. vanilla JS loaded via an import map, or another bundler-less setup) threw `ReferenceError: process is not defined` the first time one of these checks ran. All ~14 call sites now go through a shared `isDevelopmentEnv()` helper that checks `typeof process !== 'undefined'` first.
--- a/packages/table-core/src/core/columns/constructColumn.ts
+++ b/packages/table-core/src/core/columns/constructColumn.ts
@@ -1,3 +1,4 @@
+import { isDevelopmentEnv } from '../../utils'
 import type { Table_Internal } from '../../types/Table'
 import type { CellData, RowData } from '../../types/type-utils'
 import type { TableFeatures } from '../../types/TableFeatures'
@@ -74,7 +75,7 @@ export function constructColumn<
         for (let i = 0; i < keys.length; i++) {
           const key = keys[i]!
           result = result?.[key]
-          if (process.env.NODE_ENV === 'development' && result === undefined) {
+          if (isDevelopmentEnv() && result === undefined) {
             console.warn(
               `"${key}" in deeply nested key "${accessorKey}" returned undefined.`,
             )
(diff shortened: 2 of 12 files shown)


MILLERMARRU (commented):
Checked the full diff, all 14 raw `process.env.NODE_ENV` reads got routed through `isDevelopmentEnv()` (grepped for leftover unguarded reads across the changed files, none left). The `typeof process !== 'undefined'` guard is the right fix for the actual bug (#6078), a `ReferenceError` on `process` itself, not a value-checking issue.

Agreed with keeping this scoped to just the crash. The `development` vs `!== production` semantics debated on #6445 is a separate, more contentious question (default-on vs default-off checks in unconfigured production builds) and bundling it here would've blocked a straightforward crash fix behind an unrelated design decision. Good call leaving that for #6445 to resolve on its own.

One thing I'd double check before merge: `isDevelopmentEnv()` itself needs to live somewhere that doesn't get tree-shaken oddly given it's now imported into ~10 different feature files, worth confirming the `dist/utils.js` check you did in the description also holds for the tree-shaken per-feature entry points if table-core publishes those separately, not just the main bundle.

koreahghg:
Good catch! checked this by actually building the branch (`tsdown`) and inspecting `dist/`.

Only `dist/utils.js` contains any `process.env` reference across the whole build, and it's the guarded `isDevelopmentEnv()`. Traced the import graph for all 6 published entry points:

- `index.js`, `static-functions.js`, `experimental-worker-plugin.js` → do reach `utils.js` (guarded)
- `flex-render.js`, `reactivity.js`, `store-reactivity-bindings.js` → don't reference `utils.js` at all (unrelated to the dev-check code, tree-shaken out)

So the guard holds across every per-feature entry point that actually uses it — no unguarded reads leak into any of the published bundles.

PR #6582, opened 2026-09-01
fix: guard pagination pageIndex/pageSize against NaN and non-positive pageSize

koreahghg:
## 🎯 Changes

`setPageIndex`/`setPageSize` could store `NaN` in pagination state when the updater resolved to `NaN` (e.g. binding `table.setPageSize(parseInt(e.target.value))` to a numeric input and clearing it: `parseInt('')` is `NaN`). `Math.max`/`Math.min` do not clamp `NaN` — it poisons through — so `NaN` silently reached state, and `rows.slice(NaN, NaN)` in the paginated row model quietly evaluated to `[]`: the table renders zero rows with no error surfaced anywhere.

Separately, `getPageCount`/`getPageOptions` could throw `RangeError: Invalid array length` when `pageSize` was `0` or negative. This is reachable via `initialState` or a directly-supplied `pagination` state, both of which bypass `setPageSize`'s `Math.max(1, ...)` clamp — e.g. `constructTable({ ..., initialState: { pagination: { pageIndex: 0, pageSize: 0 } } })` then calling `table.getPageOptions()` crashes immediately, and that's the API most pagination-control UIs call to render page-number buttons.

This PR:
- Makes `setPageIndex`/`setPageSize` fall back to the default page index/size when the updater resolves to `NaN`, instead of storing it.
- Makes `getPageCount` return `0` for a non-positive (or non-numeric) `pageSize` instead of dividing into `Infinity`/`NaN`, so `getPageOptions` returns `[]` instead of throwing.

## ✅ Checklist

## 🚀 Release Impact

## Test plan

- Added regression tests in `rowPaginationFeature.utils.test.ts` covering: `setPageIndex`/`setPageSize` with a `NaN` update, and `getPageCount`/`getPageOptions` with `pageSize: 0` and a negative `pageSize`.
- `nx run @tanstack/table-core:test:lib` (1334 tests), `test:types`, `test:eslint`, and `build` all pass.
- `pnpm test:pr` (affected-project lint/sherif/knip/lib/types/build) passes across 418 affected projects.
- `pnpm test:e2e:affected`: all pass except the pre-existing `vue/realtime-trading` smoke test, which I confirmed also fails on `main` without this change (unrelated timing issue in that example's pause/resume flow, not touched by this PR).

## Summary by CodeRabbit

* **Bug Fixes**
  * Improved pagination handling when page index or page size values are invalid.
  * Prevented errors and invalid results when page size is zero, negative, or unavailable.
  * Pagination now falls back to safe defaults and recalculates the current page correctly.

### DIFF
--- a/.changeset/fix-pagination-nan-page-size.md
+++ b/.changeset/fix-pagination-nan-page-size.md
@@ -1,0 +1,5 @@
+---
+'@tanstack/table-core': patch
+---
+
+Fix pagination state corruption from invalid `pageSize`/`pageIndex` values: `setPageIndex`/`setPageSize` now fall back to the default state instead of storing `NaN` (e.g. from an emptied numeric input), and `getPageCount`/`getPageOptions` no longer throw when `pageSize` is `0` or negative (reachable via `initialState` or a directly-supplied `pagination` state).
--- a/packages/table-core/src/features/row-pagination/rowPaginationFeature.utils.ts
+++ b/packages/table-core/src/features/row-pagination/rowPaginationFeature.utils.ts
@@ -107,7 +107,9 @@ export function table_resetPagination<
  * Updates `pagination.pageIndex` and clamps it to the known page range.
  *
  * Unknown page counts (`undefined` or `-1`) allow any non-negative page index.
- * Known page counts clamp the index between `0` and `pageCount - 1`.
+ * Known page counts clamp the index between `0` and `pageCount - 1`. A `NaN`
+ * update (e.g. from an emptied numeric input) falls back to the default page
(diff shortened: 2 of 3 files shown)


codeCraft-Ritik (commented):
The unit tests in rowPaginationFeature.utils.test.ts directly targeting the NaN, 0, and negative boundaries look solid.

Given the wide blast radius of @tanstack/table-core, verifying pnpm test:pr across all 418 affected packages provides strong confidence. Good note explaining the unrelated vue/realtime-trading e2e failure on main as well.

PR #6587, opened 2026-09-09
fix(table-core): include defaultColumn in memoDeps for column, header, and sizing computations

### LINKED ISSUE #5275: A table doesn't get updated when a defaultColumn changes
### Describe the bug

I used one of your examples and added a defaultColumn with an initial size to the table, which works fine. However if the defaultColumn gets updated later on, this change doesn't get reflected in the actual size of the table columns. 
So if we have the below code and I set the initial size to 100 and then in the useEffect I update it to 50, it will still remain 100 for the rendered table even if the table generated by the useReactTable gets the new value. I checked the same with columnResizeDirection and it works fine. 

```
const [defaultColumn, setDefaultColumn] = React.useState({size: 200})

  React.useEffect(() => {
    setDefaultColumn({size: 50})
  }, [])

  const table = useReactTable({
    data,
    columns,
    columnResizeMode,
    defaultColumn,
    columnResizeDirection,
    debugTable: true,
    debugHeaders: true,
    debugColumns: true,
  })
```

https://codesandbox.io/p/sandbox/reacttable-lycjfh

### Steps to reproduce

As in description

### Expected behavior

The size of the columns gets updated when the defaultColumn gets changed

Every time

### Screenshots or Videos

_No response_

### Platform

macOS Chrome

### react-table version

8.11.6

### TypeScript version

4.4.2

### Additional context

_No response_

### Terms & Code of Conduct

lx3133584:
Fixes #5275

## 🎯 Changes

When a table is constructed, `constructColumn` merges `defaultColumn` into each column's `columnDef`. However, the memoized APIs across core features:
- `table_getAllColumns`
- `table_getAllFlatColumns`
- `table_getAllFlatColumnsById`
- `table_getAllLeafColumns`
- `column_getFlatColumns`
- `column_getLeafColumns`
- `table_getHeaderGroups`
- `header_getLeafHeaders`
- `header_getContext`
- `column_getSize`
- `header_getSize`
- `header_getStart`
- `table_getColumnOffsets`
- `column_getIsVisible`
- `table_getVisibleFlatColumns`
- `table_getVisibleLeafColumns`

only included `table.options.columns` in their `memoDeps`, omitting `table.options.defaultColumn`.

As a result, when `defaultColumn` was updated dynamically via `table.setOptions(...)` (such as in an effect or upon state changes), the column tree, headers, and sizing were not invalidated and continued to return stale definitions and sizes.

This PR adds `defaultColumn` to `memoDeps` across `coreColumnsFeature`, `coreHeadersFeature`, `columnSizingFeature`, and `columnVisibilityFeature`, and adds unit tests covering `defaultColumn` updates via `setOptions`.

## ✅ Checklist

## 🚀 Release Impact

## Summary by CodeRabbit

- **Bug Fixes**
  - Updated column, header, sizing, and visibility calculations when default column settings change.
  - Changes to default column metadata now appear consistently across columns and headers.
  - Updated default column sizes now correctly refresh column and header sizing.

- **Tests**
  - Added coverage for changes to default column metadata and sizing.


KevinVandy:
This has been proposed a few times lately, but I think it would be too large of a breaking change and hurt performance for most tables, especially in react. This means that it would be expected for everyone to give their defaultColumn option a stable reference with useMemo, otherwise, all the calculations would be constantly re-firing

lx3133584:
Makes total sense — unstable `defaultColumn` object references without `useMemo` in React apps would indeed cause constant re-computations across all tables. Happy to close this out if keeping the current behavior is preferred.

PR #6597, opened 2026-09-22
fix(table-core): give TableFeature hooks an all-features table type

### LINKED ISSUE #6594: Cast required to access custom options when defining a custom plugin/feature
### TanStack Table version

9.2.4

### Framework/Library version

Vue 3.5.43

When defining a custom plugin/feature, a cast is required to access any options that it adds. For example, this is can be observed in [an official example for custom plugins](https://tanstack.com/table/latest/docs/framework/react/examples/custom-plugin)--on lines 108 and 119, `table.options` requires a cast to access `onDensityChange`. Here's the relevant code for convenience:

```js
  constructTableAPIs: (table) => {
    assignTableAPIs('densityPlugin', table, {
      table_setDensity: {
        fn: (updater: Updater<DensityState>) => {
          const safeUpdater: Updater<DensityState> = (old) => {
            const newState = functionalUpdate(updater, old)
            return newState
          }
          return (table.options as TableOptions_Density).onDensityChange?.(
            safeUpdater,
          )
        },
      },
      table_toggleDensity: {
        fn: (value?: DensityState) => {
          const safeUpdater: Updater<DensityState> = (old) => {
            if (value) return value
            return old === 'lg' ? 'md' : old === 'md' ? 'sm' : 'lg' // cycle through the 3 options
          }
          return (table.options as TableOptions_Density).onDensityChange?.(
            safeUpdater,
          )
        },
      },
    })
  },
```

In theory, the internal `Table_Internal` type used in [the TableFeature definition](https://github.com/TanStack/table/blob/21d713fc4947d2a08cc2136bb055889a61412ded/packages/table-core/src/types/TableFeatures.ts#L386) should be able to load options added by plugins, but this type is stripped from the published code (I guess using the code from [this PR](https://github.com/TanStack/table/pull/6326)?), so the type exposed to custom plugin authors is the regular `Table`, which does not come with plugins by default. Interestingly, `Table` has support for `Table<any, any>` to load all plugins, but trying to set that on the `table` in the `constructTableAPIs` function causes its own type error since that doesn't match the type set by `TableFeature`.

```js
export const densityPlugin: TableFeature = {
  constructTableAPIs: (table: Table<any, any>) => {
    // can now access table.options.onDensityChange with the correct types, but now this function itself causes a type error:
/*
Type
<TFeatures extends TableFeatures, TData extends RowData>(table: Table<any, any>) => void
is not assignable to type
<TFeatures extends TableFeatures, TData extends RowData>(table: Table<TFeatures, TData>) => void
*/ 
  }
}
```

I'm not sure what the best way to solve this is; I see a few possibilities:

- leave it as is; a couple type casts isn't a huge deal.
  - in this case, the docs should be updated--e.g., [the custom features guide](https://tanstack.com/table/latest/docs/framework/react/guide/custom-features#step-3-create-the-feature-object) says "If the TypeScript types are set up correctly, you should have no TypeScript errors when you create the feature object with the new state, options, and instance APIs." but the example code immediately afterwards does not have the cast, so it actually does have TS errors.
- change TableFeature to use a different/new `Table_All` type that actually gets published
- change TableFeature to accept generic parameters for its own custom options/state/etc.? (not sure we'd have access to the right TFeatures/TData then, though, and I have no idea what that would do to typechecking performance.)

https://tanstack.com/table/latest/docs/framework/react/examples/custom-plugin

_No response_

None

### Terms & Code of Conduct

sorena-paydar:
## 🎯 Changes

Fixes #6594.

`TableFeature` hooks are typed with `Table_Internal`, which the dts rewrite strips and rewrites to `Table`. `Table` narrows `options` to `TFeatures`, so a custom feature cannot read the options it contributes itself, and every custom-plugin example had to cast:

```ts
return (table.options as TableOptions_Density).onDensityChange?.(safeUpdater)
```

This adds a published `Table_All` — `Table` widened to the all-features `options`, `state`, and row models — and rewrites `Table_Internal` onto it in `TableFeatures.d.ts` only. This is option 2 of the three the issue lists. It matches the `_All` convention already used across the public surface (`TableOptions_All`, `TableState_All`, `CachedRowModel_All` and five others are published today, and `getDefaultTableOptions` already returns `Partial<TableOptions_All>`).

Two details worth a reviewer's eye:

**The rewrite is scoped to `TableFeatures.d.ts`.** `Header`, `Cell`, `Column` and `Row` also reference `Table_Internal`. Rewriting those to `Table_All` too would widen `store` and `atoms` for ordinary consumers and make state slices optional — `examples/react/virtualized-columns-experimental` stops typechecking with `'state.columnSizing' is possibly 'undefined'`. Those files keep plain `Table`.

**`Table_All` is `Table & { …widened members }`, not a rebuild of the broad shape.** `Table_Internal` does not include `ExtractFeatureMapTypes`, so defining `Table_All` from the same parts drops the feature APIs and `table.getIsAllRowsSelected()` disappears — `examples/svelte/composable-tables` catches that. Intersecting with `Table` keeps `Table_All` a strict superset, so this is additive for existing users.

`Table_Internal` is unchanged and still stripped from the emitted declarations, so the no-internal-types invariant from #6326 still holds.

Also drops the now-unneeded casts from the custom-plugin examples (react, preact, alpine, angular, ember, octane) and the two guides that carried them. The guides already claimed "you should have no TypeScript errors when you create the feature object", which the issue notes was untrue; it is now true. The vendored `material-react-table` example keeps its `MRT_*` casts — they still compile, and changing that example is out of scope here.

The de-casted examples double as the regression test: `test:types` runs on them in CI, so if the published hook type ever stops exposing all-features options, they fail.

## ✅ Checklist

On testing, so you know exactly what was run rather than trusting the box: `@tanstack/table-core` `test:lib` (1330 tests), `test:types`, `test:eslint` and `test:build` all pass, and `test:types` passes for all six custom-plugin examples plus `svelte/composable-tables`, `react/virtualized-columns-experimental` and `react/material-react-table`. I could not complete a full `nx affected` run locally because `@tanstack/table-devtools:build` fails on my machine with `[BABEL] … Cannot read properties of undefined (reading 'explode')` in `packages/table-devtools/src/core.tsx`; that reproduces identically on a clean `main` checkout, so it is pre-existing and unrelated, but it blocks the kitchen-sink examples that depend on devtools. `test:e2e` does not apply — there is no runtime change in this PR.

## 🚀 Release Impact

## Summary by CodeRabbit

* **New Features**
  * Added the exported `Table_All` type for comprehensive table configurations across features, options, state, and row models.
  * Custom feature hooks now receive strongly typed table information, making feature-specific options and state easier to access without manual type casting.

* **Documentation**
  * Updated custom density feature guides and examples across supported frameworks to demonstrate streamlined, strongly typed callback usage.


sorena-paydar:
Good catch on the path separator — that was a real bug, not a style nit. `walkDeclarationFiles` builds paths with `path.join`, so on Windows the suffix check would never match and `TableFeatures.d.ts` would silently fall back to `Table`, meaning a Windows build would publish different types than a Linux one.

Fixed by normalizing separators in an `isDeclarationPath` helper, and applied it to the two pre-existing `endsWith` checks in `ensurePublicTypeImports` as well, since they had the same latent issue.

Rebuilt and re-verified: `TableFeatures.d.ts` still gets `Table_All` (8 references), `Header`/`Cell`/`Column`/`Row` still get plain `Table`, `Table_Internal` still does not leak, and `table-core` is green (1330 tests, types, eslint).

